PR #9, opened 2019-06-11
Move run script to root.

squidsoup:
Adding support to the demoservice for monorepos is potentially complex, so as a compromise we'll move `run` to the root, and selectively point at an individual workspace project. Currently the `serve' target in `package.json` will point to the `ui` workspace, but later will point to the single-spa root project once that is implemented. This does mean we will not have demos for the legacy maas frontend when it is migrated to this repo, but I think we can live with that given it has *never* had demos.

# QA
Run `run` from the project root. It should spin up the `ui` app on http://localhost:8000


huwshimi:
I've added https://github.com/canonical-web-and-design/MAAS-squad/issues/1236

PR #11, opened 2019-06-12
Set up Redux.

huwshimi:
Done:
- Added redux, and required deps.
- Added fake machine listing for redux demo.
- Improved project structure.
- Added second `./run` for development.

QA:
- Open the demo link, there should be a couple of machines.

Fixes: https://github.com/canonical-web-and-design/MAAS-squad/issues/1237.


ui/src/app/App.js
```diff
@@ -0,0 +1,58 @@
...
+};
+
+// Filters will live inside their respective modules.
+const getMachines = (state, props) => state.machines;
+
+const getReadyMachines = createSelector(
+  [getMachines],
+  machines => machines.filter(machine => machine.ready)
+);
+
```
squidsoup: Might be good to add another comment here like the previous one, such that action creators are in their own modules. Realise this is throwaway code...

ui/src/index.js
```diff
@@ -1,10 +1,24 @@
...
-import App from "./App";
+
+import "./scss/base.scss";
 import * as serviceWorker from "./serviceWorker";
+import App from "./app/App";
+import reducers from "./root-reducer";
+
+const store = createStore(
+  reducers,
+  window.__REDUX_DEVTOOLS_EXTENSION__ && window.__REDUX_DEVTOOLS_EXTENSION__()
```
squidsoup: Maybe leave a comment here that once we're using redux middleware like sagas, we'll probably want to change this to use the [redux-devtools-extension](https://github.com/zalmoxisus/redux-devtools-extension#13-use-redux-devtools-extension-package-from-npm).

huwshimi: Oh nice!

ui/src/app/App.js
huwshimi: Done.

PR #40, opened 2019-07-24
Open websocket connection with csrftoken from cookie.

squidsoup:
## Done
* Provide csrftoken in querystring when opening websocket connection. (No need to hack `protocol.py` now!)

## QA
* Revert hacks to `protocol.py`, login and opening the websocket connection should work.

### DIFF
--- a/ui/src/app/base/sagas/websockets.js
+++ b/ui/src/app/base/sagas/websockets.js
@@ -1,14 +1,16 @@
 import { eventChannel } from "redux-saga";
 import { all, call, put, take, race } from "redux-saga/effects";
 
+import getCookie from "./utils";
 import WebSocketClient from "../../../websocket-client";
 
 /**
  * Create a WebSocket connection via the client.
  */
-export function createConnection() {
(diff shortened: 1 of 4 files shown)


ui/src/app/base/sagas/websockets.js
```diff
@@ -1,14 +1,16 @@
...
 import WebSocketClient from "../../../websocket-client";
 
 /**
  * Create a WebSocket connection via the client.
  */
-export function createConnection() {
+export function createConnection(csrftoken) {
   return new Promise((resolve, reject) => {
-    const socketClient = new WebSocketClient(MAAS_config.ui.websocket_url); // eslint-disable-line no-undef
+    const url = `${MAAS_config.ui.websocket_url}?csrftoken=${csrftoken}`; // eslint-disable-line no-undef
```
huwshimi: I think this should probably happen as part of loading the config. Then maybe we can just pass the generated ws URL to the saga. In dev it should just pass the ws URL (or I guess not append the csrftoken if the cookie is not found).

PR #44, opened 2019-07-29
Fetch users

squidsoup:
## Done
* the Users component now fetches users, and displays a loading component.

## QA
* Load the user settings page, it should display users from MAAS.


ui/src/app/settings/actions/repositories.js
```diff
@@ -1,6 +1,6 @@
 const repositories = {};
 
-repositories.fetchRepositories = () => {
+repositories.fetch = () => {
```
squidsoup: just making the api consistent here.

ui/src/app/settings/reducers/repositories.test.js
```diff
@@ -1,6 +1,6 @@
 import repositories from "./repositories";
 
-describe("status", () => {
+describe("repositories", () => {
```
squidsoup: suspect this was copy/paste detritus.

ui/src/app/settings/containers/Users/Users.js
```diff
@@ -1,13 +1,23 @@
...
+import selectors from "app/settings/selectors";
 import { UserShape } from "app/base/proptypes";
 import baseSelectors from "app/base/selectors";
 import MainTable from "app/base/components/MainTable";
 
-export const Users = ({ authUser, users }) => {
+const Users = () => {
+  const dispatch = useDispatch();
+  const users = useSelector(selectors.users.get);
+  const authUser = useSelector(baseSelectors.auth.getAuthUser);
```
squidsoup: The new redux hooks API obviates the `connect` HoC, but introduces some coupling.

ui/src/app/base/components/Loader/Loader.js
```diff
@@ -0,0 +1,21 @@
...
+import "./Loader.scss";
+
+const Loader = ({ text, isLight, inline }) => (
+  <div
+    className={classNames("p-loader", "p-text--default", {
+      "p-loader--inline": inline
+    })}
+  >
+    <i
+      className={`p-icon--spinner u-animation--spin ${
```
huwshimi: You could update this to use `classNames` here too.

ui/src/app/settings/actions/repositories.js
huwshimi: Thanks.

ui/src/app/settings/containers/Users/Users.test.js
```diff
@@ -1,22 +1,45 @@
...
+    const wrapper = mount(
+      <Provider store={store}>
+        <MemoryRouter
+          initialEntries={[{ pathname: "/settings/users", key: "testKey" }]}
+        >
+          <Users />
+        </MemoryRouter>
+      </Provider>
     );
     expect(wrapper).toMatchSnapshot();
```
huwshimi: Should we just snapshot the component we are testing so that the provider, router etc. aren't in the snapshot? It would make the snapshot a lot more useful maybe. E.g. `expect(wrapper.find('Users')).toMatchSnapshot();`.

ui/src/app/settings/reducers/repositories.test.js
huwshimi: Oops, thanks.

huwshimi:
QA:
- When first visiting the page it flashes the table header before displaying the loading indicator, should it default to the loader?
- What should happen if you navigate away from the users page and back again? At the moment it reloads the users. If it has a list of users should it display the user list and update in the background?

ui/src/app/settings/containers/Users/Users.js
```diff
@@ -49,74 +51,64 @@ export const Users = ({ authUser, users }) => {
...
-      last_name: "",
-      sshkeys_count: 9,
-      username: "simon"
+  useEffect(() => {
+    if (!users.length) {
+      dispatch(actions.users.fetch());
     }
-  ]
-};
+  }, [dispatch, users.length]);
```
huwshimi: The `users.length` dep means this useEffect runs twice, when you first load the component and then when the users are returned from the API. This would also mean it would run if users are added/removed over the websocket.

ui/src/app/settings/containers/Users/Users.js
```diff
@@ -49,74 +51,64 @@ export const Users = ({ authUser, users }) => {
...
-      />
-    </>
-  );
-};
 
-Users.propTypes = {
-  authUser: UserShape.isRequired,
-  users: PropTypes.arrayOf(UserShape)
-};
+const Users = ({ initialCount = 20 }) => {
```
huwshimi: This is missing from the propTypes.

ui/src/app/settings/containers/Users/Users.test.js
```diff
@@ -1,23 +1,126 @@
...
-import { Users } from "./Users";
+import { Provider } from "react-redux";
+import configureStore from "redux-mock-store";
+import { MemoryRouter } from "react-router-dom";
+import Users from "./Users";
+
+const mockStore = configureStore();
 
 describe("Users", () => {
-  it("can render", () => {
```
huwshimi: Do you think we need to check any of the props on the generated table?

ui/src/app/settings/containers/Users/Users.js
```diff
@@ -52,27 +53,33 @@ const generateUserRows = (users, authUser) =>
...
+  const loaded = useFetchOnce(actions.users.fetch, selectors.users.loaded);
 
   return (
     <>
       <h4>
         Users
         {loading && <Loader text="Loading..." inline />}
       </h4>
-      {users.length > 0 && (
+      {loaded > 0 && (
```
huwshimi: This is a bool so no need for the `> 0`.

PR #45, opened 2019-07-31
Add user form

huwshimi:
Done:
- Add user add and edit pages and form. Fixes: https://github.com/canonical-web-and-design/maas-squad/issues/1353.
- Added form, input, select, textarea, button and card components.

QA:
- open the users list
- click the edit pencil icon on a user
- you should see a user form like this: https://app.zeplin.io/project/5ce5423a25863c1e0d7abb27/screen/5d38490dfcd3ff77608afba4

### DIFF
--- a/ui/src/app/settings/components/UserForm/UserForm.test.js
+++ b/ui/src/app/settings/components/UserForm/UserForm.test.js
@@ -1,0 +1,10 @@
+import { shallow } from "enzyme";
+import React from "react";
+import { UserForm } from "./UserForm";
+
+describe("UserForm", () => {
+  it("can render", () => {
+    const wrapper = shallow(<UserForm title="Add user" />);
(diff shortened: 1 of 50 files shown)


ui/src/app/settings/components/UserForm/UserForm.test.js
```diff
@@ -0,0 +1,10 @@
+import { shallow } from "enzyme";
+import React from "react";
+import { UserForm } from "./UserForm";
+
+describe("UserForm", () => {
+  it("can render", () => {
+    const wrapper = shallow(<UserForm title="Add user" />);
+    expect(wrapper).toMatchSnapshot();
+  });
+});
```
squidsoup: UserForm seems to have a lot of logic that we aren't testing. I suppose we want to avoid testing Formik itself, but perhaps we should validate that some of the form's expected behaviours?

Issue #46, opened 2019-08-01
Implement user add/edit cancel 


huwshimi:
Use the history api to go back.

PR #49, opened 2019-08-02
Add user navigation

huwshimi:
Done:
- Add an add user button.
- Make cancel go back.
- Add Row and Col components.

Fixes: #46.

QA:
- Visit /users.
- Click "Add user", you should go to the user form.
- Click "Cancel", you should go back.

### DIFF
--- a/ui/src/app/base/components/Col/Col.js
+++ b/ui/src/app/base/components/Col/Col.js
@@ -1,0 +1,73 @@
+import classNames from "classnames";
+import PropTypes from "prop-types";
+import React from "react";
+
+const colRange = [
+  1,
+  2,
+  3,
+  4,
+  5,
+  6,
+  7,
+  8,
+  9,
+  10,
+  11,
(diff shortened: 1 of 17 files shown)


ui/src/app/base/components/Col/Col.js
```diff
@@ -0,0 +1,73 @@
...
+  "4",
+  "5",
+  "6",
+  "7",
+  "8",
+  "9",
+  "10",
+  "11",
+  "12"
+];
```
squidsoup: This could be more succinct, but I'm not honestly sure if it is better!
```javascript
let colRange = Array.from({length: 13}, (v, i ) => i);
colRange = [...colRange, ...colRange.map(i => `${i}`)]
```

huwshimi: I thought about this too, I'm not sure if gains us anything except for not having strange Prettier formatting.

Issue #54, opened 2019-08-05
Refactor redux actions


squidsoup:
Currently we pass nested redux payloads to our websockets layer, e.g.: 

```javascript
repositories.fetch = () => {
  return {
    type: "WEBSOCKET_SEND",
    payload: {
      actionType: "FETCH_REPOSITORIES",
      message: {
        method: "packagerepository.list",
        params: { limit: 50 },
        type: 0
      }
    }
  };
};
```
This does have the advantage of making the saga code very succinct, but it is a bit unusual and in contravention of [flux standard actions](https://github.com/redux-utilities/flux-standard-action). Instead, we should probably dispatch a payload that looks more like:

```javascript
repositories.fetch = () => (
{
  type: "FETCH_REPOSITORIES",
  meta: {
    limit: 50
  }
})     
``` 

That way the redux actions have no knowledge of the websocket implementation, which can be encapsulated in a saga (`type`, `params`, `method` etc).

huwshimi:
@squidsoup was this closed by: https://github.com/canonical-web-and-design/maas-ui/pull/85 ?

PR #57, opened 2019-08-05
Create Storage component in Settings

Caleb-Ellis:
## Done
- Created MAAS config selectors, reducers and actions
- Created non-persistent Storage component in settings. Saving the form will be implemented in canonical-web-and-design/MAAS-squad#1389

## QA
- Check that config is loaded into state on load (in Redux Dev Tools for example)
- Check that the React storage settings match those in the Django version.
- Check that changing the Django settings correctly updates the React settings on reload.
- Check that a warning shows up for blank and vmfs6 storage layouts.
- Check that clicking save displays an alert with the currently set values.


ui/src/app/settings/containers/Settings/Settings.js
```diff
@@ -1,10 +1,15 @@
...
+import { useDispatch } from "react-redux";
 
+import actions from "app/settings/actions";
 import Routes from "../../components/Routes";
 import Section from "app/base/components/Section";
 import SettingsNav from "../../components/Nav";
 
 const Settings = () => {
+  const dispatch = useDispatch();
+  dispatch(actions.config.fetch());
```
squidsoup: Same comment as for Steve, but you'll need to read up about lifecycle management with hooks and [`useEffect`](https://reactjs.org/docs/hooks-effect.html). Also have a look at the [`useFetchOnce`](https://github.com/canonical-web-and-design/maas-ui/blob/2d85ce8fd9ce0441c9f1a23c1679d0fe8a96fad3/ui/src/app/settings/containers/Users/Users.js#L56) hook, which will be replaced soon, but is okay to use for now.

ui/src/app/settings/containers/Storage/Storage.js
```diff
@@ -1,5 +1,125 @@
...
+  );
+  const enableDiskErasing = useSelector(selectors.config.enableDiskErasing);
+  const diskEraseWithSecure = useSelector(selectors.config.diskEraseWithSecure);
+  const diskEraseWithQuick = useSelector(selectors.config.diskEraseWithQuick);
+
+  return (
+    <>
+      <h4>
+        Storage
+        {!loaded && <Loader text="Loading..." inline />}
```
squidsoup: Should thing just be `loading`, in case the logic changes at some point which means the component loads as the result of an event etc?

ui/src/app/settings/reducers/config.js
```diff
@@ -0,0 +1,25 @@
+import produce from "immer";
+
+const config = produce(
```
squidsoup: Ah, looks like you and @steverydz have overlapped a bit here. You guys may need to coordinate to get this landed.

ui/src/app/settings/selectors/config.js
```diff
@@ -0,0 +1,119 @@
...
+ * @returns {Boolean} User has loaded.
+ */
+config.loaded = state => state.config.loaded;
+
+/**
+ * Returns the MAAS config for default storage layout.
+ * @param {Object} state - The redux state.
+ * @returns {String} Default storage layout.
+ */
+config.defaultStorageLayout = createSelector(
```
squidsoup: @steverydz some good examples of reselect selector composition here for you.

ui/src/app/settings/selectors/config.js
```diff
@@ -0,0 +1,119 @@
+import { createSelector } from "reselect";
+
+const config = {};
+
+/**
+ * Returns MAAS config
+ * @param {Object} state - The redux state.
+ * @returns {Array} A list of all state.config.items.
+ */
+config.get = state => state.config.items;
```
squidsoup: @Caleb-Ellis I think we've decided to call this `all` to distinguish from a `get` selector which takes a prop as a param.

ui/src/app/settings/actions/config.js
```diff
@@ -0,0 +1,16 @@
+const config = {};
+
+config.fetch = () => {
+  return {
+    type: "WEBSOCKET_SEND",
```
squidsoup: fyi, this is a bit weird, and will be [fixed shortly](https://github.com/canonical-web-and-design/maas-ui/issues/54).

ui/src/app/settings/containers/Settings/Settings.js
```diff
@@ -1,10 +1,26 @@
...
+import React, { useEffect } from "react";
+import { useDispatch, useSelector } from "react-redux";
 
+import actions from "app/settings/actions";
+import selectors from "app/settings/selectors";
 import Routes from "../../components/Routes";
 import Section from "app/base/components/Section";
 import SettingsNav from "../../components/Nav";
 
+const useFetchOnce = (fetchAction, loadedSelector) => {
```
Caleb-Ellis: I've just copy-pasted this here, but will change to import from hooks.js once https://github.com/canonical-web-and-design/maas-ui/pull/61 has landed

ui/src/testing/utils.js
```diff
@@ -17,3 +17,26 @@ export const compareJSX = (actual, expected) => {
...
+/**
+ * Replace objects in an array with objects that have new values, given a match
+ * criteria.
+ * @param {Array} array - Array to be reduced.
+ * @param {String} key - Object key to compare the match criteria e.g. "name".
+ * @param {String} match - Match criteria e.g. "Bob".
+ * @param {Object} newValues - Values to insert or update in the object.
+ * @returns {Array} The reduced array.
+ */
+export const reduceInitialState = (array, key, match, newValues) => {
```
Caleb-Ellis: @squidsoup @huwshimi as I was writing the tests for Storage I was wondering - do we want a file somewhere that just has all the initial state you would expect from the app in steady-state (i.e. everything has loaded)? Then we could import it into any connected component test file, and use a simple function like this to mock the small pieces of state differences that we're interested in. Any thoughts?

huwshimi: From memory crbs has one of those and it seemed helpful.

squidsoup: Yep we use [this](https://github.com/canonical-web-and-design/crbs-ui/blob/master/src/testing/default-state.js) in rbac, and do something like this to ensure defaultState isn't mutated:

```javascript
  beforeEach(() => {
    jest.resetModules();
    // eslint-disable-next-line global-require
    state = require("../../../testing/default-state").default;
  });
```

ui/src/app/settings/containers/Settings/Settings.js
squidsoup: This should be imported from [`hooks.js`](https://github.com/canonical-web-and-design/maas-ui/blob/master/ui/src/app/base/hooks.js) now.

ui/src/app/settings/selectors/config.js
```diff
@@ -0,0 +1,119 @@
...
+
+/**
+ * Returns the MAAS config for enabling disk erase with secure erase.
+ * @param {Object} state - The redux state.
+ * @returns {Boolean} Enable disk erasing with secure erase.
+ */
+config.diskEraseWithSecure = createSelector(
+  [config.all],
+  configs => {
+    const configObj = configs.find(
```
squidsoup: As this pattern is repeated several times, could we factor this out to a helper function with takes `configs` and the config name?

ui/src/testing/utils.js
Caleb-Ellis: Ah, nice! I didn't realise you'd already done that.

PR #58, opened 2019-08-05
WIP: Fetch config settings

steverydz:


ui/src/app/settings/containers/Configuration/Configuration.js
```diff
@@ -1,5 +1,31 @@
...
+
+const Configuration = () => {
+  const dispatch = useDispatch();
+  const configuration = useSelector(state =>
+    selectors.configuration.all(state)
+  );
+  const loading = useSelector(selectors.configuration.loading);
+  const loaded = useSelector(selectors.configuration.loaded);
+
+  dispatch(actions.configuration.fetch());
```
squidsoup: This needs to be a [`useEffect`](https://reactjs.org/docs/hooks-effect.html), otherwise this component will fetch every time it renders. Have a look at the examples with `useFetchOnce`. That hook will be replaced soon (that logic belongs in a saga), but dispatching always needs to be part of the lifecycle or bound to an event.

ui/src/app/settings/containers/Configuration/Configuration.js
```diff
@@ -1,5 +1,31 @@
...
 
-const Configuration = () => <h4>Configuration</h4>;
+import actions from "app/settings/actions";
+import selectors from "app/settings/selectors";
+import baseSelectors from "app/base/selectors";
+import Loader from "app/base/components/Loader";
+
+const Configuration = () => {
+  const dispatch = useDispatch();
+  const configuration = useSelector(state =>
```
squidsoup: rather than just returning the entirety of configuration, it would be nice to have methods which return distinct bits of state.

ui/src/app/settings/selectors/configuration.js
```diff
@@ -0,0 +1,26 @@
+const configuration = {};
+
+/**
+ * Returns all configuration.
+ * @param {Object} state - The redux state.
+ * @returns {Array} A list of all configuration.
+ */
+configuration.all = state => {
+  return state.configuration.items;
+};
```
squidsoup: You can just say `configuration.all = state => (state.configuration.items)`.

Although `all` is not likely to be consumed directly by a component (see earlier comment), you can compose new, more specific selectors on top of this with reselect.

PR #60, opened 2019-08-06
Add pager component and paginate users table.

squidsoup:
## Done
1. Create `Pager` component.
2. Added pagination to the User table.

The `Pager` component supports both truncated and simple pagination [as described in the vanilla docs](https://docs.vanillaframework.io/patterns/pagination/).

image

## QA
1. Ensure the users table renders (if you have more than 12 users, the pager should render).
2. Ensure the *previous* button is disabled when on the first item.
3. Ensure the *next* button is disabled when on the last item.
4. Click page 1-3, the pager should not truncate initially e.g. 
image

but should truncate past 4.
5. Click the last page, or 1-3 pages from the last, it should not truncate initially, .e.g. 
image

6. In `Users.js` manually hack the `Pager` to have fewer `totalItems` than `itemsPerPage`, the `Pager` should not render. (Probably easier than hacking users in your state).
7. Hack the pager to only have a few pages, e.g. `totalItems={50}` and `itemsPerPage={10}`, you should see a simple `Pager` render without truncation.


ui/src/app/base/components/Pager/Pager.js
```diff
@@ -0,0 +1,132 @@
...
+import React from "react";
+
+import PagerButton from "app/base/components/PagerButton";
+import PagerItem from "app/base/components/PagerItem";
+
+const PagerItemSeparator = () => (
+  <li className="p-pagination__item p-pagination__item--truncation">…</li>
+);
+
+const Pager = ({
```
huwshimi: Could we call this `Pagination` to be consistent with Vanilla?

ui/src/app/base/components/Pager/Pager.js
```diff
@@ -0,0 +1,132 @@
+import PropTypes from "prop-types";
+import React from "react";
+
+import PagerButton from "app/base/components/PagerButton";
+import PagerItem from "app/base/components/PagerItem";
+
+const PagerItemSeparator = () => (
+  <li className="p-pagination__item p-pagination__item--truncation">…</li>
```
huwshimi: I think it might be safer to use `&hellip;` instead of the asci `…`.

ui/src/app/base/components/Pager/Pager.js
```diff
@@ -0,0 +1,132 @@
...
+    return null;
+  }
+
+  const pageNumbers = [];
+
+  for (let i = 1; i <= Math.ceil(totalItems / itemsPerPage); i++) {
+    pageNumbers.push(i);
+  }
+
+  const generatePagerItems = (pageNumbers, currentPage) => {
```
huwshimi: Should this be defined outside of the component?

ui/src/app/base/components/Pager/Pager.js
```diff
@@ -0,0 +1,132 @@
...
+    const lastPage = pageNumbers.length;
+    const truncated = lastPage > truncateThreshold;
+
+    let visiblePages;
+    if (truncated) {
+      let start = currentPage - 2;
+      let end = currentPage + 1;
+      if (currentPage === 1) {
```
huwshimi: Could be helpful to leave some comments about what these ifs are doing.

ui/src/app/base/components/Pager/Pager.test.js
```diff
@@ -0,0 +1,94 @@
+import { mount } from "enzyme";
+import React from "react";
+
+import Pager from "./Pager";
+
+describe("<Pager />", () => {
+  // snapshot tests
+  it("renders and matches the snapshot", () => {
+    const component = mount(
```
huwshimi: I think these can all be shallow.

ui/src/app/base/components/PagerButton/PagerButton.test.js
```diff
@@ -0,0 +1,34 @@
...
+    expect(component).toMatchSnapshot();
+  });
+
+  // unit tests
+  it("sets the appropriate classes and text if direction is 'back'", () => {
+    const component = shallow(
+      <PagerButton direction="back" onClick={jest.fn()} />
+    );
+
+    expect(component.find("button").hasClass("p-pagination__link--previous"));
```
huwshimi: How did I not know about `hasClass`?!

ui/src/app/base/components/Pager/Pager.js
squidsoup: Yep, good call.

ui/src/app/base/components/Pager/Pager.js
squidsoup: indeed...

ui/src/app/base/components/Pager/Pager.test.js
squidsoup: Quite right, thanks, just had to change an assertion for the rendered class to `PagerItem`.

PR #62, opened 2019-08-07
Create Network component with Proxy section

Caleb-Ellis:
## Done
- Created Network and ProxyForm components
- Added config selectors for proxy-related things
- Added url validation to proxy url

## QA
- Go to /MAAS/settings/network
- Check that the proxy form loads in
- Check that clicking "External" or "Peer" displays a text input below
- Check that validation for the proxy url works (i.e. error message shows if url invalid or empty, save button doesn't work)
- Check that the save button shows an alert with the correct values

Fixes canonical-web-and-design/MAAS-squad#1394

### DIFF
--- a/ui/src/app/settings/containers/Network/Network.js
+++ b/ui/src/app/settings/containers/Network/Network.js
@@ -1,5 +1,32 @@
 import React from "react";
+import { useSelector } from "react-redux";
 
-const Network = () => <h4>Network</h4>;
+import selectors from "app/settings/selectors";
+import Col from "app/base/components/Col";
+import Loader from "app/base/components/Loader";
+import ProxyForm from "app/settings/containers/ProxyForm";
+import Row from "app/base/components/Row";
+
+const Network = () => {
+  const loaded = useSelector(selectors.config.loaded);
+  const loading = useSelector(selectors.config.loading);
+
+  return (
+    <>
+      <h4>
+        Network
+        {loading && <Loader text="Loading..." inline />}
+      </h4>
+      {loaded && (
+        <Row>
+          <Col size={8}>
+            <h5>Proxy</h5>
+            <ProxyForm />
+          </Col>
+        </Row>
+      )}
+    </>
+  );
+};
 
 export default Network;
--- a/ui/src/app/settings/containers/ProxyForm/ProxyForm.js
+++ b/ui/src/app/settings/containers/ProxyForm/ProxyForm.js
@@ -1,0 +1,51 @@
+import { Formik } from "formik";
+import React from "react";
+import { useSelector } from "react-redux";
+import * as Yup from "yup";
+
+import selectors from "app/settings/selectors";
+import ProxyFormFields from "app/settings/components/ProxyFormFields";
+
+const ProxySchema = Yup.object().shape({
+  proxyType: Yup.string().required(),
+  httpProxy: Yup.string().when("proxyType", {
+    is: val => val === "externalProxy" || val === "peerProxy",
+    then: Yup.string()
+      .url("Must be a valid URL.")
+      .required("Please enter the proxy URL.")
+  })
+});
+
+const ProxyForm = () => {
+  const httpProxy = useSelector(selectors.config.httpProxy);
+  const proxyType = useSelector(selectors.config.proxyType);
+
+  return (
+    <Formik
+      initialValues={{
+        httpProxy,
+        proxyType
+      }}
+      onSubmit={(values, { setSubmitting }) => {
+        const { httpProxy, proxyType } = values;
+        const enable_http_proxy = proxyType !== "noProxy";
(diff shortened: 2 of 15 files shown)


ui/src/app/settings/components/ProxyForm/index.js
```diff
@@ -0,0 +1 @@
+export { default } from "./ProxyForm";
```
steverydz: These seems like a strange pattern to me. I know it's other places in the app, but why not just export and import from the component directly?

Caleb-Ellis: Yeah it is a bit weird. I remember trying to do something similar in vanilla react to get tree-shaking to work properly (which it never did), but that was a lib and this is an app. Maybe it makes sense if there are multiple related, exportable files within the same folder, e.g.

#### `exporter.js`
``` javascript
export { Title } from "./Title";
export { Body } from "./Body";
export { Footer } from "./Footer";
```

#### `importer.js`
``` javascript
import { Title, Body, Footer } from "./exporter.js"

Comment (squidsoup) [reply ui/src/app/settings/components/ProxyForm/index.js]: @steverydz because then your import looks like import { Table } from 'components/Table/Table'

File: ui/src/app/settings/containers/Network/Network.js [hunk 2]
```diff
@@ -1,5 +1,33 @@
...
+import selectors from "app/settings/selectors";
+import Loader from "app/base/components/Loader";
+import ProxyForm from "app/settings/components/ProxyForm";
+
+const Network = () => {
+  const loaded = useSelector(selectors.config.loaded);
+  const loading = useSelector(selectors.config.loading);
+
+  const httpProxy = useSelector(selectors.config.httpProxy);
+  const proxyType = useSelector(selectors.config.proxyType);
```
squidsoup: Please don't pass state down to children, just call `useSelector` in the child. This is particularly good in the hooks world as state is now encapsulated (and coupled...) in a component which means any props a component defined are actual props, not redux state.

Caleb-Ellis: @squidsoup Ahh right, I've still got the muscle memory to pass down props from one parent connected component. In that case maybe we _should_ stop with container/component naming and maybe instead go with a views/routes structure instead, which imo is much easier to traverse since it's reflective of the DOM. (sidenote - what's an easy way to generate a gh markdown file directory?)
```bash
...
├── views/
│   ├── Images/
│   ├── Network/
│   │   ├── NetworkContainer/
│   │   │   ├── index.js
│   │   │   └── NetworkContainer.js
│   │   ├── ProxyForm/
│   │   │   ├── index.js
│   │   │   └── ProxyForm.js
│   │   ├── DNSForm/
│   │   │   ├── index.js
│   │   │   └── DNSForm.js
...
```

Caleb-Ellis:
@squidsoup moved ProxyForm to `/containers` and used the selectors there. Still, interested to hear your thoughts on the views directory structure. :)

ui/src/app/settings/containers/Network/Network.js
```diff
@@ -1,5 +1,30 @@
...
+  const loading = useSelector(selectors.config.loading);
+
+  return (
+    <>
+      <h4>
+        Network
+        {loading && <Loader text="Loading..." inline />}
+      </h4>
+      {loaded && (
+        <div className="row">
```
squidsoup: Please use the `Row` and `Col` grid components here.

ui/src/app/settings/containers/ProxyForm/ProxyForm.js
```diff
@@ -0,0 +1,128 @@
...
+    >
+      {({
+        handleBlur,
+        handleChange,
+        handleSubmit,
+        isSubmitting,
+        touched,
+        errors,
+        values
+      }) => (
```
huwshimi: This is essentially creating a new component for these fields. I ended up splitting this bit out to make it more easily to test. See: https://github.com/canonical-web-and-design/maas-ui/blob/master/ui/src/app/settings/components/UserForm/UserForm.js#L79 and the fields component: https://github.com/canonical-web-and-design/maas-ui/blob/master/ui/src/app/settings/components/UserFormFields/UserFormFields.js.

ui/src/app/settings/containers/ProxyForm/ProxyForm.js
```diff
@@ -0,0 +1,128 @@
...
+        <Form onSubmit={handleSubmit}>
+          <Input
```
huwshimi: I created a new component for handling Formik fields to reduce most of the repetition below. See: https://github.com/canonical-web-and-design/maas-ui/blob/master/ui/src/app/settings/components/UserFormFields/UserFormFields.js#L22.

Caleb-Ellis:
@steverydz updated so that the button is disabled until a field is touched, or if the form is submitting.

PR #63, opened 2019-08-07
Add save functionality to Storage form + build ActionButton

Caleb-Ellis:
## Done
- Added save functionality to Storage form
- Reworked Storage form to use FormikField component
- Updated Storage form styling (no title, left-aligned button)
- Added ActionButton component which wraps Button. It takes props to determine whether it's in a loading or success state. It takes an explicit width prop which is a bit crap but I couldn't figure out how to store button width in the hooks world and figured this works fine for now :man_shrugging: 
- Added some hooks that essentially "fake" a delay after clicking Save, even though saving happens almost instantly and therefore saving state is `false`. Success state is simply determined as "saving state moved from true => false" so will probs need some handling for when that happens due to an error.

## QA
- Go to Storage and check that the button is disabled
- Change some values and check that button is no longer disabled
- Change values back to original (without saving) and check that button becomes disabled again
- Save some changes and check that the button text is replaced by a spinner, and then after a short time it changes to a tick
- Refresh the page and check that state has persisted

### No values changed
Screen Shot 2019-08-12 at 18 51 44-fullpage

### Values changed
Screen Shot 2019-08-12 at 18 51 54-fullpage

### Saving
Screen Shot 2019-08-12 at 18 52 00-fullpage

### Saved
Screen Shot 2019-08-12 at 18 52 01-fullpage

### DIFF
--- a/ui/src/app/base/components/ActionButton/ActionButton.js
+++ b/ui/src/app/base/components/ActionButton/ActionButton.js
@@ -1,0 +1,60 @@
+import classNames from "classnames";
+import PropTypes from "prop-types";
+import React from "react";
+
+import Button from "../Button";
+import "./ActionButton.scss";
+
+const generateButtonChildren = (children, appearance, loading, success) => {
+  const iconClass = classNames({
+    "p-icon--spinner u-animation--spin": loading,
+    "is-light": appearance === "positive" || appearance === "negative",
+    "p-icon--success":
+      success && (appearance !== "positive" && appearance !== "negative"),
+    "p-icon--success-positive": success && appearance === "positive",
+    "p-icon--success-negative": success && appearance === "negative"
+  });
+
+  return loading || success ? <i className={iconClass} /> : children;
+};
+
+const ActionButton = ({
+  appearance = "neutral",
+  children,
+  className,
+  loading,
+  success,
+  width,
+  ...props
+}) => {
+  const classes = classNames(className, `p-action-button`, {
+    "is-loading": loading,
+    "is-success": success
+  });
+  const buttonChildren = generateButtonChildren(
+    children,
+    appearance,
+    loading,
+    success
+  );
+  const style = width ? { width } : {};
+
+  return (
+    <Button
+      appearance={appearance}
+      className={classes}
+      style={style}
+      {...props}
+    >
+      {buttonChildren}
+    </Button>
+  );
+};
+
+ActionButton.propTypes = {
+  className: PropTypes.string,
+  loading: PropTypes.bool,
+  success: PropTypes.bool
+};
+
+export default ActionButton;
--- a/ui/src/app/base/hooks.js
+++ b/ui/src/app/base/hooks.js
@@ -53,7 +53,7 @@ export const useSettingsSave = (saving, setLoading, setSuccess) => {
   const prevSaving = usePrevious(saving);
 
   // Temporarily set success to true if saving changes from true to false
-  // TODO: Add a check for errors too
+  // TODO: Add a check for errors too: https://github.com/canonical-web-and-design/maas-ui/issues/39
   useEffect(() => {
     if (!prevSaving && saving) {
       setLoading(true);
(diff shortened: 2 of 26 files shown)


ui/src/app/settings/actions/config.js
```diff
@@ -13,4 +13,18 @@ config.fetch = () => {
...
 };
 
+config.saveConfig = (name, value) => {
+  return {
+    type: "WEBSOCKET_SEND",
+    payload: {
+      actionType: "SAVE_CONFIG",
+      message: {
+        method: "config.update",
+        params: { name, value },
```
squidsoup: To answer your first question TODO question, fairly certain this can be an array.

Caleb-Ellis: @squidsoup Hmm yeah I wasn't sure about that. This is the update method in the config websocket handler, which seems to expect a dict with name and value keys:

``` python
    def update(self, params):
        """Update a config value."""
        if not self.user.is_superuser:
            raise HandlerPermissionError()
        if 'name' not in params:
            raise HandlerPKError("Missing name in params")
        if 'value' not in params:
            raise HandlerValidationError("Missing value in params")
        name = params['name']
        value = params['value']
        form = get_config_form(name, {name: value})
        if form.is_valid():
            try:
                request = HttpRequest()
                request.user = self.user
                form.save(ENDPOINT.UI, request)
            except ValidationError as e:
                self._fix_validation_error(name, e.error_dict)
                raise HandlerValidationError(e.error_dict)
            return self._include_choice({
                'name': name,
                'value': Config.objects.get_config(name),
            })
        else:
            self._fix_validation_error(name, form.errors)
            raise HandlerValidationError(form.errors)
```

Unless I'm missing something (highly possible)?

ui/src/app/settings/containers/Storage/Storage.js
```diff
@@ -42,76 +69,70 @@ const Storage = () => {
...
-            handleChange,
-            handleSubmit,
-            isSubmitting,
-            values
-          }) => (
+          {({ handleBlur, handleChange, handleSubmit, touched, values }) => (
             <Form onSubmit={handleSubmit}>
               <div className="row">
                 <div className="col-8">
                   <Select
```
huwshimi: These can use the new Formik field component, see: https://github.com/canonical-web-and-design/maas-ui/pull/62#discussion_r312314179

ui/src/app/settings/actions/config.js
squidsoup: Yep, the API is a bit broken here, but we should not be dispatching multiple actions, the payload should still be an array - we'll just have to handle that in the saga.

squidsoup: Have filed [a bug](https://bugs.launchpad.net/maas/+bug/1839788).

squidsoup: @Caleb-Ellis support for [params as an array landed](https://github.com/canonical-web-and-design/maas-ui/commit/d71059c4b9709ae0b50c91c4cb7428322f43fb0e).

lilyvidenova:
@Caleb-Ellis Could you try increasing the time of the active state by 100ms? I feel like it's a little bit too quick.

ui/src/app/base/hooks.js
```diff
@@ -35,3 +38,40 @@ export const useLocation = () => {
...
+  return ref.current;
+};
+
+export const useSettingsSave = (saving, setLoading, setSuccess) => {
+  const delayTimer = useRef();
+  const saveTimer = useRef();
+  const prevSaving = usePrevious(saving);
+
+  // Temporarily set success to true if saving changes from true to false
+  // TODO: Add a check for errors too
```
steverydz: Is there a ticket for this? Might be good to put a link to it here

ui/src/app/base/components/ActionButton/ActionButton.js
```diff
@@ -0,0 +1,60 @@
...
+  return loading || success ? <i className={iconClass} /> : children;
+};
+
+const ActionButton = ({
+  appearance = "neutral",
+  children,
+  className,
+  loading,
+  success,
+  width,
```
steverydz: Why would we need to define a width? Shouldn't this be defined by the text?

Caleb-Ellis:
@lilyvidenova Fixed. Changed form to col-6, moved icon to help text, removed yellow border

Screen Shot 2019-08-13 at 11 48 03-fullpage

PR #64, opened 2019-08-07
Create settings configuration general form

steverydz:
## Done
- Created general component. Fixes: https://github.com/canonical-web-and-design/MAAS-squad/issues/1379
- Displayed initial values in general form. Fixes: https://github.com/canonical-web-and-design/MAAS-squad/issues/138.
- Implemented saving general form. Fixes: https://github.com/canonical-web-and-design/MAAS-squad/issues/1382

## QA
- Go to `/settings/configuration#general`
- See that there is a field for "MAAS name" and a checkbox to enable analytics
- Remove the value from the "MAAS name" field and see that the "Save" button is disabled
- Change the values of both the form fields
- Submit the form
- See that as you are saving the "Save" button indicates that it is saving
- Reload the page and see that the values you changed have persisted


ui/src/app/settings/components/GeneralForm/GeneralForm.js
```diff
@@ -0,0 +1,72 @@
...
+            type="checkbox"
+            value={values.enableAnalytics}
+            checked={values.enableAnalytics}
+          />
+          <Button
+            appearance="positive"
+            type="submit"
+            disabled={isSubmitting || !values.maasName}
+          >
+            Save
```
anthonydillon: Does this button component provide user feedback via a callback?

anthonydillon:
Formik mean we don't need to test validation?

ui/src/app/settings/containers/Configuration/Configuration.js
```diff
@@ -1,5 +1,29 @@
...
+import { useSelector } from "react-redux";
 
-const Configuration = () => <h4>Configuration</h4>;
+import selectors from "app/settings/selectors";
+import Loader from "app/base/components/Loader";
+import GeneralForm from "app/settings/components/GeneralForm";
+
+const Configuration = () => {
+  const loaded = useSelector(selectors.config.loaded);
+  const loading = useSelector(selectors.config.loading);
```
squidsoup: Hi @steverydz, please don't pass state down into children, instead, call selectors where the state is used (in `GeneralForm`).

Have a look at [Should I only connect my top component, or can I connect multiple components in my tree?](https://redux.js.org/faq/react-redux#should-i-only-connect-my-top-component-or-can-i-connect-multiple-components-in-my-tree)

ui/src/app/settings/selectors/config.js
```diff
@@ -98,4 +98,26 @@ config.diskEraseWithQuick = createSelector(
...
 );
 
+/**
+ * Returns the MAAS config for MAAS name.
+ * @param {Object} - The redux state.
+ * @returns {String} Then MAAS name.
+ */
+config.maasName = createSelector(
+  [config.all],
+  configs => {
```
squidsoup: With ES6, you can just say (as you do below):

```
configs => getValueFromName(configs, "maas_name");
```

ui/src/app/settings/components/GeneralForm/GeneralForm.js
squidsoup: No, this is a standard vanilla button. I'd like to see the ActionButton we use in rbac moved to the shared library though.

ui/src/app/settings/components/GeneralForm/GeneralForm.test.js
```diff
@@ -0,0 +1,15 @@
...
+import GeneralForm from "./GeneralForm";
+
+describe("GeneralForm", () => {
+  let defaultProps = {
+    maasName: "bionic-maas",
+    enableAnalytics: true
+  };
+
+  it("can render", () => {
+    const wrapper = mount(<GeneralForm {...defaultProps} />);
```
squidsoup: Please use `shallow` here instead, mounting the entire component tree makes the snapshots huge and of questionable usefulness.

ui/src/app/settings/selectors/config.js
steverydz: ah yes, I think I had a longer version initially and prettier changed it, I forgot to change it back

steverydz:
> Formik mean we don't need to test validation?

We should probably test that we're giving all the right things, but we need to be mindful of testing the library itself.

ui/src/app/settings/actions/config.js
```diff
@@ -13,4 +13,18 @@ config.fetch = () => {
   };
 };
 
+config.save = (name, value) => {
```
squidsoup: This needs to be an array - dispatching an action for each keypair would mean we make a ws request for every form field.

ui/src/app/settings/components/GeneralForm/GeneralForm.js
```diff
@@ -0,0 +1,90 @@
...
+          onSubmit={(values, { setSubmitting }) => {
+            setTimeout(() => {
+              Object.keys(values).forEach(key => {
+                dispatch(saveConfig(key, values[key]));
+              });
+              setSubmitting(false);
+            }, 500);
+          }}
+        >
+          {({
```
huwshimi: You could consider moving this into a component, see: https://github.com/canonical-web-and-design/maas-ui/pull/62#discussion_r312313966.

ui/src/app/settings/components/GeneralForm/GeneralForm.js
```diff
@@ -0,0 +1,90 @@
...
+            errors,
+            isSubmitting,
+            touched,
+            handleSubmit,
+            handleChange,
+            handleBlur,
+            values
+          }) => (
+            <Form onSubmit={handleSubmit}>
+              <Input
```
huwshimi: These can use the new Formik field component, see: https://github.com/canonical-web-and-design/maas-ui/pull/62#discussion_r312314179

ui/src/app/settings/actions/config.js
steverydz: I agree that would be preferable, but it doesn't seem possible. I've tried sending this as an array and the app crashes. Looking at the websocket handler I came to the same conclusion as @Caleb-Ellis in [his comment here](https://github.com/canonical-web-and-design/maas-ui/pull/63#discussion_r311912068).

ui/src/app/settings/components/GeneralForm/GeneralForm.js
steverydz: Done

ui/src/app/settings/components/GeneralForm/GeneralForm.js
steverydz: I've tried following this pattern but keep running into issues. Seems simpler to keep it as it is.

steverydz: What is the benefit of using the formik field?

It seems to wrap an input component which then wraps the input in a field component? That seems like a lot just to output an input.

Also why do we need to pass the data for all fields into each formik field?

ui/src/app/settings/actions/config.js
squidsoup: Not just preferable, a requirement, so this branch will need to hold until that is fixed.

ui/src/app/settings/components/GeneralForm/GeneralForm.js
squidsoup: > I've tried following this pattern but keep running into issues. Seems simpler to keep it as it is.

Please let us know what those issues are, we can't help resolve them otherwise.

squidsoup: > What is the benefit of using the formik field?
> 
> It seems to wrap an input component which then wraps the input in a field component? That seems like a lot just to output an input.
> 
> Also why do we need to pass the data for all fields into each formik field?

Management of form state, please read the formik docs.

ui/src/app/settings/actions/config.test.js
```diff
@@ -13,4 +13,18 @@ describe("config actions", () => {
       }
     });
   });
+
+  it("should handle saving config", () => {
+    expect(config.save("maas_name", "bionic-maas")).toEqual({
+      type: "WEBSOCKET_SEND",
+      payload: {
+        actionType: "SAVE_CONFIG",
```
squidsoup: Let's call this `UPDATE_CONFIG` to be consistent with the API.

ui/src/app/settings/actions/config.js
```diff
@@ -13,4 +13,18 @@ config.fetch = () => {
...
+  return {
+    type: "WEBSOCKET_SEND",
+    payload: {
+      actionType: "SAVE_CONFIG",
```
squidsoup: Let's call this `UPDATE_CONFIG` to be consistent with the API.

ui/src/app/settings/actions/config.js
squidsoup: Support for [params as an array has landed](https://github.com/canonical-web-and-design/maas-ui/commit/d71059c4b9709ae0b50c91c4cb7428322f43fb0e), so this can now be `config.save = params =>`

ui/src/app/settings/components/GeneralForm/GeneralForm.js
```diff
@@ -0,0 +1,95 @@
...
+import FormikField from "app/base/components/FormikField";
+
+const GeneralSchema = Yup.object().shape({
+  maas_name: Yup.string().required(),
+  enable_analytics: Yup.boolean()
+});
+
+const GeneralForm = () => {
+  const dispatch = useDispatch();
+  const saveConfig = actions.config.save;
```
squidsoup: `updateConfig` please

ui/src/app/settings/actions/config.js
squidsoup: Best to manipulate the array here (so that it has the form `[{name: 'a': value: 'b'}, {...}]`, rather than the calling code.

ui/src/app/settings/components/GeneralForm/GeneralForm.js
steverydz: But why can't we do that directly in the input component? The formik wrapper seems like an unnecessary layer of abstraction.

squidsoup: Having a look at the `FormikField` component, my answer would be to reduce duplication, given every field needs the same handlers and error object from formik props. @huwshimi may have more to add.

ui/src/app/settings/selectors/config.js
```diff
@@ -131,4 +131,24 @@ config.proxyType = createSelector(
...
+  [config.all],
+  configs => getValueFromName(configs, "maas_name")
+);
+
+/**
+ * Returns the MAAS config for enable analytics.
+ * @param {Object} - The redux state.
+ * @returns {Boolean} Enable analytics.
+ */
+config.enableAnalytics = createSelector(
```
squidsoup: `config.analyticsEnabled` might be better, `config.enableAnalytics` sounds like a function rather than a property.

ui/src/app/settings/actions/config.js
steverydz: done

ui/src/app/settings/actions/config.js
steverydz: done

ui/src/app/settings/components/GeneralForm/GeneralForm.js
steverydz: done

ui/src/app/settings/selectors/config.js
steverydz: done

steverydz:
@squidsoup all requested changes done

ui/src/app/settings/actions/config.js
```diff
@@ -13,4 +13,25 @@ config.fetch = () => {
   };
 };
 
+config.update = params => {
+  const configParams = Object.keys(params).map(key => {
```
Caleb-Ellis: Hmm I don't think we should be doing any data manipulation here. Generally redux actions are supposed to be dumb, i.e. take a payload => put it in an action object. So I would map the params object before it gets dispatched.

ui/src/app/settings/components/GeneralForm/GeneralForm.js
```diff
@@ -0,0 +1,96 @@
...
+            maas_name: maasName,
+            enable_analytics: analyticsEnabled
+          }}
+          validationSchema={GeneralSchema}
+          onSubmit={(values, { setSubmitting }) => {
+            setTimeout(() => {
+              dispatch(updateConfig(values));
+              // Object.keys(values).forEach(key => {
+              //   dispatch(updateConfig(key, values[key]));
+              // });
```
Caleb-Ellis: Remove comments

ui/src/app/settings/components/GeneralForm/GeneralForm.js
```diff
@@ -0,0 +1,96 @@
...
+              setSubmitting(false);
+            }, 500);
```
Caleb-Ellis: This timeout is fine for now, but just so you know in https://github.com/canonical-web-and-design/maas-ui/pull/63 I'll create an ActionButton component that has a success state. Also the action is taken immediately, only the UI is faked to make it seem like it's still saving.

ui/src/app/settings/components/GeneralForm/GeneralForm.js
```diff
@@ -0,0 +1,96 @@
...
+                <FormikField
+                  fieldKey="maas_name"
+                  label="MAAS name"
+                  required={true}
+                  type="text"
+                  formikProps={formikProps}
+                />
+                <FormikField
+                  fieldKey="enable_analytics"
+                  label="Enable Google Analytics in MAAS UI to shap improvements in user experience"
```
Caleb-Ellis: typo "shap"

steverydz: fixed

ui/src/app/settings/components/GeneralForm/GeneralForm.js
steverydz: yeah, I was waiting on that before getting too involved with this button

ui/src/app/settings/components/GeneralForm/GeneralForm.js
steverydz: dammit

Issue #53, opened 2019-08-05
Table contents could be more responsive


squidsoup:
At lower window widths, table cell contents overflow. 

e.g. email in the users table

![image](https://user-images.githubusercontent.com/130286/62435128-3346e000-b78f-11e9-91bc-d80bafefbf17.png)

PR #67, opened 2019-08-08
Add a delete confirm to the user list

huwshimi:
Done:
- Show a delete confirmation when deleting user from the user list. Fixes: https://github.com/canonical-web-and-design/maas-squad/issues/1354. Fixes: https://github.com/canonical-web-and-design/maas-ui/issues/53.

QA:
- View the list of users.
- Click the delete icon next to a user.
- A confirmation should appear.
- Click cancel, it should disappear.

### DIFF
--- a/ui/src/app/base/components/Loader/Loader.scss
+++ b/ui/src/app/base/components/Loader/Loader.scss
@@ -1,4 +1,4 @@
-@import "src/scss/vanilla";
+@import "~vanilla-framework/scss/settings_spacing";
 
 .p-icon__text {
   margin-left: $sph-inner--small;
--- a/ui/src/app/settings/containers/Users/Users.js
+++ b/ui/src/app/settings/containers/Users/Users.js
@@ -3,6 +3,7 @@ import { Link } from "react-router-dom";
 import PropTypes from "prop-types";
 import React, { useState } from "react";
 
+import "./Users.scss";
 import { useFetchOnce } from "app/base/hooks";
 import actions from "app/settings/actions";
 import selectors from "app/settings/selectors";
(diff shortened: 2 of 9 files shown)


ui/src/app/settings/containers/Users/Users.scss
```diff
@@ -0,0 +1,29 @@
...
+  padding-left: 0.5rem;
+  padding-right: 0.5rem;
+}
+
+.user-list .p-table-expanding__panel {
+  padding-top: 0;
+}
+
+.p-table-expanding th.user-list__cell--small,
+.p-table-expanding td.user-list__cell--small {
```
anthonydillon: Could this modifier be added to the table-expanding pattern scss. If someone needs to modify this styles the first place they will look will be the pattern file.

ui/src/app/settings/containers/Users/Users.scss
```diff
@@ -0,0 +1,29 @@
...
+  flex: 0 0 6rem;
+}
+
+.user-list__expanded-row {
+  background-color: #fff;
```
anthonydillon: This is also something we will want across all tables. Could you add to the table pattern locally here. For example: `p-table__row.is-active` then we could propose this to Vanilla.

ui/src/app/settings/containers/Users/Users.js
```diff
@@ -14,54 +15,85 @@ import Pagination from "app/base/components/Pagination";
...
+          content: <Link to={`/users/${user.id}`}>{user.username}</Link>,
+          role: "rowheader"
+        },
+        { content: user.email },
+        { content: 37, className: "u-align--right user-list__cell--small" },
+        { content: "Local", className: "user-list__cell--small" },
+        { content: "12 mins ago" },
+        {
+          content: user.is_superuser ? "Admin" : null,
+          className: "user-list__cell--small"
```
anthonydillon: In my experience design want more control over column widths which lead use to setting custom widths to all table columns. Which you can see here: https://github.com/maas/maas/blob/master/src/maasserver/static/scss/_tables.scss

ui/src/app/settings/containers/Users/Users.js
```diff
@@ -14,54 +15,85 @@ import Pagination from "app/base/components/Pagination";
...
+        },
+        {
+          content: user.sshkeys_count,
+          className: "u-align--right user-list__cell--small"
+        },
+        {
+          content: (
+            <>
```
anthonydillon: Is this meant to have an empty element?

ui/src/app/settings/containers/Users/Users.js
```diff
@@ -14,54 +15,85 @@ import Pagination from "app/base/components/Pagination";
...
+          className: "u-align--right user-list__cell--small"
+        }
+      ],
+      expanded: expanded,
+      expandedContent: expanded && (
+        <div className="user-list__expanded-content">
+          <Row>
+            <Col size="8">
+              Are you sure you want to delete user "{user.username}"?{" "}
+              <span className="u-text--light">
```
anthonydillon: This seems like the main message. I don't think we should be making it subtle.

ui/src/app/settings/containers/Users/Users.scss
anthonydillon: Could you use Vanilla colour variables please. `$color-x-light`

ui/src/app/settings/containers/Users/Users.js
```diff
@@ -14,54 +15,85 @@ import Pagination from "app/base/components/Pagination";
...
-          </>
-        ),
-        className: "u-align--right"
+                <i className="p-icon--edit">Edit</i>
+              </Button>
+              {user.id !== authUser.id && (
+                <Button
+                  appearance="base"
+                  className="is-small user-list__button"
+                  onClick={() => setExpandedId(user.id)}
```
anthonydillon: Your giving each row a lot of power here. They shouldn't have the ability to set other row. Maybe think about moving this state to the table component itself? And then just passing a click function to the rows?

ui/src/app/settings/containers/Users/Users.js
```diff
@@ -3,6 +3,7 @@ import { Link } from "react-router-dom";
 import PropTypes from "prop-types";
 import React, { useState } from "react";
 
+import "./Users.scss";
```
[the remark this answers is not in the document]
huwshimi: This is a great way of managing css specific to a component (https://facebook.github.io/create-react-app/docs/adding-a-stylesheet) and within the SCSS file you can import the vanilla bits you need `@import "~vanilla-framework/scss/settings_colors";`.

In this case, you're right in that these styles should not be specific to the user list.

ui/src/app/settings/containers/Users/Users.scss
```diff
@@ -0,0 +1,29 @@
+@import "src/scss/vanilla";
+
+.user-list__button.p-button--base {
```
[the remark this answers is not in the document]
huwshimi: Thanks, I've made this a global rule and filed: https://github.com/canonical-web-and-design/vanilla-framework/issues/2463

ui/src/app/settings/containers/Users/Users.scss
```diff
@@ -0,0 +1,29 @@
...
+  // Stop the icons from adding extra height to the row:
+  margin-top: -0.25rem;
+  margin-bottom: -0.25rem;
+  padding-left: 0.5rem;
+  padding-right: 0.5rem;
+}
+
+.user-list .p-table-expanding__panel {
+  padding-top: 0;
```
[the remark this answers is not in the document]
huwshimi: You're right, this should be for all the tables on a dark background.

ui/src/app/settings/containers/Users/Users.scss
huwshimi: Really I think we need more control over the column widths than this from looking at the design, do you have any suggestions about a good way to set the width of each column?

huwshimi: Oops, you answered below.

ui/src/app/settings/containers/Users/Users.js
huwshimi: This is shorthand for `React.Fragment` which can be used when you need to wrap multiple elements but don't need anything to appear in the DOM.

ui/src/app/settings/containers/Users/Users.js
huwshimi: This matches the design here: https://app.zeplin.io/project/5ce5423a25863c1e0d7abb27/screen/5d36e2233615fbb2789ce87f.

Sorry just realised I didn't link to the design in the PR.

ui/src/app/settings/containers/Users/Users.js
huwshimi: There are a couple of issues there, one is that we don't have a consistent element that opens the row (in this case it is the trash icon).

The second is that we need to be able to control that state outside the table. For this list when the user confirms the delete then we need to close that row.

ui/src/app/settings/containers/Users/Users.scss
huwshimi: Done.

ui/src/app/settings/containers/Users/Users.js
huwshimi: Done, thanks for that, I've updated these to match the design.

ui/src/app/base/components/Loader/Loader.scss
```diff
@@ -1,4 +1,4 @@
-@import "src/scss/vanilla";
+@import "~vanilla-framework/scss/settings_spacing";
```
huwshimi: These are all drive-by fixes to only include the vanilla bits we need.

squidsoup:
QA +1, thanks @huwshimi!

PR #68, opened 2019-08-08
Dynamically build websocket url if REACT_APP_WEBSOCKET_URL is undefined.

squidsoup:
## Done
Dynamically build websocket url if REACT_APP_WEBSOCKET_URL is undefined. This is useful in production (and in maas devel) where the client is served from the same host/port as the server.


ui/src/app/base/sagas/websockets.js
```diff
@@ -4,12 +4,33 @@ import { all, call, put, take, race } from "redux-saga/effects";
 import getCookie from "./utils";
 import WebSocketClient from "../../../websocket-client";
 
+/**
+ * Dynamically build a websocket url from window.location
+ *
+ */
+const buildWsUrl = () => {
+  let protocol;
```
huwshimi: If you wished: `const protocol = window.location.protocol === "https:" ? "wss:" : "ws:"`.

ui/src/app/base/sagas/websockets.js
```diff
@@ -4,12 +4,33 @@ import { all, call, put, take, race } from "redux-saga/effects";
 import getCookie from "./utils";
 import WebSocketClient from "../../../websocket-client";
 
+/**
+ * Dynamically build a websocket url from window.location
+ *
```
huwshimi: Blank line, could have a `@returns` here.

ui/src/app/base/sagas/websockets.js
```diff
@@ -4,12 +4,33 @@ import { all, call, put, take, race } from "redux-saga/effects";
...
  * Create a WebSocket connection via the client.
  */
 export function createConnection(csrftoken) {
   return new Promise((resolve, reject) => {
-    const url = `${process.env.REACT_APP_WEBSOCKET_URL}?csrftoken=${csrftoken}`;
+    let url;
+    if (process.env.REACT_APP_WEBSOCKET_URL) {
+      url = `${process.env.REACT_APP_WEBSOCKET_URL}?csrftoken=${csrftoken}`;
+    } else {
+      url = buildWsUrl();
```
huwshimi: Does this also need the `?csrftoken=` bit?

squidsoup: Oops, it does indeed. thanks.

ui/src/app/base/sagas/websockets.js
squidsoup: thanks

PR #69, opened 2019-08-08
Display a delete confirmation on user details

huwshimi:
Done:
- Show a delete confirmation when deleting user from a user details page. Fixes: https://github.com/canonical-web-and-design/maas-squad/issues/1355

QA:
- View the list of users and click on a user.
- Click the delete button.
- A confirmation should appear.
- Click cancel, it should disappear.


ui/src/app/base/components/Modal/Modal.js
```diff
@@ -0,0 +1,36 @@
...
+  <div className="p-modal" onClick={close}>
+    <div
+      className="p-modal__dialog"
+      role="dialog"
+      aria-labelledby="modal-title"
+      aria-describedby="modal-description"
+      onClick={evt => evt.stopPropagation()}
+    >
+      <header className="p-modal__header">
+        <h2 className="p-modal__title" id="modal-title">
```
anthonydillon: Do you need id's here. That limits the use to one per view.

ui/src/app/base/components/Modal/Modal.js
```diff
@@ -0,0 +1,36 @@
...
+        </h2>
+        <button
+          className="p-modal__close"
+          aria-label="Close active modal"
+          onClick={close}
+        >
+          Close
+        </button>
+      </header>
+      <div id="modal-description">{children}</div>
```
anthonydillon: Same here

ui/src/app/settings/containers/User/User.js
```diff
@@ -36,8 +39,8 @@ const User = () => {
           </Button>
           <Button
             appearance="negative"
-            element={Link}
-            to={`/users/${user.id}/delete`}
+            onClick={openPortal}
+            style={{ marginLeft: "1rem" }}
```
anthonydillon: Could you file an issue on Vanilla if the default doesn't work for this case.

ui/src/scss/base.scss
```diff
@@ -31,3 +31,15 @@ body {
   @extend %icon;
   @include maas-icon-edit($color-mid-dark);
 }
+
+.u-text--light {
```
anthonydillon: This class isn't used

ui/package.json
```diff
@@ -16,6 +16,7 @@
     "react-router-hash-link": "1.2.2",
     "react-router-prop-types": "1.0.4",
     "react-scripts": "3.0.1",
+    "react-useportal": "^0.1.46",
```
squidsoup: nice!

ui/src/app/settings/containers/User/User.js
huwshimi: Yep, it's this one: https://github.com/canonical-web-and-design/vanilla-framework/issues/2408, we'll just have to remember to update this when that gets fixed.

ui/src/app/base/components/Modal/Modal.js
huwshimi: Good point. The ids are used by the aria attributes above, I've updated the component to generate unique ids. I've also filed: https://github.com/canonical-web-and-design/maas-ui/issues/72 to update the form field ids to do the same.

ui/src/scss/base.scss
huwshimi: It's used here: https://github.com/canonical-web-and-design/maas-ui/pull/69/files#diff-1ee9d1795a53072cc392e1124036416bR89.

PR #83, opened 2019-08-12
update setting up instructions based on feedback

steverydz:
## Done
Update the setup instructions in the README based on feedback from people that have been using it

## QA
Read through and see that it makes sense

### DIFF
--- a/ui/README.md
+++ b/ui/README.md
@@ -1,22 +1,41 @@
 # Setting up MAAS UI
 
-**Note**: You will need an instance of MAAS running in the same container in order to run MAAS UI.
+**Note: You will need an instance of MAAS running in order to run MAAS UI.**
 
-**Note**: You will need your public and private SSH keys inside your container to be able to clone the repo. For help with this see [ssh-copy-id](https://www.ssh.com/ssh/copy-id).
+## Generate an SSH key in your container (LXD and multipass)
 
-Inside your MAAS container (either LXD or multipass) go to the parent directory of your MAAS installation, clone the MAAS UI repo and set it's upstream and origin:
+**Note: The following instructions should be run in the same container that also contains your local MAAS.**
+
+Inside your container (LXD or multipass) [generate a new SSH key](https://help.github.com/en/articles/generating-a-new-ssh-key-and-adding-it-to-the-ssh-agent) and [add it to your Github account](https://help.github.com/en/articles/adding-a-new-ssh-key-to-your-github-account).
+
+## Clone the repositiory
+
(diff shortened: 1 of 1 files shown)


ui/README.md
```diff
@@ -1,22 +1,41 @@
...
-**Note**: You will need an instance of MAAS running in the same container in order to run MAAS UI.
+**Note: You will need an instance of MAAS running in order to run MAAS UI.**
 
-**Note**: You will need your public and private SSH keys inside your container to be able to clone the repo. For help with this see [ssh-copy-id](https://www.ssh.com/ssh/copy-id).
+## Generate an SSH key in your container (LXD and multipass)
 
-Inside your MAAS container (either LXD or multipass) go to the parent directory of your MAAS installation, clone the MAAS UI repo and set it's upstream and origin:
+**Note: The following instructions should be run in the same container that also contains your local MAAS.**
+
+Inside your container (LXD or multipass) [generate a new SSH key](https://help.github.com/en/articles/generating-a-new-ssh-key-and-adding-it-to-the-ssh-agent) and [add it to your Github account](https://help.github.com/en/articles/adding-a-new-ssh-key-to-your-github-account).
```
squidsoup: @steverydz can you explain why we're recommending this? If someone already has a key, they should copy it to the container.

ui/README.md
```diff
@@ -27,12 +46,23 @@ REACT_APP_MAAS_URL="http://<your-local-maas-ip>:5240/MAAS"
...
 ```
 
+## Install Docker inside your container (LXD and multipass)
+
+The simplest way is to use [the convenience script](https://docs.docker.com/install/linux/docker-ce/ubuntu/#install-using-the-convenience-script). Once the script has run you will need to add your user to the docker group:
+
+```
+sudo usermod -aG docker $USER
+newgrp docker
+```
```
squidsoup: I don't recall ever doing this, are you sure `./run` doesn't do this itself?

Caleb-Ellis: I can confirm I had to do this when setting up docker in the lxc container (running bionic).

steverydz: I had to do it and so did @lilyvidenova. I've also always had to do it when setting up a new laptop and running any of the other projects that use `./run`

ui/README.md
steverydz: Ultimately it's because it's more straightforward, especially for those who aren't developers because they either have to copy and paste then figure out how to set the right permissions or try to use `ssh-copy-id` which always fails because there is no public key.

lilyvidenova:
I just want to mention, I'm super happy with the instructions as they are currently. :)

PR #92, opened 2019-08-13
Update general form

steverydz:
## Done
- Refactor general form into the containers view to match the same pattern as the rest of the forms
- Remove configuration component as not needed
- Redirected `/configuration` to `/configuration/general`
- Add action button component

## QA
- Go to `/configuration/general`
- See that form still works as expected
- See that the action button is used on the form


squidsoup:
The action button needs to reflect that the save was actually successful, i.e. `UPDATE_CONFIG_SUCCESS` was dispatched and set some state on `config`.

steverydz:
@squidsoup I'm not sure what you mean? The button does reflect when it's successful. It's the same button used on the other forms.

I'm not sure what you mean by setting state on config? If the values from the form are updated what else would need to be done?

steverydz:
@squidsoup It looks like @Caleb-Ellis has made [a card](https://github.com/canonical-web-and-design/maas-ui/issues/95) for this.

Caleb-Ellis:
@steverydz I've [created a card](https://github.com/canonical-web-and-design/maas-ui/issues/95) that outlines the issue. As I wrote in [my PR](https://github.com/canonical-web-and-design/maas-ui/pull/79) I think that can be solved as a separate issue.

ui/src/app/settings/containers/Configuration/GeneralForm/GeneralForm.js
```diff
@@ -0,0 +1,62 @@
...
+import config from "app/settings/selectors/config";
+import { formikFormDisabled } from "app/settings/utils";
+import { useSettingsSave } from "app/base/hooks";
+import ActionButton from "app/base/components/ActionButton";
+import Form from "app/base/components/Form";
+import GeneralFormFields from "app/settings/containers/Configuration/GeneralFormFields";
+
+const CommissioningSchema = Yup.object().shape({
+  maas_name: Yup.string().required(),
+  enable_analytics: Yup.string()
```
Caleb-Ellis: Isn't this a boolean?

steverydz: good catch, fixed

Issue #39, opened 2019-07-24
Surface errors to the client


squidsoup:
Currently `WEBSOCKET_ERROR`s are set in state, but not displayed to the user.

PR #97, opened 2019-08-14
Rewrite ActionButton to control own timeouts + calculate width dynamically

Caleb-Ellis:
## Done
- ActionButton width is now determined by the text, rather than hardcoded.
- Updated instances of ActionButton to remove width prop.
- Timer functions have been moved out of useSettingsSave in hooks.js and in to the ActionButton component itself.
- Removed useSettingsSave hook and replaced with config.saved state.
- Update config reducer so UPDATE_SETTINGS_START sets saved to false.

## QA
- Check all the ActionButtons behave as before
- Change one of the ActionButtons to different text
- Check that the button width remains constant from saving to success

### DIFF
--- a/ui/src/app/base/components/ActionButton/ActionButton.js
+++ b/ui/src/app/base/components/ActionButton/ActionButton.js
@@ -1,58 +1,108 @@
 import classNames from "classnames";
 import PropTypes from "prop-types";
-import React from "react";
+import React, { useEffect, useRef, useState } from "react";
 
-import Button from "../Button";
 import "./ActionButton.scss";
 
-const generateButtonChildren = (children, appearance, loading, success) => {
-  const iconClass = classNames({
-    "p-icon--spinner u-animation--spin": loading,
-    "is-light": appearance === "positive" || appearance === "negative",
-    "p-icon--success":
-      success && (appearance !== "positive" && appearance !== "negative"),
-    "p-icon--success-positive": success && appearance === "positive",
-    "p-icon--success-negative": success && appearance === "negative"
-  });
-
-  return loading || success ? <i className={iconClass} /> : children;
-};
+export const LOADER_MIN_DURATION = 400; // minimium duration (ms) loader displays
+export const SUCCESS_DURATION = 2000; // duration (ms) success tick is displayed
 
 const ActionButton = ({
   appearance = "neutral",
   children,
   className,
+  disabled,
+  inline,
   loading,
   success,
-  width,
   ...props
 }) => {
-  const classes = classNames(className, `p-action-button`, {
-    "is-loading": loading,
-    "is-success": success
-  });
-  const buttonChildren = generateButtonChildren(
-    children,
-    appearance,
-    loading,
-    success
+  const [width, setWidth] = useState(0);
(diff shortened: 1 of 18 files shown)


steverydz:
@Caleb-Ellis the button seems to stay in the done state now, even when I change a field after saving.

Caleb-Ellis:
@steverydz should be fixed now. I've split the loader and success timers up so that they're cleared properly now.

steverydz:
@Caleb-Ellis works great :+1:

squidsoup:
Hey Caleb, this is looking much better thanks!

The success timeout however doesn't appear to be working. To reproduce:

1. Kill MAAS
2. Save something

The button doesn't resolve to a failed state.

Caleb-Ellis:
@squidsoup Hmm I feel like that's the correct behaviour. If I kill MAAS, there's an `UPDATE_CONFIG_START` action that never resolves or fails (`config.saving` stays true) so really the button showing an infinite spinner is correctly reflecting state. Handling timeout errors should be something handled in the sagas, right? Which seems a bit out of scope for this branch, and sagas are scary.

As for the button resolving to a failed state, that doesn't match up with the designs. They show an error notification on top of the form, which I figured was going to be addressed separately in (this issue)[https://github.com/canonical-web-and-design/maas-ui/issues/39].

squidsoup:
@Caleb-Ellis I've probably misinterpreted what you `successTimeout` is doing.

> Handling timeout errors should be something handled in the sagas, right?

That's a fair point. I'll have another look on Monday, sorry for the misunderstanding!

ui/src/app/base/components/ActionButton/ActionButton.js
```diff
@@ -1,58 +1,108 @@
...
+    <button
+      className={buttonClasses}
+      disabled={disabled}
+      ref={ref}
+      style={width ? { width: `${width}px` } : undefined}
       {...props}
     >
-      {buttonChildren}
-    </Button>
+      {showLoader || showSuccess ? <i className={iconClasses} /> : children}
```
huwshimi: In the design it shows the text and the spinner at the same time.

Caleb-Ellis: @huwshimi oh yeah we did try that at first, but the changing text made the button width jump around so we opted to just show a spinner / success icon instead

ui/src/app/base/components/ActionButton/ActionButton.js
```diff
@@ -1,58 +1,108 @@
...
-    appearance,
-    loading,
-    success
+  const [width, setWidth] = useState(0);
+  const [showLoader, setShowLoader] = useState(false);
+  const [showSuccess, setShowSuccess] = useState(false);
+  const ref = useRef(null);
+
+  // Store default button width
+  useLayoutEffect(() => {
```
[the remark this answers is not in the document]
Caleb-Ellis: @squidsoup no actually this is a carry-over from me trying (and failing) to find a way to allow you to change the text + width of a button after it had already been rendered. But that's probably unnecessary and I didn't get it to work so I'll replace it with `useEffect`

PR #99, opened 2019-08-14
Reorganise component directory

Caleb-Ellis:
## Done
- `app/settings/containers` renamed to `app/settings/views` to match the nav. This folder should pretty much be in line with the settings url scheme. Within each subfolder are all the non-generic components specific to that section of the url (e.g all the `/users` stuff or all the `/storage` stuff that don't appear anywhere else).
- `app/settings/components` name unchanged. This folder should include all the components that specific to settings, but shared amongst views.
- `base/settings/components` name unchanged. This folder should include all the components that are (or could be) used in multiple places, regardless of view. Essentially the components we could include in a component library.
- moved `app/settings/views/Settings` to `app/settings/views` (i.e. the root of the views folder). This is admittedly a bit weird, but it is basically the parent component of all the settings component. Open to suggestions for where to put it.

## QA
- Check that everything still runs fine.
- Discuss and suggest alternatives!


squidsoup:
This all seems sensible to me. I'm not sure about the last point, any thoughts @huwshimi?

huwshimi:
Seems good to me. For that last point, I think it's OK as it wraps all those other views.

PR #100, opened 2019-08-14
Handle type 2 sync messages and update config state

squidsoup:
## Done
* Handle incoming 'sync' websocket messages. (messages with `type: 2` and an `action`)
* No longer update config state on receiving a `type: 1` response.
* Update config state upon receiving a sync message.
* `UPDATE_SUCCESS` now results in setting `saved` in state. The UI should rely on this state, and we'll just assume that the state becomes eventually consistent reliably and quickly. Sync messages tend to follow response messages within milliseconds, so practically this should be fine I think.

## QA
* Save a form.
* Check the diff for SUCCESS actions in redux dev tools, ensure item state is not touched.
* Check the diff for SYNC actions, ensure the changed fields value is updated in state.

### DIFF
--- a/ui/src/app/settings/reducers/config.js
+++ b/ui/src/app/settings/reducers/config.js
@@ -5,24 +5,27 @@ const config = produce(
     switch (action.type) {
       case "FETCH_CONFIG_START":
         draft.loading = true;
-        return;
+        break;
       case "FETCH_CONFIG_SUCCESS":
         draft.loading = false;
         draft.loaded = true;
         draft.items = action.payload;
-        return;
+        break;
       case "UPDATE_CONFIG_START":
         draft.saving = true;
-        return;
+        break;
       case "UPDATE_CONFIG_SUCCESS":
         draft.saving = false;
+        draft.saved = true;
+        break;
+      case "UPDATE_CONFIG_SYNC":
         draft.items = draft.items.map(item => {
           if (item.name === action.payload.name) {
             return action.payload;
           }
           return item;
         });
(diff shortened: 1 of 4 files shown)


ui/src/app/settings/reducers/config.test.js
```diff
@@ -90,7 +90,36 @@ describe("config reducer", () => {
       loading: false,
```
squidsoup: Diff is weird here, need to expand.

ui/src/app/settings/reducers/config.js
```diff
@@ -5,24 +5,27 @@ const config = produce(
...
+        break;
       case "UPDATE_CONFIG_START":
         draft.saving = true;
-        return;
+        break;
       case "UPDATE_CONFIG_SUCCESS":
         draft.saving = false;
+        break;
+      case "UPDATE_CONFIG_SYNC":
+        draft.synced = true;
```
huwshimi: Actually, not sure about setting synced here and we'd have to manually reset it. This is not in response to anything and it might be better to only do this on a `_SUCCESS` message.

Doing it here will also mean that you could get false positives if you're waiting for a sync and a different update comes over the socket.

squidsoup: In practical terms the UI can probably respond to a SUCCESS, but that makes the UI eventually consistent which is maybe a bit weird.

squidsoup:
@Caleb-Ellis if you are good with this, feel free to land it.

PR #103, opened 2019-08-16
Only fetch once from sagas

squidsoup:
## Done
1. Replace custom hook `useFetchOnce` with similar logic in `sendMessage` saga.
2. Add `model` meta field to actions, a string representing the corresponding root level object in redux state (e.g. users, repositories, config), so that dynamic selectors can operate over state in sagas.
3. `FETCH_*_LOADED` is now dispatched instead, which should ensure `loaded: true` and `loading: false` are set in state. `FETCH_*_LOADED` isn't _strictly_ necessary, but it felt like `FETCH_*` should have a corresponding response of some kind, otherwise it feels like the action was swallowed. Feel free to disagree!

## QA
With the redux inspector, or websocket network inspector open:
1. Visit the Users table
2. Follow a link to another view
3. Return to the Users table, expect to see:
  a. `FETCH_USERS_LOADED` dispatched
  b. No additional `FETCH_USERS_START` or `FETCH_USERS_SUCCESS` dispatched, nor any websocket requests for `users.list`.

### DIFF
--- a/ui/src/app/base/sagas/websockets.js
+++ b/ui/src/app/base/sagas/websockets.js
@@ -118,7 +118,8 @@ export function* handleMessage(socketChannel, socketClient) {
  * @param {Object} action.
  * @returns {Bool} - action is a request action.
  */
-const isWebsocketRequestAction = action => action.meta && action.meta.method;
+export const isWebsocketRequestAction = action =>
+  action.meta && action.meta.method;
 
 /**
  * Build a message for websocket requests.
@@ -151,7 +152,6 @@ export function* sendMessage(socketClient) {
     if (method.endsWith("list")) {
       const loaded = yield select(isLoaded, model);
       if (loaded) {
-        yield put({ type: `${type}_LOADED` });
         continue;
       }
     }
--- a/ui/src/app/base/sagas/websockets.js
+++ b/ui/src/app/base/sagas/websockets.js
@@ -118,8 +118,7 @@ export function* handleMessage(socketChannel, socketClient) {
  * @param {Object} action.
  * @returns {Bool} - action is a request action.
  */
-export const isWebsocketRequestAction = action =>
-  action.meta && action.meta.method;
+const isWebsocketRequestAction = action => action.meta && action.meta.method;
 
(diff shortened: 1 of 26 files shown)


ui/src/app/settings/containers/Network/NetworkDiscoveryForm/NetworkDiscoveryForm.test.js
```diff
@@ -17,7 +17,7 @@ describe("NetworkDiscoveryForm", () => {
         items: [
           {
             name: "active_discovery_interval",
-            value: 0,
+            value: "0",
```
squidsoup: This was an unrelated props bug, not sure how this landed? 😕

steverydz: how does this bug manifest itself?

squidsoup: @steverydz a warning in the console from proptypes

ui/src/app/base/sagas/websockets.js
```diff
@@ -134,6 +145,16 @@ export function* sendMessage(socketClient) {
     const data = yield take(action => isWebsocketRequestAction(action));
     const { type, meta } = data;
     const params = data.payload ? data.payload.params : null;
+
+    const { method, model } = meta;
+    // If method is 'list' and data has loaded, do not fetch again.
+    if (method.endsWith("list")) {
```
huwshimi: We may also need to check for "get" if we start to use those calls.

ui/src/app/settings/actions/config.js
```diff
@@ -4,6 +4,7 @@ config.fetch = () => {
   return {
     type: "FETCH_CONFIG",
     meta: {
+      model: "config",
       method: "config.list",
```
huwshimi: I wonder if the method should just be "list" and we combine the model and method in `buildMessage`. Not sure.

Issue #108, opened 2019-08-19
Add messages when users are saved


huwshimi:
When a user is added/updated there should be a message on the user list.

squidsoup:
@lilyvidenova I think we probably need a timed notification that disappears after a few seconds, or something similar, do we have existing patterns for that in maas?

lilyvidenova:
@squidsoup, @huwshimi - yes, currently we have a notification. I'd like to keep this pattern (at least for now). So on successful addition of anything, we should show an "information notification" reading "$username added successfully." 

I'd like to do something more fancy (like highlighting the row in the table in some way) but I'll leave it in the backlog for next cycle.

cassiocassio:
The juju desktop toasts that disappear automatically do so rather quickly. 

I can't find vanilla guidance about the timeout; if it's too fast, you feel cheated; if it's too slow, you'll waste attention and effort trying to clear it


(Adding to my a of patterns / guidance  to improve - Paris @kwm14 )
https://docs.vanillaframework.io/patterns/notification/ 

It's an open question about if self-exploding messages should even exist in the first place....

the best debate about this I could find:

https://ux.stackexchange.com/questions/11203/how-long-should-a-temporary-notification-toast-appear

PR #110, opened 2019-08-19
Add and edit users over the API

huwshimi:
Done:
- Add sagas/actions/reducers for creating/updating users.
- Connect the user create/update forms to the API. Fixes: https://github.com/canonical-web-and-design/maas-squad/issues/1373.

QA:
- Navigate to the users list.
- Click on the Add user button.
- Fill out the form and submit.
- You should be redirected to the user list and the new user should be in the list.
- Repeat with editing a user.
Note: a message should appear when you are redirected: https://github.com/canonical-web-and-design/maas-ui/issues/108.


ui/src/app/settings/utils.js
```diff
@@ -20,16 +20,14 @@ const simpleObjectEquality = (obj1, obj2) => {
...
-  const { initialValues, isSubmitting, errors, values } = formikProps;
+  const { initialValues, errors, values } = formikProps;
   let hasErrors = false;
 
   if (errors) {
     hasErrors = Object.keys(errors).length > 0;
   }
 
-  return (
-    isSubmitting || simpleObjectEquality(initialValues, values) || hasErrors
```
huwshimi: isSubmitting causes issues when controlling the state from redux (e.g. when server errors are returned in the state).

ui/src/app/base/components/FormikField/FormikField.js
```diff
@@ -12,15 +12,29 @@ const FormikField = ({
...
+    touched
+  } = formikProps;
+  const { serverErrors = {}, invalidValues = {} } = status;
   return (
     <Component
-      error={formikProps.touched[fieldKey] && formikProps.errors[fieldKey]}
+      error={
+        (touched[fieldKey] &&
+          (errors[fieldKey] ||
+            (value === invalidValues[fieldKey] && serverErrors[fieldKey]))) ||
```
Caleb-Ellis: This conditional is a bit hard to parse. Is it possible to split it out into something readable like:
``` javascript
error={isTouched && (hasFormikErrors || hasServerErrors)}
```
?

ui/src/app/settings/utils.js
Caleb-Ellis: That's fair. Might be a good idea to disable the button if `saving` is true in that case.

huwshimi: Is there a way to disable the button without the vanilla styles changing? Maybe it's OK for them to change, would just look different to the designs that's all.

huwshimi: Nevermind, looks like that case is already handled in the action button.

ui/src/app/base/components/FormikField/FormikField.js
huwshimi: Thanks, updated.

ui/src/app/settings/views/Users/UserFormFields/UserFormFields.js
```diff
@@ -15,6 +19,22 @@ const togglePassword = (event, passwordVisible, showPassword) => {
...
+    if (Object.keys(errors).length) {
+      const formikErrors = {};
+      const invalidValues = {};
+      Object.keys(errors).forEach(field => {
+        formikErrors[field] = errors[field].join(" ");
+        invalidValues[field] = formikProps.values[field];
+      });
+      formikProps.setStatus({ serverErrors: formikErrors, invalidValues });
+    }
+  }, [errors]); // eslint-disable-line react-hooks/exhaustive-deps
```
squidsoup: I think we should revisit this, but I have a migraine, so now is not the time.

ui/src/app/settings/reducers/users.js
```diff
@@ -11,14 +11,49 @@ const users = produce(
...
+      case "CREATE_USERS_SUCCESS":
+        draft.errors = {};
+        draft.saved = true;
+        draft.saving = false;
+        break;
+      case "UPDATE_USER_SYNC":
+        for (let i in draft.items) {
+          if (draft.items[i].id === action.payload.id) {
+            draft.items[i] = action.payload;
+            break;
```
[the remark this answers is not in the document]
huwshimi: I've updated the config reducer to match this one. This one is more performant as it breaks once it finds the item to update instead of mapping over all items.

PR #111, opened 2019-08-19
Build read-only RepositoriesList component + scaffolding for remaining repos sections

Caleb-Ellis:
## Done
- Built RepositoriesList component that shows all repos' names, urls, and enabled states (+ filtering and paginated at 20 per page)
- Built basic scaffolding for add and edit repository forms

## QA
- Go to `:8000/MAAS/settings/repositories`
- Check that the data in the table matches the data at `:5240/MAAS/#/settings/repositories`
- Check that the default repos are disabled (the first two)
- Check that filtering works as expected
- Change RepositoriesList prop `initialCount` to something low (e.g. 2) and check that pagination works as expected
- Click "Add PPA" or "Add repository" and check that the add form shows up (currently empty)
- Click the edit icon in a row and check that the edit form shows up (currently empty)
- Click the bin icon in a row and check that the row expands with delete confirmation

### Default
Screen Shot 2019-08-19 at 14 07 38-fullpage

### Filter
Screen Shot 2019-08-19 at 14 07 45-fullpage

### Delete
Screen Shot 2019-08-19 at 14 07 52-fullpage

### Add
Screen Shot 2019-08-19 at 14 07 58-fullpage

### Edit
Screen Shot 2019-08-19 at 14 08 02-fullpage


ui/src/app/settings/containers/Repositories/RepositoriesList/RepositoriesList.js
```diff
@@ -0,0 +1,160 @@
...
+      className: expanded ? "p-table__row is-active" : null,
+      columns: [
+        {
+          content: name,
+          role: "rowheader"
+        },
+        { content: repo.url },
+        {
+          content: (
+            <Input
```
huwshimi: It looks like the design shows "Yes" or "No" text for this column.

ui/src/app/settings/containers/Repositories/RepositoriesList/RepositoriesList.js
```diff
@@ -0,0 +1,160 @@
...
+  const [itemsPerPage] = useState(initialCount);
+
+  const indexOfFirstItem = (currentPage - 1) * itemsPerPage;
+
+  const paginate = pageNumber => setCurrentPage(pageNumber);
+
+  return (
+    <>
+      <div className="p-table-actions">
+        <SearchBox onChange={setSearchText} value={searchText} />
```
huwshimi: The design has the default text set to "Search package repositories"

ui/src/app/settings/containers/Repositories/RepositoriesList/RepositoriesList.scss
```diff
@@ -0,0 +1,32 @@
...
+  }
+
+  // URL
+  &:nth-child(2) {
+    flex: 0 0 40%;
+  }
+
+  // Enabled
+  &:nth-child(3) {
+    flex: 0 0 15%;
```
huwshimi: In the design I think this is even smaller.

ui/src/app/settings/containers/Repositories/RepositoriesList/RepositoriesList.scss
```diff
@@ -0,0 +1,32 @@
...
+  }
+
+  // Actions
+  &:nth-child(4) {
+    flex: 0 0 15%;
+  }
+
+  .p-form__label {
```
huwshimi: Once the input is changed to text these rules can be removed.

ui/src/app/settings/containers/Repositories/RepositoryAdd/RepositoryAdd.test.js
```diff
@@ -0,0 +1,11 @@
+import { mount } from "enzyme";
+import React from "react";
+
+import RepositoryAdd from "./RepositoryAdd";
+
+describe("RepositoryAdd", () => {
+  it("can render", () => {
+    const wrapper = mount(<RepositoryAdd />);
```
huwshimi: I think this can be shallow.

ui/src/app/settings/containers/Repositories/RepositoryForm/RepositoryForm.scss
```diff
@@ -0,0 +1,12 @@
...
+.repo-form {
+  margin-top: 0.5rem;
+}
+
+.repo-form__buttons {
+  border-top: 1px solid $color-mid-light;
+  margin-top: 1rem;
+  padding-top: 1rem;
+  text-align: right;
+}
```
huwshimi: We should probably turn this into pattern as we have this same code duplicated in a number of places.

ui/src/app/settings/selectors/repositories.js
```diff
@@ -1,10 +1,52 @@
...
+
+/**
+ * Get repositories that match a term.
+ * @param {Object} state - The redux state.
+ * @param {String} term - The term to match against.
+ * @returns {Array} A filtered list of repositories.
+ */
+repositories.search = (state, term) => {
+  return state.repositories.items.filter(
+    repo => repo.name.includes(term) || repo.url.includes(term)
```
huwshimi: Because we override the names of the repos, this won't match against the displayed names. I wonder if there's any way to include those values.

ui/src/app/settings/containers/Repositories/RepositoryForm/RepositoryForm.scss
Caleb-Ellis: Good call. I've created [a card for it](https://github.com/canonical-web-and-design/maas-ui/issues/119) that someone can pick up in a follow-up branch.

ui/src/app/settings/selectors/repositories.js
Caleb-Ellis: Hmm this is a tricky one. In RBAC we had a separate field in state for filter term. That way we could compose a selector for search that would include the filter term and a subset/mutated list of entities (e.g `repositories.getForSettings` or something that just renames the defaults). I've created [a card](https://github.com/canonical-web-and-design/maas-ui/issues/120) because I think a robust solution might take a bit of work with actions/reducers/selectors.

Caleb-Ellis:
Updated table:

Screen Shot 2019-08-20 at 10 22 13-fullpage

lyubomir-popov:
Looks good, just a few minor comments:

In the old maas all tables havbe a small indent. I've made an [issue](https://github.com/canonical-web-and-design/vanilla-framework/issues/2502) to provide this with vanilla, but in the meantime, can we keep the override? In the old maas, it is located in _base_tables.scss:

```
      &:first-of-type {
        padding-left: $sph-inner--small;
      }
```

I do not understand the purpose of the empty Edit repo screen:
image
Shouldn't we just disable the button if empty?

Padding-right on the last column should be $sph-inner--small:
image
This is an issue with the expanding cell, which is technically the last child so the rule that targets it doesn't apply to the last visible cell containing the buttons.
I've filed a [vanilla issue](https://github.com/canonical-web-and-design/vanilla-framework/issues/2503), in the meantime it would be nice if this could have a local override.

Caleb-Ellis:
@lyubomir-popov I've updated the PR with the two suggested overrides.

For the Add and Edit forms just ignore them for now, they're just there as scaffolding for when I actually write the content.

With the icon positioning I think that will make things unnecessarily tricky to align, especially given the "ACTIONS" label above. I would suggest instead that we include the delete button but disable it - at least that way we can add a tooltip or something explaining why they can't be deleted. @lilyvidenova maybe you'd like to weigh in?

lilyvidenova:
@Caleb-Ellis yep, super happy with including the delete icon but disabled + a tooltip reading "This ppa | repository cannot be deleted." (and if we can make this reusable so it applies to all other tables where there is a mixture of edit + delete and just edit, it'll be THE best)

Caleb-Ellis:
Updated:

10 153 61 90_8000_MAAS_settings_repositories (1)

lyubomir-popov:
much better thanks @Caleb-Ellis I've +1 it.

PR #113, opened 2019-08-19
Change Vanilla $multi to 1 + add Vanilla overrides file

Caleb-Ellis:
## Done
- Updated Vanilla settings to increase density (`$multi = 1`)
- Added a Vanilla overrides file **which we should definitely use**. If we can keep overrides in the same place and to a minimum we should get far less surprising regressions whenever we update Vanilla.

## QA
- Go to the users table and check that the density has increased.
- Click around the forms and check that everything looks okay.

### Users table
Screen Shot 2019-08-19 at 15 23 22-fullpage

### Form
Screen Shot 2019-08-19 at 15 23 34-fullpage

### DIFF
--- a/ui/src/scss/_patterns_table.scss
+++ b/ui/src/scss/_patterns_table.scss
@@ -7,16 +7,3 @@
 .p-table-expanding--light .p-table__row.is-active {
   background-color: $color-x-light;
 }
-
-// Stop the icons from adding extra height to the row:
-td:not(.p-table-expanding__panel) [class*="p-button"] {
-  margin-top: -0.25rem;
-  margin-bottom: -0.25rem;
-  padding-left: 0.5rem;
-  padding-right: 0.5rem;
-}
-
-.u-align-icons--top {
-  align-items: flex-start;
-  display: flex;
-}
--- a/ui/src/scss/_vanilla-overrides.scss
+++ b/ui/src/scss/_vanilla-overrides.scss
@@ -1,0 +1,37 @@
+/**
+ * Vanilla overrides.
+ * {Date} {Person}: {Description of issue}
+ * {Link to relevant GitHub issue}
+ */
+
+// 19-08-2019 Caleb: Form help text too close to input when $multi = 1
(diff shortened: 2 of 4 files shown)


ui/src/scss/_vanilla-overrides.scss
```diff
@@ -0,0 +1,23 @@
+/**
+ * Vanilla overrides.
+ * {Date} {Person}: {Description of issue}
+ * {Link to relevant GitHub issue}
+ */
+
+// 19-08-2019 Caleb: Form help text too close to input when $multi = 1
+// https://github.com/canonical-web-and-design/vanilla-framework/issues/2488
+input:not([type="checkbox"]) ~ .p-form-help-text,
```
squidsoup: Thanks for ensuring these get upstreamed @Caleb-Ellis

ui/src/scss/_vanilla-overrides.scss
```diff
@@ -0,0 +1,23 @@
+/**
+ * Vanilla overrides.
+ * {Date} {Person}: {Description of issue}
+ * {Link to relevant GitHub issue}
```
huwshimi: What now?!

huwshimi:
One small QA thing when hovering buttons:

Screen Shot 2019-08-20 at 12 49 40 pm

Caleb-Ellis:
@huwshimi should be fixed now

ui/src/scss/_patterns_table.scss
```diff
@@ -19,4 +19,5 @@ td:not(.p-table-expanding__panel) [class*="p-button"] {
 .u-align-icons--top {
   align-items: flex-start;
   display: flex;
+  padding-top: 0.3125rem;
```
lyubomir-popov: Ideally there should be an issue for this as well.

Caleb-Ellis: That's fair. I'll move it overrides file with [this Vanilla issue](https://github.com/canonical-web-and-design/vanilla-framework/issues/2504).

PR #116, opened 2019-08-19
Update commissioning field options on change

steverydz:
## Done
Updated the settings commissioning form so that the second fields options update when the first option changes. Fixes [#1414](https://github.com/canonical-web-and-design/MAAS-squad/issues/1414)

## QA
- Go to `/settings/configuration/commissioning`
- See that the field labelled `Default minimum kernel version` options are related to the above field (i.e. same version number 18.04 etc.)
- Change the field labelled `Default Ubuntu release used for commissioning` and see that the options in the field beneath are updated to reflect the new value


ui/src/app/settings/reducers/general/minHweKernels.js
```diff
@@ -0,0 +1,25 @@
+import produce from "immer";
+
+const minHweKernels = produce(
```
squidsoup: this reducer should just be called `general`, so please move this file to `general.js` and rename `const minHweGenerals` here to `general`. 

The reason being, all of these reducers for general properties are going to follow the same pattern, so instead of having nineteen distinct reducers, we can just have one with fallthrough cases.

e.g.
```
case "FETCH_GENERAL_MIN_HWE_KERNELS_SUCCESS":
case "FETCH_GENERAL_OSINFO_SUCCESS":
case ...
  draft[`${action.meta.method}`].loading = false;
  draft[`${action.meta.method}`].items = ...;
  ...
````

Currently, you'll have to parse `action.meta.method` and strip `general.` to get the key for the method name (which is also a suitable model name in state, e.g. `state.general.min_hwe_kernels`). I'll be landing a branch soon with a refactoring that will mean `meta.method` no longer has the `general.` prefix, and will be a value like `min_hwe_kernels`.

File: ui/src/root-reducer.js [hunk 2]
```diff
@@ -8,6 +8,7 @@ export default history =>
   combineReducers({
     auth: baseReducers.auth,
     config: settingsReducers.config,
+    minHweKernels: settingsReducers.minHweKernels,
```
squidsoup: `general: settingsReducers.general,`

ui/src/app/settings/actions/general/general.js
```diff
@@ -0,0 +1,14 @@
+const general = {};
+
+general.fetch = () => {
```
Caleb-Ellis: This should probably be called `general.fetchOsInfo`

ui/src/app/settings/selectors/general/general.js
```diff
@@ -0,0 +1,46 @@
...
+/**
+ * Returns kernels data
+ * @param {Object} state - the redux state
+ * @returns {Array} the available kernel options
+ */
+general.getKernalOptionsByKey = (state, key) => {
+  let kernelOptions = [];
+
+  if (state.general.items.kernels) {
+    kernelOptions = state.general.items.kernels.ubuntu[key];
```
Caleb-Ellis: This will break the UI if state.general.items.kernels.ubuntu doesn't exist.

I think this needs to be rewritten because really it requires knowledge of two things: the release *and* the OS. Otherwise it's only returning Ubuntu kernels.

ui/src/app/settings/reducers/general/general.js
```diff
@@ -0,0 +1,25 @@
...
+        draft.items = action.payload;
+        break;
+      default:
+        return draft;
+    }
+  },
+  {
+    loading: false,
+    loaded: false,
+    items: {}
```
Caleb-Ellis: Hmm admittedly this is a bit of a weird one. I feel like `items` should be used for arrays of the same type of entity, and maybe this should just have the different data objects at the top level e.g:

``` js
{
  loading: false,
  loaded: false,
  osInfo: null,
  hweKernels: null,
  minHweKernels: null,
  ...
}
```

That way when accessing them you would access `state.general.osInfo` instead of `state.general.items.osInfo` which seems unnecessarily nested.

Also I'm not sure if it's a good or bad idea to set the default to whatever data structure we expect to receive the data, so I've opted for `null` here - maybe @squidsoup or @huwshimi have a better idea for this.

ui/src/app/settings/selectors/general/general.js
```diff
@@ -0,0 +1,46 @@
...
+ * @returns {Boolean} - general is loading
+ */
+general.loaded = state => state.general.loaded;
+
+/**
+ * Returns kernels data
+ * @param {Object} state - the redux state
+ * @returns {Array} the available kernel options
+ */
+general.getKernalOptionsByKey = (state, key) => {
```
Caleb-Ellis: What is `key` here? I think it would be beneficial to give it a more descriptive name and include it in the doc block above.

ui/src/app/settings/selectors/general/general.js
steverydz: The field that this populates only requires ubuntu kernels so won't need knowledge of the os

Caleb-Ellis: Ahh, right. In that case I would name the selector `getUbuntuKernelOptions` or something, and I would still add a check for `state.general.kernels.ubuntu`.

ui/src/app/settings/actions/general/general.js
steverydz: Done

ui/src/app/settings/selectors/general/general.js
steverydz: Done

ui/src/app/settings/reducers/general/general.js
steverydz: Done

ui/src/app/settings/selectors/general/general.js
steverydz: Fixed

PR #123, opened 2019-08-20
Fix commissioning form

steverydz:
This is carried over from https://github.com/canonical-web-and-design/maas-ui/pull/116

I had some git issues and ended up just making a new PR.

There was some discussion on the previous PR which is worth reviewing.

## Done
Updated the settings commissioning form so that the second fields options update when the first option changes. Fixes [#1414](https://github.com/canonical-web-and-design/MAAS-squad/issues/1414)

## QA
- Go to `/settings/configuration/commissioning`
- See that the field labelled `Default minimum kernel version` options are related to the above field (i.e. same version number 18.04 etc.)
- Change the field labelled `Default Ubuntu release used for commissioning` and see that the options in the field beneath are updated to reflect the new value

### DIFF
--- a/ui/src/app/settings/selectors/general/general.js
+++ b/ui/src/app/settings/selectors/general/general.js
@@ -1,0 +1,45 @@
+const general = {};
+
+/**
+ * Returns object with osinfo data
+ * @param {Object} state - the redux state
+ * @returns {Object} The osinfo data
+ */
+general.osinfo = state => state.general.osInfo;
+
+/**
+ * Returns true if general is loading
+ * @param {Object} state - the redux state
+ * @returns {Boolean} - general is loading
+ */
+general.loading = state => state.general.loading;
+
+/**
+ * Returns true if general is loaded
+ * @param {Object} state - the redux state
+ * @returns {Boolean} - general is loading
+ */
+general.loaded = state => state.general.loaded;
+
+/**
+ * Returns kernels data
+ * @param {Object} state - the redux state
+ * @param {String} os - the OS to get kernel options for
+ * @returns {Array} the available kernel options
+ */
+general.getUbuntuKernelOptions = (state, os) => {
+  let kernelOptions = [];
+
+  if (state.general.osInfo.kernels && state.general.osInfo.kernels.ubuntu) {
+    kernelOptions = state.general.osInfo.kernels.ubuntu[os];
+  }
+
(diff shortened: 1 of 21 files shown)


squidsoup:
Hey @steverydz, this is looking much better! Unfortunately it looks like there's a race condition somewhere which means changing default release, and default min kernel results in an error.

Try the following:

1. Change the default ubuntu release.
2. Select a new min kernel
3. Refresh, note that the min kernel has not been updated and that an error is returned from the server, .e.g. `"'hwe-18.04-edge' is not a valid default_min_hwe_kernel. It should be one of: '', 'ga-16.04-lowlatency', 'ga-16.04', 'hwe-16.04-lowlatency', 'hwe-16.04', 'hwe-16.04-lowlatency-edge', 'hwe-16.04-edge'."`

squidsoup:
Hey @steverydz, I've proposed a fix for the problem [here](https://github.com/canonical-web-and-design/maas-ui/pull/127). Could you QA and review that for me when you get a chance please?

ui/src/app/settings/selectors/general/general.js
```diff
@@ -0,0 +1,47 @@
...
+  }
+
+  kernelOptions = [["", "No minimum kernel"]].concat(kernelOptions);
+
+  return kernelOptions.map(option => {
+    return {
+      value: option[0],
+      label: option[1]
+    };
+  });
```
squidsoup: When immediately returning you can just say:

```
  return kernelOptions.map(option => (
    {
      value: option[0],
      label: option[1]
    };
  ));
```

ui/src/app/settings/views/Settings.js
```diff
@@ -10,6 +10,7 @@ const Settings = () => {
   const dispatch = useDispatch();
   useEffect(() => {
     dispatch(actions.config.fetch());
+    dispatch(actions.general.fetchOsInfo());
```
squidsoup: The pattern we have in place so far (with the exception of config) is to dispatch fetch requests for data where it used, so could this live in the Commissioning component?

ui/src/app/settings/selectors/general/general.js
steverydz: Done

ui/src/app/settings/views/Settings.js
steverydz: Done

squidsoup:
QA +1 thanks @steverydz

PR #131, opened 2019-08-21
Response error json fallback

squidsoup:
## Done
The API is inconsistent about returning errors as JSON, so we need to fallback to directly assigning errors that aren't parsable. 

## QA
* Attempt to delete the user with id: 2, this should result in an error which should endup in user state.

see: https://bugs.launchpad.net/maas/+bug/1840887

### DIFF
--- a/ui/src/app/base/sagas/websockets.js
+++ b/ui/src/app/base/sagas/websockets.js
@@ -102,9 +102,19 @@ export function* handleMessage(socketChannel, socketClient) {
         response.request_id
       );
       if (response.error) {
+        let error;
+        try {
+          error = JSON.parse(response.error);
+        } catch {
+          // the API doesn't consistently return JSON, so we fallback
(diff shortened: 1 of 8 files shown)


Caleb-Ellis:
Ahh yes, `'("Cannot delete some instances of model \'User\' because they are referenced through a protected foreign key: \'Node.owner\'", <QuerySet [<Node: schpft (full-bee)>, <Node: ntsgtg (noble-locust)>, <Node: t7xmcy (actual-cod)>, <Node: mnwheh (large-kiwi)>, <Node: nyaaff (honest-rat)>, <Node: gmnbk3 (casual-buck)>, <Node: 6m4bky (sound-bedbug)>]>)'`, of course!

Anyway looks good! But Huw's changes are still showing up in this PR? I think it's probably fine to merge but maaaaaybe rebase just in case. but this branch might need a rebase?

PR #132, opened 2019-08-21
Implement deleting repositories over the API

Caleb-Ellis:
## Done
- Implemented deleting repos over the api
- Renamed actions to match model names
  - `REPOSITORIES` => `PACKAGEREPOSITORY`
  - `USERS` => `USER`

## QA
- Add some repos at `:5240/MAAS/#/settings/repositories`
- Go to `:8000/MAAS/settings/repositories` and see that you can delete them successfully
- Refresh maas-ui and check back on `:5240/MAAS/#/settings/repositories` to make sure they're gone for real
- Do the same for users just in case the renaming messed something up unexpectedly

### DIFF
--- a/ui/src/app/settings/actions/repositories.js
+++ b/ui/src/app/settings/actions/repositories.js
@@ -2,7 +2,7 @@ const repositories = {};
 
 repositories.fetch = () => {
   return {
-    type: "FETCH_REPOSITORIES",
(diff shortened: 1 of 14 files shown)


Caleb-Ellis:
Since this is basically a copy+paste+rename job from #128 I'm going to merge this with only one approval. I don't think the renaming of actions is contentious because it's really just keeping things uniform.

PR #133, opened 2019-08-21
Create settings deploy component

steverydz:
## Done
Created the component for the configuration settings deploy form.

**Drive by**: 
- Moved the action dispatch for general config in the commissioning form from the `CommissioningForm` component to `Commissioning`.
- Fixed a bug with updating the second value in the commissioning form

## QA
- Go to `:5240/MAAS/#/images` and find the **Other images** section near the bottom of the page, select all available images and save
- Go to `:8000/MAAS/settings/configuration/deploy`
- Change the value of **Default operating system used for deployment** and see that the options for **Default OS release used for deployment** are updated
- Change values in both of the fields and save
- Refresh and see that those values persist
- Go to `:5240/MAAS/settings/general` and see that the form fields for the **Deploy** form have the same values as MAAS UI
- Change the values here and save them
- Go back to MAAS UI, refresh and see that the values are the same as what you changed them to in MAAS
- Go to the commissioning form and check that works as expected

### DIFF
--- a/ui/src/app/settings/selectors/config.js
+++ b/ui/src/app/settings/selectors/config.js
@@ -332,7 +332,7 @@ config.discoveryIntervalOptions = createSelector(
 
 /**
  * Returns the MAAS config for kernel parameters.
- * @param {object} state - The redux state.
+ * @param {Object} state - The redux state.
  * @returns {String} Kernel parameters.
  */
 config.kernelParams = createSelector(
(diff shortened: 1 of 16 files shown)


ui/src/app/settings/selectors/config.js
```diff
@@ -332,12 +332,51 @@ config.discoveryIntervalOptions = createSelector(
...
+
+/**
+ * Returns the possible default OS options reformatted as objects.
+ * @param {Object} state - The redux state.
+ * @returns {Array} Default OS options.
+ */
+config.defaultOSystemOptions = createSelector(
+  [config.all],
+  configs => {
+    const configObj = configs.find(config => config.name === "default_osystem");
```
Caleb-Ellis: There's a helper function `getOptionsFromName` above that you can use instead

steverydz: done

ui/src/app/settings/views/Configuration/DeployFormFields/DeployFormFields.js
```diff
@@ -0,0 +1,52 @@
...
+
+  return (
+    <>
+      <FormikField
+        label="Default operating system used for deployment"
+        component={Select}
+        options={defaultOSystemOptions}
+        fieldKey="default_osystem"
+        formikProps={formikProps}
+        onChange={e => {
```
steverydz: @squidsoup @huwshimi I'm having an issue here. When changing the value for the top field, the formik value for the second field isn't updated.

I'm trying to do this manually using the built in methods that formik provides but `distroSeriesOptions` doesn't update in time for this to work. I've tried calling `useSelector` within the `onChange` handler but I can't actually do that because of how hooks work.

huwshimi: Hmm... I can't reproduce this. When I change the OS the series field does update for me (switches between centos 6/bionic).

Caleb-Ellis: @huwshimi the second field visibly updates, but the value within formikProps.values itself doesn't. You have to manually click the field for it to update.

squidsoup: @steverydz hey steve, did you resolve this issue?

steverydz: @squidsoup yeah, I made a selector that returns all options formatted in a way that you can easily access what you want based on the key

Caleb-Ellis:
Having thought about this a bit more I think ignore what I said above. At the moment state is only updated on `UPDATE_CONFIG_SYNC` and we should keep that consistent. I think the point is still valid, that state should be as up to date as possible, but the solution doesn't need to belong in this PR.

PR #136, opened 2019-08-22
Add component docs

huwshimi:
Done:
- Add styleguidist and basic generic component docs.

QA:
- Run `./run exec --expose-port 6060 yarn docs`.
- You should see docs for components.

### DIFF
--- a/ui/src/app/base/components/ActionButton/ActionButton.md
+++ b/ui/src/app/base/components/ActionButton/ActionButton.md
@@ -1,21 +1,21 @@
-## Default
+### Default
 
 ```jsx
 <ActionButton appearance="positive">Click me!</ActionButton>
 ```
 
-## Loading
+### Loading
 
 ```jsx
-<ActionButton appearance="positive" loading={true}>
+<ActionButton appearance="positive" loading>
   Click me!
 </ActionButton>
 ```
 
-## Success
+### Success
 
 ```jsx
-<ActionButton appearance="positive" success={true}>
+<ActionButton appearance="positive" success>
   Click me!
 </ActionButton>
 ```
--- a/ui/src/app/base/components/ActionButton/ActionButton.md
+++ b/ui/src/app/base/components/ActionButton/ActionButton.md
@@ -1,0 +1,21 @@
+## Default
+
+```jsx
+<ActionButton appearance="positive">Click me!</ActionButton>
+```
+
+## Loading
+
+```jsx
+<ActionButton appearance="positive" loading={true}>
+  Click me!
+</ActionButton>
+```
+
+## Success
+
+```jsx
+<ActionButton appearance="positive" success={true}>
+  Click me!
+</ActionButton>
+```
--- a/ui/src/app/base/components/ActionButton/ActionButton.md
+++ b/ui/src/app/base/components/ActionButton/ActionButton.md
@@ -15,7 +15,18 @@
 ### Success
 
 ```jsx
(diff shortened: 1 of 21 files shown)


ui/src/app/base/components/Button/Button.md
```diff
@@ -0,0 +1,73 @@
...
+<span>Everything you need to get started with Vanilla. </span>
+<Button appearance="neutral" inline={true}>
+  Inline button
+</Button>
+```
+
+## Icon
+
+```jsx
+<Button appearance="p-button" hasIcon={true}>
```
squidsoup: Is this correct? Looks like we're passing a vanilla class to this prop?

ui/src/app/base/components/ActionButton/ActionButton.md
```diff
@@ -0,0 +1,21 @@
+## Default
+
+```jsx
+<ActionButton appearance="positive">Click me!</ActionButton>
+```
+
+## Loading
+
+```jsx
+<ActionButton appearance="positive" loading={true}>
```
squidsoup: all the props set to `{true}` can just be the props themselves, e.g.

`<ActionButton appearance="positive" loading>`
`<Button appearance="neutral" disabled>`

etc.

ui/src/app/base/components/Card/Card.md
```diff
@@ -0,0 +1,54 @@
+## Default
```
squidsoup: All the child headers of Card (default, header etc) use the same heading (h2) as Card, which is a bit confusing. Should these be `H3`s here, e.g. `### Default`?

ui/src/app/base/components/Col/Col.md
```diff
@@ -0,0 +1,160 @@
+## Grid
+
+```jsx
```
squidsoup: Would the rendered examples be clearer here with borders around the cells?

ui/src/app/base/components/ActionButton/ActionButton.md
huwshimi: Done.

ui/src/app/base/components/Button/Button.md
huwshimi: Oops, fixed.

ui/src/app/base/components/Card/Card.md
huwshimi: Good idea, thanks.

ui/src/app/base/components/Col/Col.md
huwshimi: Done, I've made this match the Vanilla docs: https://docs.vanillaframework.io/patterns/grid/.

ui/src/app/base/components/ActionButton/ActionButton.md
```diff
@@ -0,0 +1,21 @@
...
+```jsx
+<ActionButton appearance="positive" loading>
+  Click me!
+</ActionButton>
+```
+
+### Success
+
+```jsx
+<ActionButton appearance="positive" success>
```
Caleb-Ellis: This isn't working because the internal logic requires that the button moves from loading to success. Do you know if there's a way to have a "dynamic" example that can show these state transitions? Otherwise you can probably remove this section.

huwshimi: I've added a second button to transition the state when clicked.

PR #140, opened 2019-08-22
Fetch general data for repository adding/editing

Caleb-Ellis:
## Done
- Created general action creator
- Fetch general data `components_to_disable`, `known_architectures` and `pockets_to_disable` in RepositoryForm
- Rewrote `general` state structure so that each field manages its own loading and loaded state.
``` js
general: {
  knownArchitectures: {
    loading: false,
    loaded: false,
    data: []
  },
  osInfo: {
    loaded: false,
    loading: false,
    data: {}
  },
...
```
This has ended up creating a loooooot of similar/duplicated code for reducers and selectors, but I think it's easier to reason about when everything is cleanly separated. Further down the line we could figure out the parts to refactor. 

It's also a bit nested which is unfortunate, I guess the alternative would be to have each loading/loaded field at the top level, i.e.: 
``` js
general: {
  knownArchitectures: [],
  knownArchitecturesLoading: false,
  knownArchitecturesLoaded: false,
  osInfo: {},
  osInfoLoading: false,
  osInfoLoaded: false,
...
```
However would we have errors as a separate field as well? Since there are 17 general websocket methods it seems like this general object could get pretty wide (though not sure if that's necessarily a bad thing).

## QA
- Go to /repositories and click on Add PPA, Add repository or edit a repo
- Check that redux actions are dispatched to fetch general data and it is reflected in state
- Refresh the page and check that a loader shows while the repos and general data are fetched


ui/src/app/settings/actions/general/general.js
```diff
@@ -1,14 +1,24 @@
...
+ */
+const generateGeneralAction = method => () => ({
+  type: `FETCH_GENERAL_${method.toUpperCase()}`,
+  meta: {
+    model: "general",
+    method,
+    type: 0
+  }
+});
+
```
squidsoup: Nice

ui/src/app/settings/views/Repositories/RepositoryForm/RepositoryForm.js
```diff
@@ -1,40 +1,65 @@
...
 export const RepositoryForm = ({ title }) => {
+  const generalLoaded = useSelector(selectors.general.loaded);
+  const reposLoaded = useSelector(selectors.repositories.loaded);
+
+  const dispatch = useDispatch();
+  useEffect(() => {
+    if (!reposLoaded) {
+      dispatch(actions.repositories.fetch());
+    }
+    if (!generalLoaded) {
```
squidsoup: Currently, these actions all get dispatched twice.

There is a more sinister issue though - given _any_ general fetch action will set `state.general.loaded`, if another component dispatched a general fetch action (not even one of the ones here!), these three actions would never be dispatched.

I think the problem here is that these items in state are all sharing the same `loaded` property, rather than having their own.

Caleb-Ellis: Hmm I guess I can create reducers for each piece of general data, each with their own loading and loaded fields, then the payload from success could live in a data field, like:
``` js
{
  osInfo: {
    loaded: false,
    loading: false,
    data: {}
  },
  knownArchitectures: {
    loading: false,
    loaded: false,
    data: []
  }
...

**Comment by squidsoup** [comment]
Looks good now, thanks Caleb. Please drop any instances where you're setting `type: 0` in actions before landing though - this is now just done in the saga (since anything with a `method` is always going to be a `MESSAGE_TYPES.REQUEST`).

File: ui/src/app/settings/actions/general/general.js [hunk 3]
```diff
@@ -1,14 +1,24 @@
...
+ * Generate a redux action creator for general websocket methods
+ * @param {String} method - Name of the general websocket method
+ * @returns {Function} Redux action creator for fetching general data
+ */
+const generateGeneralAction = method => () => ({
+  type: `FETCH_GENERAL_${method.toUpperCase()}`,
+  meta: {
+    model: "general",
+    method,
+    type: 0
```
squidsoup: As mentioned, please drop `type: 0` here

ui/src/app/settings/actions/general/general.test.js
```diff
@@ -1,6 +1,28 @@
 import general from "./general";
 
 describe("general actions", () => {
+  it("should handle fetching components to disable", () => {
+    expect(general.fetchComponentsToDisable()).toEqual({
+      type: "FETCH_GENERAL_COMPONENTS_TO_DISABLE",
+      meta: {
+        model: "general",
+        method: "components_to_disable",
+        type: 0
```
squidsoup: and here

ui/src/app/settings/actions/general/general.test.js
```diff
@@ -1,6 +1,28 @@
...
+    });
+  });
+
+  it("should handle fetching known architectures", () => {
+    expect(general.fetchKnownArchitectures()).toEqual({
+      type: "FETCH_GENERAL_KNOWN_ARCHITECTURES",
+      meta: {
+        model: "general",
+        method: "known_architectures",
+        type: 0
```
squidsoup: and here

ui/src/app/settings/actions/general/general.test.js
```diff
@@ -11,4 +33,15 @@ describe("general actions", () => {
...
     });
   });
+
+  it("should handle fetching pockets to disable", () => {
+    expect(general.fetchPocketsToDisable()).toEqual({
+      type: "FETCH_GENERAL_POCKETS_TO_DISABLE",
+      meta: {
+        model: "general",
+        method: "pockets_to_disable",
+        type: 0
```
squidsoup: and here

ui/src/app/settings/views/Repositories/RepositoryForm/RepositoryForm.test.js
```diff
@@ -1,11 +1,152 @@
...
+      </Provider>
+    );
+
+    expect(store.getActions()).toEqual([
+      {
+        type: "FETCH_PACKAGEREPOSITORY",
+        meta: {
+          model: "packagerepository",
+          method: "list",
+          type: 0
```
squidsoup: and here

ui/src/app/settings/views/Repositories/RepositoryForm/RepositoryForm.test.js
```diff
@@ -1,11 +1,152 @@
...
+      </Provider>
+    );
+
+    expect(store.getActions()).toEqual([
+      {
+        type: "FETCH_GENERAL_COMPONENTS_TO_DISABLE",
+        meta: {
+          model: "general",
+          method: "components_to_disable",
+          type: 0
```
squidsoup: and here

ui/src/app/settings/views/Repositories/RepositoryForm/RepositoryForm.test.js
```diff
@@ -1,11 +1,152 @@
...
+      </Provider>
+    );
+
+    expect(store.getActions()).toEqual([
+      {
+        type: "FETCH_GENERAL_KNOWN_ARCHITECTURES",
+        meta: {
+          model: "general",
+          method: "known_architectures",
+          type: 0
```
squidsoup: and here

ui/src/app/settings/views/Repositories/RepositoryForm/RepositoryForm.test.js
```diff
@@ -1,11 +1,152 @@
...
+      </Provider>
+    );
+
+    expect(store.getActions()).toEqual([
+      {
+        type: "FETCH_GENERAL_POCKETS_TO_DISABLE",
+        meta: {
+          model: "general",
+          method: "pockets_to_disable",
+          type: 0
```
squidsoup: and here

PR #144, opened 2019-08-23
Toggle username

huwshimi:
Done:
- Added toggle between username and full name. Fixes: https://github.com/canonical-web-and-design/maas-squad/issues/1404.
- Added missing user details.

QA:
- View the user list.
- You should see machine count, last login and type for users with those details.
- In the first column click on "Real name", you should see the real name for those users that have one.

### DIFF
--- a/ui/src/app/settings/views/Users/UsersList/UsersList.js
+++ b/ui/src/app/settings/views/Users/UsersList/UsersList.js
@@ -11,7 +11,6 @@ import selectors from "app/settings/selectors";
 import baseSelectors from "app/base/selectors";
 import Button from "app/base/components/Button";
 import Col from "app/base/components/Col";
-import VanillaLink from "app/base/components/Link";
 import Loader from "app/base/components/Loader";
 import Pagination from "app/base/components/Pagination";
 import MainTable from "app/base/components/MainTable";
@@ -28,7 +27,6 @@ const generateUserRows = (
 ) =>
   users.map(user => {
     const expanded = expandedId === user.id;
-    // const last_login = parse(user.last_login, "yyyy-LL-DD H:mm", "bob");
     // Dates are in the format: Thu, 15 Aug. 2019 06:21:39.
     const last_login = user.last_login
       ? format(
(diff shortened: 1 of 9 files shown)


ui/src/app/settings/views/Users/UsersList/UsersList.js
```diff
@@ -149,7 +166,37 @@ const Users = ({ initialCount = 20 }) => {
...
+            {
+              className: "p-table-multi-header",
+              content: (
+                <>
+                  <VanillaLink
+                    className={classNames("p-table-multi-header__link", {
+                      "is-active": displayUsername
+                    })}
+                    onClick={e => {
+                      e.preventDefault();
```
Caleb-Ellis: In Vanilla you can get a link that looks like a button, but I wonder if it's a good idea to have a pattern for a button that looks like a link. I added `p-button--link` to RBAC because that way you could get all the default button behaviour (tabbing, focus, no need to prevent default) and eslint wouldn't complain about using links wrong.

lilyvidenova:
Looks good, I have two comments: 

- [ ] could we make sure the search works on the names as well as the usernames? 
- [ ] can we change the N/A to an em dash (⁠—)

ui/src/app/settings/views/Users/UsersList/UsersList.js
```diff
@@ -20,21 +23,33 @@ const generateUserRows = (
   authUser,
   expandedId,
   setExpandedId,
-  dispatch
+  dispatch,
+  displayUsername
 ) =>
   users.map(user => {
     const expanded = expandedId === user.id;
+    // const last_login = parse(user.last_login, "yyyy-LL-DD H:mm", "bob");
```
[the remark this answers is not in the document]
huwshimi: C'mon bob.

ui/src/app/settings/views/Users/UsersList/UsersList.js
huwshimi: Thanks, I've added the --link pattern from rbac.

huwshimi:
@lilyvidenova I've made those two changes. I hope you don't mind, I'll land this because they're small changes.

PR #145, opened 2019-08-23
Add MESSAGE_TYPES const and rename SYNC to NOTIFY.

squidsoup:
## Done
* Replace magic websocket request type integers with more descriptive constant values (e.g. `type: 0` -> `MESSAGE_TYPES.REQUEST`).
* Rename `SYNC` to `NOTIFY` as that's what update messages are referred to as on the server.

from `protocols.py`:

```python
class MSG_TYPE:
    #: Request made from client.
    REQUEST = 0

    #: Response from server.
    RESPONSE = 1

    #: Notify message from server.
    NOTIFY = 2

    #: Connectivity checks
    PING = 3
    PING_REPLY = 4
```

We don't handle pings so I've left these out.

## QA
Ensure the app still works (sorry, realise that's pretty vague).

### DIFF
--- a/ui/src/app/base/actions.js
+++ b/ui/src/app/base/actions.js
@@ -1,5 +1,3 @@
-import MESSAGE_TYPES from "app/base/constants";
-
 export const connectWebSocket = () => {
   return {
     type: "WEBSOCKET_CONNECT"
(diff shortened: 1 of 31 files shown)


ui/src/app/base/actions.js
```diff
@@ -10,7 +12,7 @@ export const fetchAuthUser = () => {
     meta: {
       model: "user",
       method: "auth_user",
-      type: 0
+      type: MESSAGE_TYPES.REQUEST
```
huwshimi: As discussed let's move this into the saga as they always need to be the same type.

PR #157, opened 2019-08-27
Display a message when adding/updating/removing users

huwshimi:
Done:
- Added global messages actions/reducers.
- Added a Notification component.
- Updated users form/list to show a message when users are created, updated or deleted. Fixes: https://github.com/canonical-web-and-design/maas-squad/issues/1410. Fixes: https://github.com/canonical-web-and-design/maas-ui/issues/108.

QA:
- Add, update or delete a user.
- You should get a message.

### DIFF
--- a/ui/src/app/base/components/Section/Section.js
+++ b/ui/src/app/base/components/Section/Section.js
@@ -1,11 +1,30 @@
+import { useDispatch, useSelector } from "react-redux";
 import PropTypes from "prop-types";
 import React from "react";
 
 import "./Section.scss";
+import { messages as messageActions } from "app/base/actions";
 import Col from "app/base/components/Col";
+import Notification from "app/base/components/Notification";
+import selectors from "app/base/selectors";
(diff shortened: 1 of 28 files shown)


lilyvidenova:
@huwshimi Can we just put a full stop at the end of each message?

ui/src/app/base/components/Section/Section.js
```diff
@@ -23,6 +42,7 @@ const Section = ({ children, sidebar, title }) => {
           </Col>
         )}
         <Col size={sidebar ? 10 : 12} className="section__content">
+          {generateMessages(messages, dispatch)}
```
Caleb-Ellis: This could probably be a separate component - NotificationsList or something.

huwshimi:
@lilyvidenova Thanks, I've added the full stops.

ui/src/app/base/components/Section/Section.js
huwshimi: Good idea. Done.

PR #160, opened 2019-08-27
Implement adding and editing repositories

Caleb-Ellis:
## Done
- Implemented adding and editing repositories

## QA
- Click "Add repository", fill in the forms correctly and click "Save repository". Check that you are redirected to the repo list and the new repo is there with a notification.
- Click the edit button for the repo and edit some fields. Click "Save repository" and check that you are redirected to the repo list with a notification.
- Click "Add repository" and use the same name you used before. Click "Save repository" and check that a server error shows underneath the Name field.
- Delete a repo and see there is a notification.
- In another browser tab, make some changes to repos from the old maas settings and check that the details are updated in real-time.
- Do the same with users because I've used a couple of custom hooks there too

### DIFF
--- a/ui/src/app/base/hooks.js
+++ b/ui/src/app/base/hooks.js
@@ -2,6 +2,8 @@ import { useEffect, useRef } from "react";
 import { useContext } from "react";
 import { __RouterContext as RouterContext } from "react-router";
 
+import { simpleObjectEquality } from "app/settings/utils";
+
 // Router hooks inspired by: https://github.com/ReactTraining/react-router/issues/6430#issuecomment-510266079
 // These should be replaced with official hooks if/when they become available.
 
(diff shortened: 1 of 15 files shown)


lilyvidenova:
- I can't seem to edit the "Ubuntu archive" repo - I see the button loading and then nothing happens.

- 
Screenshot 2019-08-28 at 16 55 40

Caleb-Ellis:
@lilyvidenova - Good catch on the Ubuntu archive. Turns out you can't disable a default repo so I should disable that field. Also caught another bug - Distributions and Components shouldn't show up in default repo forms either. D:

Fixed the default repo name.

I left the table header though - that's just how Vanilla tables look: https://docs.vanillaframework.io/base/tables/
I'd prefer filing an issue on Vanilla than having a local override.

lilyvidenova:
Lyubo has already filed an issue about the table stuff apparently, so all good there! 

And great about the rest 🚀

squidsoup:
Hey Caleb, this looks excellent thank you, however there's an issue where errors aren't displayed the *second* time an invalid form is submitted.

## To reproduce
1. edit a repo
2. Set an invalid url like `http://1`
3. Submit the form, you'll see an error.
4. Edit the url again to another invalid url like `http://2`
5. Submit the form again, `UPDATE_PACKAGEREPOSITORY_ERROR` is dispatched, however there is no error payload, and the UI doesn't display a subsequent error.

Caleb-Ellis:
@squidsoup good catch - updated so errors are cleaned up on submit.

PR #163, opened 2019-08-27
Upgrade to react-redux 7.1.1 and update snapshots.

squidsoup:
## Done
Upgrade from react-redux 7.1.0 -> 7.1.1. Normally we would let renovate do this, but it broke some snapshots.


squidsoup:
The failing labeler action can be ignored, a [known issue with forks](https://github.com/actions/labeler/issues/12).

Issue #164, opened 2019-08-28
Fetch machines, devices and controllers


huwshimi:
These are required by the dhcp snippet list.

PR #167, opened 2019-08-28
Fetch machines, controllers and devices

huwshimi:
Done:
- Implement actions, reducers and selectors for machines, controllers and devices. Fixes: #164.

QA:
- Visit /settings/dhcp.
- You should subnets, machines, controllers or devices in brackets after some items.

### DIFF
--- a/ui/src/app/settings/views/Dhcp/DhcpList/DhcpList.js
+++ b/ui/src/app/settings/views/Dhcp/DhcpList/DhcpList.js
@@ -2,17 +2,42 @@ import { useDispatch, useSelector } from "react-redux";
 import React, { useEffect } from "react";
 
 import actions from "app/settings/actions";
+import {
+  controller as controllerActions,
+  device as deviceActions,
+  machine as machineActions
+} from "app/base/actions";
+import {
+  controller as controllerSelectors,
+  device as deviceSelectors,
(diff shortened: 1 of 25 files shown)


ui/src/app/settings/views/Dhcp/DhcpList/DhcpList.js
```diff
@@ -2,17 +2,42 @@ import { useDispatch, useSelector } from "react-redux";
...
 import Loader from "app/base/components/Loader";
 import selectors from "app/settings/selectors";
 
 // Temporarily create a item component, this will be replaced with a table.
-const DhcpListItem = ({ id, name, subnet }) => {
+const DhcpListItem = ({ id, name, node, subnet }) => {
   const subnetItem = useSelector(state =>
     selectors.subnet.getById(state, subnet)
   );
+  const controllerItem = useSelector(state =>
```
squidsoup: I don't think we should be calling memoised selectors in our components (https://react-redux.js.org/next/api/hooks#using-memoizing-selectors), but we can fix that in a later branch.

squidsoup: As discussed this is completely wrong apologies, we shouldn't **declare** memoised selectors in components, which we never do.

PR #170, opened 2019-08-30
Explicitly fetch config in components that require it.

Caleb-Ellis:
## Done
- Explicitly fetch config in components that require it, if it hasn't already been loaded

## QA
- Load the app from users
- Navigate to a form that requires config (e.g. General)
- Check that the page actually loads.


squidsoup:
Nice! 🎉

PR #173, opened 2019-09-02
Fetch scripts, add routes and basic component.

squidsoup:
## Done
* Configure [http-proxy-middleware](https://create-react-app.dev/docs/proxying-api-requests-in-development#configuring-the-proxy-manually) to work around CORS issues when running the dev server.
* Fetch scripts from http API.
* Add basic component, displaying lists of commissioning and testing scripts.

## QA
* Upload a commissioning script if you don't have one from old maas.
* Visit /MAAS/scripts/testing and /MAAS/scripts/commissioning - you should see a list of scripts.
* Confirm that scripts are in state.

### Note
* As this is using the http api, and we have no push updates, the Scripts component deliberately refetches and refreshes scripts on every load.

### DIFF
--- a/ui/src/app/base/sagas/http.js
+++ b/ui/src/app/base/sagas/http.js
@@ -11,7 +11,12 @@ export const api = {
       Accept: "application/json",
       "X-CSRFToken": csrftoken
     };
-    return fetch(SCRIPTS_API, { headers }).then(response => response.json());
+    return fetch(SCRIPTS_API, { headers }).then(response => {
+      if (!response.ok) {
+        throw Error(response.statusText);
+      }
+      return response.json();
+    });
   }
 };
 
@@ -28,7 +33,7 @@ export function* fetchScriptsSaga() {
   } catch (error) {
     yield put({
       type: `FETCH_SCRIPTS_ERROR`,
-      error
+      error: error.message
     });
   }
 }
--- a/ui/src/app/base/sagas/http.js
+++ b/ui/src/app/base/sagas/http.js
@@ -1,0 +1,38 @@
+import { call, put, takeLatest } from "redux-saga/effects";
+
+import getCookie from "./utils";
+
+const SCRIPTS_API = `/MAAS/api/2.0/scripts/`;
+
+export const api = {
+  fetchScripts: csrftoken => {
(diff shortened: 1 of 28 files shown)


squidsoup:
@lilyvidenova Just marking this "design needed" for the "test" vs "testing" issue.

squidsoup:
@Caleb-Ellis if you happen to run your maas and maas-ui on different hosts, be aware that you have to copy *both* your csrftoken and sessionid cookies for the http api requests to succeed. If you run them both in the same place, shouldn't be an issue.

ui/src/app/base/sagas/http.js
```diff
@@ -0,0 +1,38 @@
...
+    yield put({ type: `FETCH_SCRIPTS_START` });
+    response = yield call(api.fetchScripts, csrftoken);
+    yield put({
+      type: `FETCH_SCRIPTS_SUCCESS`,
+      payload: response
+    });
+  } catch (error) {
+    yield put({
+      type: `FETCH_SCRIPTS_ERROR`,
+      error
```
Caleb-Ellis: I didn't get error in state when I forgot to update `env.development.local` because the error was an instance of Error. Maybe add a check here, `error: error instanceof Error ? error.message : error`... or maybe we should add a card to handle errors more thoroughly.

squidsoup: Thanks @Caleb-Ellis, quite right thank you, I've improved the error handling.

ui/src/app/settings/components/Routes/Routes.js
```diff
@@ -53,7 +53,16 @@ const Routes = () => (
       component={NetworkDiscoveryForm}
     />
     <Redirect from="/network" to="/network/proxy" />
-    <Route exact path="/scripts" component={Scripts} />
+    <Route
+      exact
+      path="/scripts/commissioning"
+      render={props => <Scripts {...props} type="commissioning" />}
```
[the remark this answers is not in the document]
squidsoup: Thank @Caleb-Ellis I might land this as is, but will bear that in mind while working on the rest of the view - I might end up adopting your approach as well, although the type prop does seem nice.

Issue #77, opened 2019-08-09
Add delete dhcp snippet confirmation


squidsoup:
* Create a confirmation to delete a dhcp snippet (as an expanding row).
* Create new redux action to delete a dhcp snippet.
* Dispatch action on clicking delete, and close confirmation if successful.

PR #179, opened 2019-09-05
Delete dhcp

huwshimi:
Done:
- Added deleting DHCP snippets from the list. Fixes: #77.
- Move the delete confirmation to a reusable component.

QA:
- Open the dhcp list.
- Click delete and confirm.
- The snippet should be removed from the list and a message should appear.

### DIFF
--- a/ui/src/app/settings/views/Dhcp/DhcpList/DhcpList.js
+++ b/ui/src/app/settings/views/Dhcp/DhcpList/DhcpList.js
@@ -17,6 +17,7 @@ import {
   device as deviceSelectors,
   machine as machineSelectors
 } from "app/base/selectors";
+import { messages } from "app/base/actions";
 import Button from "app/base/components/Button";
 import Code from "app/base/components/Code";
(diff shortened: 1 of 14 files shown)


ui/src/app/settings/views/Dhcp/DhcpList/DhcpList.js
```diff
@@ -222,6 +224,16 @@ const DhcpList = ({ initialCount = 20 }) => {
     dispatch(machineActions.fetch());
   }, [dispatch]);
 
+  useEffect(() => {
+    if (saved && deletingName) {
+      dispatch(
+        messages.add(`${deletingName} removed successfully.`, "information")
```
squidsoup: We should probably make the message types a const somewhere. No need to do that here though.

huwshimi: Good idea, I've added: https://github.com/canonical-web-and-design/maas-ui/issues/181

Issue #202, opened 2019-09-12
Display username in form title when editing


huwshimi:
When users are being edited the username should be shown in the form title, just like the dhcp edit form.

PR #204, opened 2019-09-13
Display username in title when editing

huwshimi:
Done:
- Display the username in the form title when editing a user. Fixes: #202.

QA:
- View an edit user form
- The username should in the title and should change when the username input changes.

### DIFF
--- a/ui/src/app/base/components/FormCard/FormCard.js
+++ b/ui/src/app/base/components/FormCard/FormCard.js
@@ -11,7 +11,7 @@ export const FormCard = ({ children, title }) => {
     <Card highlighted={true} className="card-form">
       <Row>
         <Col size="2">
-          <h4>{title}</h4>
+          <h4 className="card-form__title">{title}</h4>
         </Col>
         <Col size="8">{children}</Col>
       </Row>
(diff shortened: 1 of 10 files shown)


ui/src/app/base/components/FormCard/FormCard.js
```diff
@@ -11,7 +11,7 @@ export const FormCard = ({ children, title }) => {
     <Card highlighted={true} className="card-form">
       <Row>
         <Col size="2">
-          <h4>{title}</h4>
+          <h4 className="card-form__title">{title}</h4>
```
Caleb-Ellis: classname doesn't match - should it be `card-form` or `form-card`?

ui/src/app/base/components/FormCard/FormCard.scss
```diff
@@ -4,6 +4,12 @@
   margin-top: 0.5rem;
 }
 
+.form-card__title {
+  overflow: hidden;
+  text-overflow: ellipsis;
+  white-space: nowrap;
```
Caleb-Ellis: Since "Editing" is always displayed there's not much room for the actual name, so I think this looks better as:
``` scss
.form-card__title {
  word-break: break-word;
}
```

ui/src/app/base/components/FormCard/FormCard.js
huwshimi: Oops!

ui/src/app/base/components/FormCard/FormCard.scss
huwshimi: Good suggestion.

Issue #181, opened 2019-09-05
Define message types


huwshimi:
Set the global message types as constants, see: https://github.com/canonical-web-and-design/maas-ui/pull/179#discussion_r321066272.

PR #205, opened 2019-09-13
Add a hook for adding messages

huwshimi:
QA:
- Set the notification types as constants. Fixes: #181.
- Add a hook for adding messages based on a state change.

QA:
- check a few lists and forms and adding, updating a deleting should all display messages.


ui/src/app/base/components/Notification/Notification.js
```diff
@@ -2,6 +2,13 @@ import classNames from "classnames";
 import PropTypes from "prop-types";
 import React, { useEffect, useRef } from "react";
 
+export const notificationTypes = {
```
Caleb-Ellis: this could probably live in an enum or constants file in `app/base`

PR #216, opened 2019-09-19
Fetch data required for machines list.

Caleb-Ellis:
## Done
- Fetched data required for building machine list with double row.

## QA
- Go the new machine list.
- Check that a loading spinner shows while machines load.
- Check that a bunch of things are fetched and show up in redux, and those things match what are fetched in the angular machines list (check Network tab of developer tools).


ui/src/app/machines/views/MachineList/MachineList.js
```diff
@@ -1,5 +1,67 @@
...
-export default MachineList;
+const DnsForm = () => {
+  const dispatch = useDispatch();
+
+  const machines = useSelector(machineSelectors.all);
+  const machinesLoaded = useSelector(machineSelectors.loaded);
+  const machinesLoading = useSelector(machineSelectors.loading);
+
+  useEffect(() => {
+    if (!machinesLoaded) {
```
huwshimi: I'm not sure if this is necessary, the saga will not fire this again if the data has been fetched and this effect will only rerun if dispatch changes (which it won't in our app as we never swap out the store).

huwshimi: I'll fix this up in my branch.

PR #221, opened 2019-09-24
Add copy to clipboard to the Code component

huwshimi:
Done:
- Add copy to clipboard to the Code component. Fixes: https://github.com/canonical-web-and-design/MAAS-squad/issues/1524.

QA:
- Run `./run exec --expose-port 6060 yarn docs`.
- Visit `[multipass]:6060/#/Code`.
- In the "Copyable" example click on the clipboard icon.
- Paste the contents of your clipboard somewhere, it should match the input.

### DIFF
--- a/ui/src/app/base/components/Code/Code.js
+++ b/ui/src/app/base/components/Code/Code.js
@@ -1,6 +1,6 @@
 import classNames from "classnames";
 import PropTypes from "prop-types";
-import React from "react";
+import React, { useRef } from "react";
 
 const Code = ({
   children,
@@ -10,6 +10,7 @@ const Code = ({
   numbered,
   ...props
 }) => {
+  const input = useRef(null);
   if (inline) {
     return (
       <code className={className} {...props}>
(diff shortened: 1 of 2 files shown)


ui/src/app/base/components/Code/Code.js
```diff
@@ -10,21 +10,34 @@ const Code = ({
...
       </code>
     );
   } else if (copyable) {
+    const handleClick = evt => {
+      input.current.focus();
+      input.current.select();
+      try {
+        document.execCommand("copy");
+      } catch (err) {
+        console.error("Copy was unsuccessful");
```
steverydz: Should we maybe use a logging library such as [loglevel](https://github.com/pimterry/loglevel) for example, rather than directly logging to the console?

huwshimi:
Thanks @steverydz, I've filed https://github.com/canonical-web-and-design/maas-ui/issues/224 to keep track of the logging issue.

I also unassigned Lily as this will get a UX review as part of the MAAS keys epic.

ui/src/app/base/components/Code/Code.js
squidsoup: @steverydz structured logging would be nice if we could log report to sentry, but *what* sentry?

PR #225, opened 2019-09-25
Upgrade react-router-dom to 5.1.0

huwshimi:
Done:
- Upgrade react-router-dom to 5.1.0.

QA:
- Check that routing works as before.


huwshimi:
@Caleb-Ellis thanks, updated the yarn.lock.

PR #227, opened 2019-09-25
Fix redirect for network section of settings

### LINKED ISSUE #220: Network discovery page in settings is unreachable
It seems there's a redirect in place so that `/settings/network/network-discovery` instead goes to `/settings/network/proxy`. It probably shouldn't do that.

steverydz:
## Done
Add the exact property to the redirect in the network section so other views are reachable. Fixes #220

## QA
- Go to the settings section
- Click the **Network discovery** link in the side navigation
- See that you are able to get to that view
- Change the URL so you are only going to `/network`
- See that it still redirects to `/network/proxy`

### DIFF
--- a/ui/src/app/settings/components/Routes/Routes.js
+++ b/ui/src/app/settings/components/Routes/Routes.js
@@ -74,10 +74,11 @@ const Routes = () => {
       />
       <Route
         exact
-        path={`${match.path}/network/network-discovery"`}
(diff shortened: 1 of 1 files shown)


Caleb-Ellis:
nvm figured it out, there's an extra double quote in the Route component for NetworkDiscoveryForm - sneaky bastard!

steverydz:
@Caleb-Ellis good spot - fixed

PR #230, opened 2019-09-27
Display a list of SSL keys

huwshimi:
Done:
- Fetch and list SSL keys. Fixes: https://github.com/canonical-web-and-design/MAAS-squad/issues/1535

QA:
- Add some SSL keys via the old preferences UI (you'll find some examples if you search "BEGIN CERTIFICATE" in the maas code).
- Visit /account/prefs/ssl-keys.
- You should see your keys.
Note: copying to clipboard has not yet been implemented.

### DIFF
--- a/ui/src/app/preferences/actions/index.js
+++ b/ui/src/app/preferences/actions/index.js
@@ -1,1 +1,2 @@
 export { default as sshkey } from "./sshkey";
+export { default as sslkey } from "./sslkey";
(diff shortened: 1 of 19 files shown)


huwshimi:
Thanks @Caleb-Ellis! There's a bug filed on the API to return that string correctly: https://bugs.launchpad.net/maas/+bug/1845640.

PR #234, opened 2019-09-30
Add user details form

huwshimi:
Done:
- Add form to edit your details. Fixes: https://github.com/canonical-web-and-design/maas-squad/issues/1510.
- Link to user prefs from the user list. Fixes: https://github.com/canonical-web-and-design/maas-squad/issues/1499.

QA:
- Click your username in the header.
- Change your details and save.


ui/src/app/preferences/views/Details/DetailsButtons/DetailsButtons.js
```diff
@@ -0,0 +1,36 @@
+import PropTypes from "prop-types";
+import React from "react";
+
+import ActionButton from "app/base/components/ActionButton";
+
+export const DetailsButtons = ({
```
squidsoup: Just curious here, what makes this a `DetailsButtons` component? It seems like a generic component, maybe a `ButtonsPanel`?

huwshimi: Just had another look and the other forms that have a button like this just include ActionButton directly, so I'll update this to do that.

huwshimi: Oh actually now that I've looked at it again, the form needs to take a prop that is a reference to a component hence me having this small component for this one case.

PR #235, opened 2019-09-30
Allow server to parse metadata if header exists.

squidsoup:
## Done
If a script is uploaded without any additional parameters, the server will in fact look for the presence of a metadata header, hence there's no need to parse the yaml clientside other than to check for the presence of the header.

## QA
Upload a script with a metadata header, e.g.

```script
#!/bin/bash
# --- Start MAAS 1.0 script metadata ---
# name: my-script
# title: My Script
# description: An amazing script
# --- End MAAS 1.0 script metadata ---
echo 'This script is amazing'.
```

Ensure the name and description are set appropriately.

Additionally, upload a normal script, without a metadata header, it should upload correctly as well.

### DIFF
--- a/ui/src/app/base/sagas/http.js
+++ b/ui/src/app/base/sagas/http.js
@@ -16,6 +16,17 @@ const handleErrors = response => {
   return response;
 };
 
+const handlePromise = response => {
+  if (response.headers) {
+    const contentType = response.headers.get("Content-Type");
+    if (contentType.includes("application/json")) {
+      return Promise.all([response.ok, response.json()]);
+    } else {
+      return Promise.all([response.ok, response.text()]);
+    }
+  }
+};
+
 export const api = {
   scripts: {
     fetch: csrftoken => {
@@ -38,7 +49,7 @@ export const api = {
         method: "POST",
         body: JSON.stringify({ name, type, script: contents })
       })
-        .then(response => Promise.all([response.ok, response.json()]))
+        .then(handlePromise)
         .then(([responseOk, body]) => {
           if (!responseOk) {
             throw body;
@@ -79,9 +90,13 @@ export function* uploadScriptSaga(action) {
       payload: response
     });
   } catch (errors) {
+    let error = errors;
+    if (typeof error === "string") {
+      error = { "Upload error": error };
+    }
     yield put({
       type: "UPLOAD_SCRIPT_ERROR",
-      errors
+      errors: error
     });
   }
 }
--- a/ui/src/app/settings/views/Scripts/ScriptsUpload.js
+++ b/ui/src/app/settings/views/Scripts/ScriptsUpload.js
@@ -109,7 +109,7 @@ const ScriptsUpload = ({ type }) => {
     isDragReject
   } = useDropzone({
     onDrop,
-    accept: "text/*",
+    accept: "text/*, application/x-csh, application/x-sh",
     maxSize: MAX_SIZE_BYTES,
     multiple: false
   });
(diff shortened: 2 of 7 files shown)


ui/src/app/settings/views/Scripts/ScriptsUpload.js
```diff
@@ -20,8 +20,8 @@ const ScriptsUpload = ({ type }) => {
   const hasErrors = useSelector(scriptSelectors.hasErrors);
   const errors = useSelector(scriptSelectors.errors);
   const saved = useSelector(scriptSelectors.saved);
-  const [savedScript, setSavedScript] = useState();
-  const [script, setScript] = useState();
+  const [savedScript, setSavedScript] = React.useState();
+  const [script, setScript] = React.useState();
```
squidsoup: I wasn't able to find a way to easily mock a named import, hence this change here. I'm usually loathe to make changes to code to facilitate tests, but in this case I couldn't find a solution - suggestions welcome.

Caleb-Ellis: #### ScriptUpload.test.js
``` js
import * as React from "react";
```
Seems to work and then you don't need to use `React.useState();` in `ScriptUpload.js`.

ui/src/app/settings/views/Scripts/ScriptsUpload.js
```diff
@@ -67,15 +67,32 @@ const ScriptsUpload = ({ type }) => {
...
-        setScript({ name: scriptName, script: binaryStr });
+        try {
+          const scriptArray = binaryStr.split("\n");
+          scriptArray.forEach(line => {
+            if (line.includes("--- Start MAAS 1.0 script metadata ---")) {
+              hasMetadata = true;
+            }
+          });
+        } catch {
+          console.error("Unable to parse script for metadata.");
```
Caleb-Ellis: Shouldn't this be an error notification?

squidsoup: Not in this case no, if the metadata can't be parsed we still attempt a normal upload.

squidsoup:
> I keep getting an error `Invalid filetype, please try again` when trying to upload bash scripts, but everything else works fine if I remove `accept: "text/*"` from `useDropzone()`. Am I doing something wrong?

Weird, I can't replicate this. Which browser are you using?

ui/src/app/settings/views/Scripts/ScriptsUpload.js
squidsoup: Nice! That's a much better solution, thank you @Caleb-Ellis!

squidsoup: @Caleb-Ellis hmm, I can't actually get that to work, do you have the hasMetadata test passing?

squidsoup: @Caleb-Ellis based on @hatched's suggestion, I refactored the file reading code out, and mocked it for these tests, which meant I can make assertions against store.getActions() which is much nicer I think.

ui/src/app/base/sagas/http.js
```diff
@@ -79,9 +90,13 @@ export function* uploadScriptSaga(action) {
       payload: response
     });
   } catch (errors) {
+    let error = errors;
+    if (typeof error === "string") {
+      error = { "Upload error": error };
+    }
```
squidsoup: Added some additional error handling for 500s.

ui/src/app/settings/views/Scripts/ScriptsUpload.js
```diff
@@ -92,7 +79,7 @@ const ScriptsUpload = ({ type }) => {
     isDragReject
   } = useDropzone({
     onDrop,
-    accept: "text/*",
+    accept: "text/*, application/x-csh, application/x-sh",
```
squidsoup: This should fix your Firefox issue @Caleb-Ellis

ui/src/app/settings/views/Scripts/ScriptsUpload.js
Caleb-Ellis: Agreed, this is a much more sensible separation!

Issue #238, opened 2019-10-01
Sync SSL key changes


huwshimi:
When SSL keys are added or removed we need to sync those changes to the Redux state.

Currently blocked on: https://bugs.launchpad.net/maas/+bug/1846141.

PR #239, opened 2019-10-01
Add SSL keys form

huwshimi:
Done:
- Add a form for adding SSL keys. Fixes: https://github.com/canonical-web-and-design/maas-squad/issues/1508.

QA:
- Visit /account/prefs/ssl-keys.
- Click "Add SSL key".
- Paste a key (search MAAS for "-----BEGIN CERTIFICATE-----").
- Save and you should be redirected back to the list.
- Refresh and the key should be in the list (waiting on API support for sync: https://github.com/canonical-web-and-design/maas-ui/issues/238).

### DIFF
--- a/ui/src/app/base/actions/user/user.test.js
+++ b/ui/src/app/base/actions/user/user.test.js
@@ -57,7 +57,7 @@ describe("user actions", () => {
   });
 
   it("can handle cleaning users", () => {
-    expect(user.cleanup({ name: "kookaburra" })).toEqual({
+    expect(user.cleanup()).toEqual({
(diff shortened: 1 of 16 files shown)


squidsoup:
@huwshimi might be good to have a design review on this one. While in purely practical terms this doesn't matter, the word wrapping due to the small text area, or size of the text in the text area could be prettier.

image

Might be nice to have the key displayed similar to what you'd expect to see in an editor, e.g. 

image

huwshimi:
@squidsoup as discussed I've made a note in zeplin about this, we'd also need to set the font to mono so the lines were equal length.

Issue #193, opened 2019-09-10
Provide UI feedback if websocket disconnects & reconnects


squidsoup:
Having implemented `reconnecting-websocket` in https://github.com/canonical-web-and-design/maas-ui/pull/192 it would be nice if we also provided some UI feedback when the connection drops and reconnects.

PR #270, opened 2019-10-11
WS disconnect notification

huwshimi:
Done:
- Notify the user if the websocket disconnects. Fixes: #193.

QA:
- Load the UI.
- Stop your MAAS, the UI should update.
- Restart your MAAS, the UI should eventually return.


squidsoup:
This is awesome, thank you @huwshimi! Works brilliantly.

Issue #159, opened 2019-08-27
Loading state for tables


lilyvidenova:
All tables (in Settings to begin with) should follow the same loading pattern: https://zpl.io/ag6ZPQM 

![image](https://user-images.githubusercontent.com/17980375/63760552-74766f80-c8b7-11e9-9009-2d53bff186ce.png)

PR #272, opened 2019-10-11
Table loading state

huwshimi:
Done:
- Update settings and preferences lists to use a shared table component.
- Add new loading state. Fixes: #159.

QA:
- Click through all the lists in settings and preferences.
- There should be a new loading state (see: #159).
- Everything else should function as it did before.


huwshimi:
Thanks @Caleb-Ellis I've addressed both of those issues.

Issue #31, opened 2019-07-22
Add window title


huwshimi:
Display the current page in the window title + fetch and display the maas name in the format `[page name] | [maas name] MAAAS` e.g. "Settings | mymaas MAAS".

steverydz:
I was going to tackle this issue but haven't bee able to get the app running.

I've copied `.env.development` to `.env.development.local` and put the following inside it:

```
REACT_APP_BASENAME="/MAAS/settings"
REACT_APP_MAAS_URL="http://maas.local:5240/MAAS",
REACT_APP_WEBSOCKET_URL="ws://maas.local:5240/MAAS/ws"
```

I had my local MAAS running. I then ran `./run` from the root of the project. The development server started but in the browser all I got was a message: `Failed to connect. Please try refreshing your browser.`

I couldn't find the info on how to hack a local MAAS, but I understood that wasn't necessary anymore.

Not sure if I'm missing any steps or misunderstanding something here?

squidsoup:
@steverydz are you running maas and maas-ui from the same container? If not, you'll need to manually add a csrf token.

steverydz:
@squidsoup Ah, thanks. That has worked.

PR #273, opened 2019-10-11
Display window titles

huwshimi:
Done:
- Display window titles for each page. Fixes: #31.

QA:
- Click around, the window title should change accordingly.

### DIFF
--- a/ui/src/app/settings/views/Repositories/RepositoriesList/RepositoriesList.js
+++ b/ui/src/app/settings/views/Repositories/RepositoriesList/RepositoriesList.js
@@ -7,6 +7,7 @@ import { packagerepository as repositoryActions } from "app/base/actions";
 import { packagerepository as repositorySelectors } from "app/base/selectors";
 import { getRepoDisplayName } from "../utils";
 import { useAddMessage } from "app/base/hooks";
+import { useWindowTitle } from "app/base/hooks";
 import Button from "app/base/components/Button";
 import Loader from "app/base/components/Loader";
 import MainTable from "app/base/components/MainTable";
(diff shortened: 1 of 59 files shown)


ui/src/app/settings/views/Repositories/RepositoriesList/RepositoriesList.js
```diff
@@ -100,6 +101,9 @@ export const Repositories = () => {
     repositorySelectors.search(state, searchText)
   );
   const dispatch = useDispatch();
+
+  useWindowTitle("Repositories");
```
Caleb-Ellis: This is "Package repos" in the sidenav - should we keep them consistent?

ui/src/app/settings/views/Scripts/ScriptsList.js
```diff
@@ -155,6 +156,9 @@ const ScriptsList = ({ type = "commissioning" }) => {
   const userScripts = useSelector(state =>
     scriptSelectors.search(state, searchText, type)
   );
+
+  useWindowTitle(`${type} scripts`);
```
Caleb-Ellis: Might need to capitalize the type here because it's showing as e.g "commissioning scripts"

ui/src/app/settings/views/Repositories/RepositoriesList/RepositoriesList.js
huwshimi: Thanks!

ui/src/app/settings/views/Scripts/ScriptsList.js
huwshimi: Ah yeah, good spot, thanks.

PR #278, opened 2019-10-16
Enable logging out

huwshimi:
Done:
- Enable logging the user out. Fixes: https://github.com/canonical-web-and-design/maas-squad/issues/1526

QA:
- Log into the UI.
- Click Logout in the top right.
- You should see the login form.
- Log in again, you should be correctly logged in again.

### DIFF
--- a/ui/src/app/base/components/Header/Header.js
+++ b/ui/src/app/base/components/Header/Header.js
@@ -1,10 +1,11 @@
 import { Link } from "react-router-dom";
-import { useSelector } from "react-redux";
+import { useDispatch, useSelector } from "react-redux";
 import classNames from "classnames";
 import React, { useState } from "react";
 
 import "./Header.scss";
 import { auth } from "app/base/selectors";
+import { status as statusActions } from "app/base/actions";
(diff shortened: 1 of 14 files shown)


ui/src/app/base/components/Header/Header.js
```diff
@@ -213,7 +215,11 @@ export const Header = () => {
               <li className="p-navigation__link" role="menuitem">
                 <a
                   className="p-dropdown__item"
-                  href={generateURL("accounts/logout/")}
+                  href="##"
```
steverydz: Why `##` rather than `#`?

huwshimi: To satisfy the linter :)

steverydz: Ah ok. Maybe we should override that rule if we have to hack around it?

huwshimi: Yeah good idea, I've filed: https://github.com/canonical-web-and-design/maas-ui/issues/284

PR #281, opened 2019-10-16
Only restrict settings to admins

huwshimi:
- Allow non-admins to view everything but settings. Fixes: https://github.com/canonical-web-and-design/maas-squad/issues/1536.

QA:
- Log in as an admin, you should be able to see settings, machines and preferences.
- Log in as a non-admin, you should be able to see machines and preferences but not settings.


steverydz:
http://maas.local:8000/MAAS/api/2.0/account/?op=list_authorisation_tokens

PR #308, opened 2019-10-20
Fix legacy tests

squidsoup:
## Done
Misc test fixes as a result of directory structure changes and upgrading to the Angularjs LTS.

## QA
Run `yarn test`. Tests should pass.


legacy/src/app/controllers/node_details.js
```diff
@@ -1419,7 +1419,7 @@ function NodeDetailsController(
     }
     if (node.cpu_speed) {
       const speedText =
-        node.cpu_speed > 1000
+        node.cpu_speed >= 1000
```
squidsoup: Fixed a small bug here!

PR #323, opened 2019-10-24
Configure production build

squidsoup:
## Done
* Configure production build with html, css and js optimisation and hashed assets.
* Configured html-loader to ignore angularjs template tags (e.g. `{$ foo $}` and `{{ bar }}`).
* Fixed a template bug in _script-runtime.html_ which was breaking the minification parser.

## QA
* Run `yarn clean-build`
* Run `yarn build`, ensure assets are correctly output to `./dist` and minified.

### DIFF
--- a/legacy/webpack.config.js
+++ b/legacy/webpack.config.js
@@ -1,8 +1,9 @@
 const webpack = require("webpack");
 const CopyWebpackPlugin = require("copy-webpack-plugin");
+const TerserJSPlugin = require("terser-webpack-plugin");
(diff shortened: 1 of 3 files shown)


legacy/webpack.config.js
```diff
@@ -37,7 +41,15 @@ module.exports = {
       },
       {
         test: /\.html$/,
-        use: ["html-loader"]
+        use: {
+          loader: "html-loader",
+
```
huwshimi: Extra space here, but whatevs.

Issue #327, opened 2019-10-24
Remove the React machine list route (for now)


Caleb-Ellis:
When in the React side of things, clicking "Machines" in the nav takes you to the almost-complete-but-not-quite React machine list. It should probably go to the angular machine list.

Issue #325, opened 2019-10-24
Redirect all unknown paths to the angular machine list


Caleb-Ellis:
At the moment, going to `http://<maas-ip>:8400` returns a 404 - you have to specifically navigate to `http://<maas-ip>:8400/MAAS/` so we should add a redirect for this case. Similarly, we should have a catch-all redirect for any unknown paths.

Issue #280, opened 2019-10-16
Add/Fix Header and Footer in index.html


squidsoup:
Previously the header and footer were templated by django. Currently a rendered header from my dev mass is just copy/pasted into index to allow for navigating around the legacy client, but this needs to be tidied up significantly. 

This will require making API requests for maas name and version to display in the footer.

PR #336, opened 2019-10-25
Share header

huwshimi:
Done:
- Add a shared package in the monorepo including the Header.
- Use the shared header for both the ui and legacy projects. Part of https://github.com/canonical-web-and-design/maas-ui/issues/280
- Redirect / to /MAAS in the proxy. Fixes: https://github.com/canonical-web-and-design/maas-ui/issues/325.
- Link to the angular machine list in the header. Fixes: https://github.com/canonical-web-and-design/maas-ui/issues/327.

QA:
- If you have a legacy/.env.local or a ui/.env.development.local then either update or remove them.
- In the root run `./run`.
- Open the site and check that in the header you can navigate between the react/angular clients and that you see your username in the header and can log out.


huwshimi:
@squidsoup comparing the CI runs from this branch and the previous one the time went from 3:00 to 3:08. Makes sense because it's only building one component.

PR #342, opened 2019-10-29
Add a shared footer

huwshimi:
Done:
- Add a footer component.
- Display the footer in both the angular and react clients. Fixes: https://github.com/canonical-web-and-design/maas-ui/issues/280.
- Drive-by fix to add lint/test to the top level package.

QA:
- Login and view the react and angular clients.
- Both should display the footer.


legacy/src/app/entry.js
```diff
@@ -328,8 +328,7 @@ function unhideRSDLinks() {
   rsdLinks.forEach(link => link.classList.remove("u-hide"));
 }
 
-/* @ngInject */
-const displayTemplate = ($rootScope, $window, $http) => {
+const renderHeader = ($window, $http) => {
```
squidsoup: This should be decorated with `/* @ngInject */`

ngAnnotate does try to automatically find params that need to be injected, and it must be working in this case, but it doesn't always, so best to be explicit.

huwshimi: As discussed, this should be fine as the injection is set to `explicitOnly`.

PR #343, opened 2019-10-29
Configure app from websocket API.

squidsoup:
## Done
* Configure app via websocket API, and manually bootstrap (removes *angular-deferred-bootstrap*).
* Added basic loading component to display before app bootstraps.

## QA
Load app, navigate about. There should be no console errors. Data set from config like version/app name should display correctly.

### DIFF
--- a/legacy/src/index.html
+++ b/legacy/src/index.html
@@ -42,7 +42,6 @@
     </header>
     <main class="u-no-margin--top" id="main-content" data-maas-error-overlay>
       <div class="wrapper--inner">
-        <h3 ng-loading>Loading MAAS...</h3>
         <div class="ng-hide"></div>
       </div>
       <div class="u-no-margin--top" data-ng-view></div>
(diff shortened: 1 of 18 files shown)


legacy/src/index.html
```diff
@@ -36,14 +36,9 @@
     <div id="header"></div>
     <main class="u-no-margin--top" id="main-content" data-maas-error-overlay>
       <div class="wrapper--inner">
+        <h3 ng-loading>Loading MAAS...</h3>
```
huwshimi: Here's a suggestion for making this look a bit more consistent:

```jsx
<header ng-loading class="p-strip--light is-shallow u-no-padding--bottom page-header">
  <div class="row">
    <h1 class="page-header__title">Loading MAAS...</h1>
  </div>
</header>
```

huwshimi:
@squidsoup I'm not sure if this is working correctly. I think the window title is not updated so maybe that's a sign the config is being set after the app is loaded.

PR #346, opened 2019-10-29
Remove labeler action until issue with PRs from forks resolved.

squidsoup:
## Done 
Remove labeler action until issue labelling PRs from forks resolved https://github.com/actions/labeler/issues/12

## QA
There should be no labelling action below.


samajammin:
@squidsoup I tried this same approach but am still getting failures w/ my labeler. Have you been able to solve it? From what I can see you're still having the same issues w/ PRs. Please let me know if you do!

PR #361, opened 2019-10-30
Set ActionButton dimensions on loading state transition.

Caleb-Ellis:
## Done
- Updated ActionButton so that it explicitly sets width *and* height when transitioning to loading, instead of just width on mount. This means that if for some reason the content of the button changes it will actually be the correct height/width as according to Vanilla and will only store the dimensions for the loading => success phase.

## QA
- Go to any Settings form, make changes and save. Check that the button dimensions stay consistent.
- Open the inspector and change the button text to something else. Save again and check that the dimensions are consistent.

### DIFF
--- a/ui/src/app/base/components/ActionButton/ActionButton.js
+++ b/ui/src/app/base/components/ActionButton/ActionButton.js
@@ -17,23 +17,22 @@ const ActionButton = ({
   success,
   ...props
 }) => {
-  const [width, setWidth] = useState(0);
+  const [height, setHeight] = useState();
+  const [width, setWidth] = useState();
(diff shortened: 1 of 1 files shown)


ui/src/app/base/components/ActionButton/ActionButton.js
```diff
@@ -89,7 +90,14 @@ const ActionButton = ({
       className={buttonClasses}
       disabled={disabled}
       ref={ref}
-      style={width ? { width: `${width}px` } : undefined}
+      style={
```
Caleb-Ellis: This is a bit awkward tbh but it means that instead of the `<button>` having an empty style attribute it has no style attribute at all

PR #365, opened 2019-10-30
Update config key for maas_uuid.

squidsoup:
## Done
MAAS was using an incorrect key for maas_uuid (fixed here https://code.launchpad.net/~blake-rouse/maas/+git/maas/+merge/374934), so update the client to reflect this.

## QA
Start the app, in the console inspect `window.CONFIG`, it should have a valid uuid instead of an empty string.

### DIFF
--- a/legacy/src/app/bootstrap.js
+++ b/legacy/src/app/bootstrap.js
@@ -70,7 +70,7 @@ const bootstrapOverWebsocket = config => {
         const requiredConfigKeys = [
           "completed_intro",
           "maas_name",
-          "maas_uuid",
+          "uuid",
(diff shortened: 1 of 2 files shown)


legacy/src/app/bootstrap.js
```diff
@@ -70,7 +70,7 @@ const bootstrapOverWebsocket = config => {
         const requiredConfigKeys = [
           "completed_intro",
           "maas_name",
-          "maas_uuid",
+          "uuid",
```
huwshimi: In entry.js there are a couple of references to `$window.CONFIG.maas_uuid` which will need to be updated.

PR #406, opened 2019-11-06
Disable Add License key button if no valid OSes.

### LINKED ISSUE #397: No back or close button in white card views
image

squidsoup:
## Done
Disable "Add License key" button if no valid licensed operating systems are synced.

## QA
* Delete any licensed images should you have one (e.g. a fake Windows server image)
* Visit the license key page in settings. The add button should be disabled and display a tooltip.

fixes #397

### DIFF
--- a/ui/src/app/settings/components/SettingsTable/SettingsTable.js
+++ b/ui/src/app/settings/components/SettingsTable/SettingsTable.js
@@ -35,11 +35,20 @@ export const SettingsTable = ({
         ) : (
           <div className="p-table-actions__space-left"></div>
         )}
-        {buttons.map(({ label, url }) => (
-          <Button element={Link} to={url} key={url}>
-            {label}
-          </Button>
-        ))}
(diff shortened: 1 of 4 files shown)


ui/src/app/settings/components/SettingsTable/SettingsTable.js
```diff
@@ -35,11 +35,20 @@ export const SettingsTable = ({
         ) : (
           <div className="p-table-actions__space-left"></div>
         )}
-        {buttons.map(({ label, url }) => (
-          <Button element={Link} to={url} key={url}>
-            {label}
-          </Button>
-        ))}
+        {buttons.map(({ label, url, disabled = false, tooltip }) =>
```
Caleb-Ellis: I wonder if it would be a good idea to include a tooltip prop for the Button component itself? I've noticed it's quite a common pattern in MAAS and RBAC to disable a button (or checkbox) and have a tooltip display why.

Issue #417, opened 2019-11-06
Build result includes absolute URL to assets


blakerouse:
In the process of integrating this into the maas development as a submodule. I have the building working, so thats great!

Issue is the build results include references to assets with an initial '/' which tells the browser to load the assets from the root of the domain. This does not work as it should be relative to the full path not the domain.

Output of ui/index.html, note the `<link href="/static/css/2.f4eb038f.chunk.css" rel="stylesheet">` and others in the file including the favicon.

```
<!doctype html>
<html lang="en">

<head>
    <meta charset="utf-8" />
    <title>MAAS</title>
    <link rel="shortcut icon" href="/maas-favicon-32px.png" />
    <link href="/static/css/2.f4eb038f.chunk.css" rel="stylesheet">
    <link href="/static/css/main.f0cbae66.chunk.css" rel="stylesheet">
</head>

<body><noscript>You need to enable JavaScript to run this app.</noscript>
    <div id="root">
        <header class="p-navigation is-dark">
            <div class="p-navigation__row row">
                <div class="p-navigation__banner">
                    <div class="p-navigation__logo"><svg xmlns="http://www.w3.org/2000/svg" width="100" height="25.2"
                            viewBox="545.3 412.6 100 25.2" alt="" class="p-navigation__image">
                            <path fill="#E95420"
                                d="M557.9 412.6c-7 0-12.6 5.7-12.6 12.6 0 7 5.7 12.6 12.6 12.6 7 0 12.6-5.7 12.6-12.6 0-7-5.6-12.6-12.6-12.6z">
                            </path>
                            <g fill="#FFF">
                                <path
[...]

PR #420, opened 2019-11-07
Fix static paths

huwshimi:
Done:
- Set the public url for the built react app.
- Set the correct path for the legacy favicon.
Fixes: #https://github.com/canonical-web-and-design/maas-ui/issues/417.

QA:
- `yarn build-all`
- In ./build/ui/index.html all the hrefs should start with /MAAS/.
- In ./build/legacy/index.html the favicon should start with /MAAS/.


huwshimi:
@squidsoup I've updated this to build the legacy css/js to assets/

blakerouse:
Thanks for the move of the CSS and the JS into the assets folder, makes it easier to integrate.

PR #430, opened 2019-11-07
Change hardcoded paths to variables

### LINKED ISSUE #422: Replace hardcoded paths with vars
Replace any references to "/MAAS" or "/#/" or "/r/" with `process.env` variables.

steverydz:
## Done
Changes hardcoded paths including `/MAAS`, `/#` and `/r` to use environment variables. Fixes #422.

## QA
- Go to a pod details page and compose a machine
- Click the link which reads `More about spaces` and see that the link works as expected
- Login and out and make sure that works as expected


steverydz:
@huwshimi I haven't been able to change the paths in `proxy/index.js` without breaking the app. It seems like it doesn't have access to `process.env`.

huwshimi:
This is looking great @steverydz!

I just tried and the process.env variables are working for me for the proxy. It currently uses one for `process.env.MAAS_URL`. You will however have to redefine `BASENAME="/MAAS"` and `REACT_BASENAME="/r"` etc. inside ./proxy/.env.

Is it possible you have a ./proxy/env.local or something that's overwriting the changes? Are you running it with `./run`? I'm not sure what could be causing it to not see `process.env`.

steverydz:
Thanks @huwshimi, it was that I didn't have the variables in `proxy/.env`

Issue #382, opened 2019-11-01
Missing RSD tab in navigation


Caleb-Ellis:
Looks like it shows up as part of a field in `config` in state.

huwshimi:
Fetch over the ws from `general.navigation_options`.

PR #435, opened 2019-11-08
Show or hide RSD nav

huwshimi:
Done:
- Only display the RSD tab when enabled. Fixes: https://github.com/canonical-web-and-design/maas-ui/issues/382.

QA:
- Run the UI, you probably won't have the RSD option in your header.
- Fake the showRSD option in App.js or entry.js (unless you have a MAAS with RSD).

### DIFF
--- a/legacy/src/app/bootstrap.js
+++ b/legacy/src/app/bootstrap.js
@@ -94,7 +94,7 @@ const bootstrapOverWebsocket = () => {
         messagesReceived.push(3);
         break;
 
-      // general.version
+      // general.navigation_options
       case 4:
         config.navigation_options = msg.result;
         messagesReceived.push(4);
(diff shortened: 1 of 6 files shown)


legacy/src/app/bootstrap.js
```diff
@@ -90,8 +90,14 @@ const bootstrapOverWebsocket = config => {
         config.version = msg.result;
         messagesReceived.push(3);
         break;
+
+      // general.version
```
steverydz: I think this comment is wrong. Shouldn't it be `general.navigation_options`?

ui/src/app/App.test.js
```diff
@@ -223,4 +223,17 @@ describe("App", () => {
     );
     expect(wrapper.find("Login").exists()).toBe(true);
   });
+
+  it("can show the RSD link", () => {
```
steverydz: I think we should also have a test to make sure the RSD link isn't displayed

legacy/src/app/bootstrap.js
steverydz: I think this comment is wrong. Shouldn't it be `general.navigation_options`?

ui/src/app/App.test.js
huwshimi: Done.

PR #453, opened 2019-11-12
Fix styles for autocomplete list

### LINKED ISSUE #381: Weird script select + tags input styling
## maas:
screenshot_24

## maas-ui:
screenshot_22

steverydz:
## Done
Remove default list styles from autocomplete list. Fixes #381

## QA
- On the machine list page select all the machines
- Select all machines that can be tagged (tip: select all machines, select tag from the take action menu and update your selection)
- When entering tags into the tag input see that the suggestion list doesn't have the default bulleted list style

### DIFF
--- a/legacy/src/scss/_patterns_tag-input.scss
+++ b/legacy/src/scss/_patterns_tag-input.scss
@@ -29,6 +29,8 @@
         min-width: 160px;
         width: 100%;
         box-sizing: border-box;
+        list-style: none;
+        padding-left: 0;
 
         .suggestion-item {
           float: left;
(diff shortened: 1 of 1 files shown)


steverydz:
> This fixes the bullet styling but the X's to remove items still look strange.
> 
> Anyway do we know why this has changed in the first place? All the scss has been copy+pasted over so I would assume it would all be identical.

It's actually not a problem with the SCSS. The way the directive is used has changed. It's now done [the proper way](https://github.com/canonical-web-and-design/maas-ui/pull/286). Unfortunately this means all the customisation that was done to the script that used to be in the old MAAS codebase has gone.

PR #469, opened 2019-11-12
Fix button link active state colour

### LINKED ISSUE #462: Machine summary: p-button--link not setting overriding all states
The custom  class .p-button--link does not override all the necessary states to turn a button into a link.
As a result, it looks like this when pressed:

image

For comparison, this is what pressed links look like in vanilla:

image

steverydz:
## Done
Overridden Vanilla button styles for button links which are in both active and hover states to remove the grey background. Fixes #462.

## QA
- Go to the machine summary page
- Click on one of the test links e.g. "Test CPU..."
- See that as you click the background remains the same colour

### DIFF
--- a/legacy/src/scss/_patterns_button.scss
+++ b/legacy/src/scss/_patterns_button.scss
@@ -75,6 +75,11 @@ $vanilla-2-icon-button-side-padding: $sph-inner--small * 1.5;
     &:visited {
       color: $color-link;
     }
+
+    // Override inherited button style from Vanilla
+    &:active {
+      background-color: inherit !important;
+    }
   }
 
   [class*="p-button"] [class^="p-icon"],


legacy/src/scss/_patterns_button.scss
```diff
@@ -70,6 +70,11 @@ $vanilla-2-icon-button-side-padding: $sph-inner--small * 1.5;
     &:hover {
       background-color: inherit;
       text-decoration: underline;
+
+      // Override inherited button style from Vanilla
+      &:active {
```
Caleb-Ellis: I think it's okay to move this outside `&:hover` with an `!important` because it's an explicit override. As it is currently you can click on the button, hold the click and move the cursor outside and the button turns grey.

I'm thinking maybe this should make it's way into Vanilla because we're using a lot of links like buttons across the apps. And I'm not a fan of using *actual* links where you have to `event.preventDefault()` and `href="##"` or whatever.

steverydz: Done. I definitely agree about not using links for things that aren't links. I'll propose it to Vanilla for the next working group meeting.

PR #471, opened 2019-11-13
Display required indicators when adding machine from machine list.

### LINKED ISSUE #454: Add machine: required fields indistinguishable from optional fields
image

When I try to add a machine, the buttons only get enabled when MAC is filled in. 
(found through trial and error). 

[] Should there be an indication of which fields are required, or alternatively, which fields are optional. 
[] Shouldn't skipping required fields trigger a tooltip like this one?

image

[] The tooltip above appears on a seemingly optional field, and it only happens when you add a new MAC address field. Could it be that this is appearing on the wrong field?

[] Should there be placeholders to indicate the format of the information that needs to be typed in? Or are these familiar to admins? E.g. "Power address".

Caleb-Ellis:
## Done
- Show required asterisks in "Add machine" dropdown
- Disable save button if one of the required fields is missing

## QA
- Add a machine from the machine list
- Check that MAC address is required
- Change the power type and check that required fields show up correctly
- Check the submit button is disabled if some of the required fields are not filled in

Fixes #454

### DIFF
--- a/legacy/src/app/controllers/add_hardware.js
+++ b/legacy/src/app/controllers/add_hardware.js
@@ -368,18 +368,15 @@ function AddHardwareController(
   }
 
   // Validate that all the parameters are there for the given power type.
-  function powerParametersHasError(power_type, parameters) {
-    var i;
-    for (i = 0; i < power_type.fields.length; i++) {
-      var field = power_type.fields[i];
-      var value = parameters[field.name];
+  function powerParametersHasError(power) {
+    const { parameters, type } = power;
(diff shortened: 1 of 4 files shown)


lyubomir-popov:
Is the tooltip that used to appear in the right column of fields when you add a new mac field fixed by this pr?

Caleb-Ellis:
@lyubomir-popov Oops, I thought I fixed it by making sure the form can't be submitted if required fields are missing. Turns out the "Add MAC" button was submitting the form prematurely.

PR #473, opened 2019-11-13
Fix form button spacing Vanilla override.

### LINKED ISSUE #464: Inline style not adhering to "Single direction margin declaration" convention used in Vanilla
image

[Single direction margin declaration](https://csswizardry.com/2012/06/single-direction-margin-declarations/) is a css best practice aimed to eliminate margin collapse among other things.

In vanilla, we apply margins at the bottom of elements.

Often the last element in a group requires more space. One way to do that with vanilla (admittedly it requires a lot of classes, but doesn't require custom css): 
```.p-strip.is-shallow.u-no-padding--top```

I can see this in maas-ui, which is an acceptable alternative, but it should ideally pull the value from one of the predefined values in maas, like the padding of the shallow strip,  so that it stays consistent with other elements: 

```
.p-form__control:last-of-type {
    margin-bottom: 0.5rem;
}

```

In the case of the screenshot above, this small margin is subject to margin collapse, and the bigger inline margin "swallows" it.  Why not just increase the .p-form__control:last-of-type margin and get rid of the inline one? Are there cases where that leads to too much space so you resorted to an inline style?

Just trying to understand the use case here, this might be something vanilla should make easy to achieve without any custom code.

### LINKED ISSUE #465: Settings>Network>Proxy: Form markup issues
Two issues:

1. Sibling radio buttons in vanilla sit closer together than they appear in maas:

image

This is because each pair input+label is in a separate parent. Is this required for validation?

2. The preceding text is wrapped in a label. A label has smaller padding, because it expects a field to follow it. Wrapping it in a `<p>` would result in some more spacing under it, further aiding correct visual grouping (description + set of radios + button)

Comparison of 1 and 2 before/after suggested changes:

Before:
image

After:
image

### PR DESCRIPTION
**Pull request by Caleb-Ellis**
## Done
- Wrap form fields in a parent div, so the resultant markup of a FormikForm becomes:
``` html
<form>
    <div>
        ...form fields...
    </div>
    <div>
        ...buttons...
    </div>
</form>
```
This makes it possible to apply `:last-of-type` to the last form field in which to apply `margin-bottom`.

screenshot_27

Fixes #464, fixes #465

### DIFF
--- a/ui/src/app/base/components/FormikFormButtons/FormikFormButtons.js
+++ b/ui/src/app/base/components/FormikFormButtons/FormikFormButtons.js
@@ -6,11 +6,10 @@ export const FormikFormButtons = ({
   actionDisabled,
   actionLabel,
   actionLoading,
-  actionSuccess,
-  showCancel
+  actionSuccess
 }) => {
   return (
-    <div style={{ marginTop: "1rem" }}>
+    <div>
       <ActionButton
         appearance="positive"
         className="u-no-margin--bottom"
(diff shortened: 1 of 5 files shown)


lyubomir-popov:
Thanks, this fixes 464, but the markup issues in 465 are still there: image

PR #475, opened 2019-11-14
Reset user preference form values on submit.

### LINKED ISSUE #474: User preferences form does not reset values after save
The user preferences form does not reset values after save. This means the Formik form keeps the old values in its internal state, i.e. change username from "foo" to "bar", the form still thinks that "foo" is the username so the button is enabled even though nothing has changed since save.

Caleb-Ellis:
## Done
- Call Formik resetForm after UserPreferences form submit

## QA
- Go to user prefs page
- Change username and check that the nav updates, a notification appears, and the button becomes disabled whenever the text input matches the current username

Fixes #474

### DIFF
--- a/ui/src/app/base/components/UserForm/UserForm.js
+++ b/ui/src/app/base/components/UserForm/UserForm.js
@@ -86,7 +86,7 @@ export const UserForm = ({
       buttons={buttons}
       errors={errors}
       initialValues={initialValues}
-      onSubmit={values => {
+      onSubmit={(values, { resetForm }) => {
         const [firstName, ...lastNameParts] = values.fullName.split(" ");
         const params = {
           email: values.email,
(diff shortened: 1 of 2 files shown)


Caleb-Ellis:
@steverydz that's a separate issue I feel. All that's being fixed here is that the save button is disabled on save now that the form values are reset correctly. The notifications are handled with the `useAddMessage` hook and the `NotificationsList` component. Maybe we should create a new issue to have more flexible notifications i.e. allow single or temporary notifications?

steverydz:
> @steverydz that's a separate issue I feel. All that's being fixed here is that the save button is disabled on save now that the form values are reset correctly. The notifications are handled with the `useAddMessage` hook and the `NotificationsList` component. Maybe we should create a new issue to have more flexible notifications i.e. allow single or temporary notifications?

Fair enough. You have my blessing in that case :pray:

Issue #499, opened 2019-11-21
Add in table actions and icons to the machines table


steverydz:
The previous machines table has action dropdowns for certain fields and icons to represent statuses.

![](https://user-images.githubusercontent.com/361637/69291961-d8ab9400-0c58-11ea-8dae-5a3c7dbc40aa.png)

huwshimi:
This work will be tracked as part of this epic: https://github.com/canonical-web-and-design/MAAS-squad/issues/1696.

Issue #498, opened 2019-11-21
Add secondary table headers and values in the machines table


steverydz:
Add the secondary (sub) table headers and their corresponding values to the machines table.

Note: This will require an adjustment to the `MainTable` component from https://github.com/canonical-web-and-design/react-components.

![](https://user-images.githubusercontent.com/361637/69291961-d8ab9400-0c58-11ea-8dae-5a3c7dbc40aa.png)

huwshimi:
@steverydz would you mind filing issues for the changes to MainTable?

Also to note for whoever picks this up, we have a header toggle in the user list component which could probably be turned into a reusable component to handle some of this: https://github.com/canonical-web-and-design/maas-ui/blob/master/ui/src/app/settings/views/Users/UsersList/UsersList.js#L160

huwshimi:
@steverydz Oops, I totally missed that you already filed https://github.com/canonical-web-and-design/react-components/issues/40.

steverydz:
@huwshimi This might not be an issue now as @Caleb-Ellis [pointed out that components can be passed into the content param](https://github.com/canonical-web-and-design/react-components/issues/40#issuecomment-557181707).

huwshimi:
This work will be tracked as part of this epic: https://github.com/canonical-web-and-design/MAAS-squad/issues/1747

PR #500, opened 2019-11-21
Add columns to machines table

### LINKED ISSUE #495: Add machine columns
Add the columns to match the existing machine list.

Screen Shot 2019-11-21 at 12 16 51 pm

steverydz:
## Done
Add all columns to the machines and populated the top level headers and values. Fixes #495 

Note that this PR doesn't add the second row of headers and values or in-table actions and icons because that requires adjustments to the [MainTable react component](https://canonical-web-and-design.github.io/react-components/?path=/story/maintable--default-story).

There are issues to add these:
- https://github.com/canonical-web-and-design/maas-ui/issues/498
- https://github.com/canonical-web-and-design/maas-ui/issues/499

There is also an issue to add the required flexibility to the `MainTable` component:
- https://github.com/canonical-web-and-design/react-components/issues/40

## QA
- Go to the machines listing
- Compare the table columns with the old machine listing (just the top row as mentioned above)


ui/src/app/machines/views/MachineList/MachineList.js
```diff
@@ -24,6 +24,30 @@ import { machine as machineSelectors } from "app/base/selectors";
 import { nodeStatus } from "app/base/enum";
 import { useWindowTitle } from "app/base/hooks";
 
+const formatStorageUnit = gb => {
```
Caleb-Ellis: Could these live in a utility file somewhere? We use unit conversion functions in like a million places in the old angular code, each written slightly differently. Maybe we could start centralising some of these patterns.

steverydz: Yeah I think you're right. I pretty much just lifted these from the old codebase which was a bit lazy of me :roll_eyes:

steverydz:
@Caleb-Ellis I've moved the formatting stuff to a utils directory so this is ready for review again

PR #501, opened 2019-11-21
Pluralise machine counts in machines table

steverydz:
## Done
Pluralise machine counts in the machines table e.g. "1 machine" instead of "1 machines". Fixes #479.

## QA
- Go to the machines listing
- See that each group that has multiple machines has a pluralised count e.g. "5 machines"
- See that each group that has only one machine has a singular count e.g. "1 machine"

### DIFF
--- a/ui/package.json
+++ b/ui/package.json
@@ -13,6 +13,7 @@
     "http-proxy-middleware": "0.20.0",
     "immer": "5.0.0",
     "path-parse": "1.0.6",
+    "pluralize": "^8.0.0",
     "prop-types": "15.7.2",
     "react": "16.12.0",
     "react-dom": "16.12.0",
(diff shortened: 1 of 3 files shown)


ui/src/app/machines/views/MachineList/MachineList.js
```diff
@@ -60,6 +60,14 @@ const getSortValue = (machine, currentSort) => {
   }
 };
 
+const pluraliseMachineCount = count => {
```
Caleb-Ellis: I reckon we should just use pluralize and do away with having hundreds of little pluralizing functions everywhere: https://www.npmjs.com/package/pluralize
It's obviously heavier than ternaries, but... it's just neater.

steverydz: Good call. Done

ui/package.json
```diff
@@ -13,6 +13,7 @@
     "http-proxy-middleware": "0.20.0",
     "immer": "5.0.0",
     "path-parse": "1.0.6",
+    "pluralize": "^8.0.0",
```
Caleb-Ellis: should be an absolute dependency `"pluralize": "8.0.0"`

PR #528, opened 2019-12-02
More explicit messages when changing bmc node config

tai271828:
This pull request will elaborate bmc behavior more to reduce confusion like this bug report https://bugs.launchpad.net/maas/+bug/1810796

### DIFF
--- a/legacy/src/app/partials/node-details.html
+++ b/legacy/src/app/partials/node-details.html
@@ -554,7 +554,7 @@
                         <p class="p-notification__response">This power controller
                         manages {$ node.power_bmc_node_count - 1 $} other
                         {$ node.power_bmc_node_count > 3 ? "nodes" : "node" $}.
-                        Changing power parameters will affect these nodes.</p>
+                        Changing the IP address or outlet delay will affect all these nodes.</p>
                     </div>
                 </div>
             </div>


lilyvidenova:
Hi @tai271828, thanks for your contribution! I'm afraid I don't really understand what the new message is trying to say - could you be a bit more descriptive here so we can work out the best wording? :)

tai271828:
@lilyvidenova  Hi, thank you for the comments. Here is the context:

The power configuration of layout would pop up a message to tell users `This power controller manages N other node. Changing power parameters will affect these nodes.` ( The number N depends on real cases) when using the power device `American Pwer Conversion (APC) PDU`.

Selection_293

The original message `This power controller manages N other node. Changing power parameters will affect these nodes.` may make users think

1. if he/she changes one of the fields (IP, outlet number, or outlet delay), all fields of all outlets and nodes controlled by the same APC PDU will become the same. In fact, only IP and outlet delay will be changed to be the same for all nodes managed by the PDU. The field outlet number is always isolated and specific for each node.
2. this pop-up message is a bug because he/she expects all outlet parameters (IP, outlet number, and outlet delay) should be isolated. Change any field of the outlet of one node should not affect another one node (in fact the IP and outlet delay will change to be the same), so people will file a bug report against MaaS like https://bugs.launchpad.net/maas/+bug/1810796


I hope the additional information will help us to work out the best wording.

lilyvidenova:
@tai271828 this makes sense, thanks! Let's make it clearer by changing the message to: 

'Changing the IP address or outlet delay will affect all these nodes.' 
Thanks! :)

tai271828:
Sounds great. I will update this pull request soon.

tai271828:
Fixed by following the suggestion. Please review.

lilyvidenova:
Awesome, thanks @tai271828 :)

Issue #524, opened 2019-11-29
Loading indicators not aligned


squidsoup:
We have two loading indicators, one of which is offset to the right which looks a bit weird.

![2019-11-29 09 58 48](https://user-images.githubusercontent.com/130286/69861117-935a1880-128f-11ea-9a66-ee4d999c1d37.gif)

PR #557, opened 2019-12-10
Upgrade vanilla 2.5.0

huwshimi:
## Done
- Upgraded Vanilla to 2.5.0.
- Upgrade react-components to 0.3.0.
- Replaced overrides with utilities.
- Fixed the loading position. Fixes: https://github.com/canonical-web-and-design/maas-ui/issues/524

## QA
- Load the react ui, check that the loading spinners are aligned with the header logo.
- View the settings and preferences and check that the icon buttons are unchanged in the lists (e.g. users, api keys).
- Check that all the settings pages now load.

### DIFF
--- a/proxy/index.js
+++ b/proxy/index.js
@@ -42,6 +42,9 @@ app.use(
   })
 );
 
+// Proxy the HMR url to the React client.
+app.use(proxy("/main.*.hot-update.js", { target: "http://localhost:8401/" }));
+
 // Proxy the remaining URLs to the Angular client.
 app.use(
   proxy(`${process.env.BASENAME}/`, {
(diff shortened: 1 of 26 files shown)


proxy/index.js
```diff
@@ -42,6 +42,9 @@ app.use(
   })
 );
 
+// Proxy the HMR url to the React client.
+app.use(proxy("/main.*.hot-update.js", { target: "http://localhost:8401/" }));
```
huwshimi: Drive-by to load the HMR js.

ui/src/app/settings/components/Routes/Routes.js
```diff
@@ -68,16 +66,8 @@ const Routes = () => {
         to={`${match.path}/configuration/general`}
       />
       <Route exact path={`${match.path}/users`} component={UsersList} />
-      {!externalAuthURL && (
```
huwshimi: Drive-by to fix the URLs (this check is now done in the components).

proxy/index.js
squidsoup: Nice

Issue #566, opened 2019-12-11
Add analytics to React UI


huwshimi:
From: https://github.com/canonical-web-and-design/maas-squad/issues/1495#issuecomment-563195858

"Could we add an event on every submit button, please? On both Settings and User preferences. And one on the checkbox in General for when the checkbox to share analytics is checked/unchecked?"

PR #570, opened 2019-12-12
Send analytics events when forms are saved.

huwshimi:
## Done
- Send events when forms are saved in settings and preferences.
- Send an event when the GA checkbox is changed.

## QA
- This one's a bit tricky to QA, you could add something like: `console.log("send event", eventCategory, eventAction, eventLabel);` in analytics.js and then check that you get events when saving forms and changing whether GA is checked in the General settings.

## Fixes
Fixes: https://github.com/canonical-web-and-design/maas-ui/issues/566.


squidsoup:
Works nicely, thanks @huwshimi!

PR #619, opened 2020-01-06
Switch the Usabilla code to the 2.7 campaign

anthonydillon:
## Done
* Switch the Usabilla code to the 2.7 campaign

## QA
* Load the UI
* Check that the Feedback tab appears on the right side
* Click and check it looks like this one:
https://codepen.io/anthonydillon/pen/PowEeeE

## Fixes
Fixes https://github.com/canonical-web-and-design/MAAS-squad/issues/1714


huwshimi:
@anthonydillon did you mean to say that the feedback button should or shouldn't appear on the right side? I can't see it (it's being hidden by our code: https://github.com/canonical-web-and-design/maas-ui/blob/master/shared/src/components/Footer/Footer.js#L9).

Caleb-Ellis:
@anthonydillon is there a way to detect if a user's ad blocker prevents the usabilla script from loading? The "Give feedback" link in the footer does nothing when clicked - maybe we can remove it if usabilla doesn't load.

Also as @huwshimi said the tab isn't showing atm.

anthonydillon:
@huwshimi it's hidden and has a set of rules set in Usabillia to display to. For example: 5% of users, period of time on the page.

lilyvidenova:
Currently it shows to returning users, who have seen at least 2 pages and have spent more than 120 seconds on a page (I’ll disable it for about a month to give people a chance to use 2.7 first)

PR #631, opened 2020-01-08
Add power column to new machine listing.

squidsoup:
## Done
Implement power column in machine listing.

## QA
Ensure power type, and state display correctly with the appropriate icon.

## Notes
Would really appreciate some thoughts on the css here - it remains unchanged from the implementation in the angular app, but perhaps this is a good opportunity to make improvements.


ui/src/app/machines/views/MachineList/PowerColumn/PowerColumn.test.js
```diff
@@ -0,0 +1,96 @@
...
+        >
+          <PowerColumn systemId="abc123" />
+        </MemoryRouter>
+      </Provider>
+    );
+
+    expect(wrapper.find('[data-test="power_state"]').text()).toEqual("off");
+  });
+
+  it("displays the correct power type", () => {
```
anthonydillon: Do you need both tests here? They will both fail it something goes wrong right?

squidsoup: I think a good testing principal is tests should try to isolate a single failure, so should generally just have a single assertion.

ui/src/app/machines/views/MachineList/PowerColumn/PowerColumn.js
```diff
@@ -0,0 +1,49 @@
...
+  );
+
+  const iconClass = classNames({
+    "p-icon--power-on": machine.power_state === "on",
+    "p-icon--power-off": machine.power_state === "off",
+    "p-icon--power-error": machine.power_state === "error",
+    "p-icon--power-unknown": machine.power_state === "unknown"
+  });
+
+  if (machine.powerTransition) {
```
Caleb-Ellis: `machine.powerTransition` is actually not part of the machine object returned from the websocket - it was chucked on as part of the angular code. So for now this won't do anything.

ui/src/scss/_patterns_icons.scss
```diff
@@ -0,0 +1,49 @@
+.p-icon--chevron {
```
Caleb-Ellis: Vanilla changed it to `p-icon--contextual-menu` so you can use that and remove this class

ui/src/scss/_patterns_double-row.scss
```diff
@@ -0,0 +1,53 @@
+@mixin maas-double-row {
+  $checkbox-space: 1rem + $sph-inner;
+  $icon-space: map-get($icon-sizes, default) + $sph-inner--small;
+
+  .p-double-row {
+    overflow: visible;
+
+    %p-double-row__icon-container {
+      display: block;
+      float: left;
```
Caleb-Ellis: We should definitely move away from using floats. They end up requiring a lot of workarounds to get things to line up, so I would suggest using a simple grid for the double row instead. I've quickly written an example here: https://pastebin.canonical.com/p/zdDZyz4WmC/

Note this only makes sense for columns that have an icon or checkbox - the other ones could probably just have a flex column.

ui/src/app/machines/views/MachineList/PowerColumn/PowerColumn.js
squidsoup: Ah thanks @Caleb-Ellis quite right, it is indeed only in response to an action. I think we'd probably handle this differently in redux, rather than tacking on a property to the machine model, so I'm just going to remove this for now until we implement actions.

ui/src/scss/_patterns_double-row.scss
squidsoup: Excellent, thank you @Caleb-Ellis - that's much simpler and clearer!

ui/src/app/machines/views/MachineList/MachineList.js
```diff
@@ -134,8 +135,8 @@ const generateRows = (rows, hiddenGroups, setHiddenGroups) =>
           content: <NameColumn showMAC={false} systemId={row.system_id} />
         },
         {
-          content: row.power_state,
-          className: "u-upper-case--first"
+          content: <PowerColumn systemId={row.system_id} />,
+          className: "p-table__col--power p-double-row"
```
Caleb-Ellis: I don't think we need `p-double-row` here - that would just be for the cells that don't have an icon/checkbox in them. Also `p-table__col--power` isn't doing anything yet, but in the original angular css it was responsible for the widths and breakpoints for the power column. Probs fine to just leave it for now

ui/src/scss/_patterns_double-row.scss
```diff
@@ -0,0 +1,38 @@
+@mixin maas-double-row {
+  $icon-space: map-get($icon-sizes, default) + $sph-inner--small;
+
+  .p-double-row {
+    display: flex;
+    flex-direction: column;
+  }
+
+  .p-double-row {
```
Caleb-Ellis: I think we only need one of these classes :P

ui/src/scss/_patterns_icons.scss
```diff
@@ -0,0 +1,44 @@
...
+}
+
+.p-icon--locked {
+  @extend %icon;
+  background-image: url("data:image/svg+xml,%3Csvg width='16px' height='16px' viewBox='0 0 16 16' version='1.1' xmlns='http://www.w3.org/2000/svg' xmlns:xlink='http://www.w3.org/1999/xlink'%3E%3Cdefs%3E%3Cpath d='M8,3e-05 C5.7926,3e-05 4,1.79263 4,4.00003 L4,7.00003 L2,7.00003 L2,15.99953 L14,15.99953 L14,7.00003 L12,7.00003 L12,4.00003 C12,1.79263 10.207,3e-05 8,3e-05 L8,0 L8,3e-05 Z M8,1.00003 C9.6706,1.00003 11,2.32933 11,4.00003 L11,7.00003 L5,7.00003 L5,4.00003 C5,2.32933 6.3293,1.00003 8,1.00003 Z M9,9.50003 L9,13.99953 L7,13.99953 L7,9.99953 L9,9.50003 L9,9.50003 Z' id='padlock-icon'%3E%3C/path%3E%3C/defs%3E%3Cg id='padlock-16'%3E%3Cuse id='lock-icon' fill='"+vf-url-friendly-color(
+      $color-mid-dark
+    )+"' fill-rule='nonzero' xlink:href='%23padlock-icon'%3E%3C/use%3E%3C/g%3E%3C/svg%3E%0A");
+}
+
+@mixin maas-icon-power-on($color) {
```
Caleb-Ellis: This should probably just be called `maas-icon-power`.

PR #639, opened 2020-01-10
LP: #1859032 - Fix bug where newly composed machine would not appear in KVM page.

Caleb-Ellis:
## Done
* Fixed bug where newly composed machine would not appear in KVM page. I'm not proud of the solution (at all) but I'm not terribly excited about rewriting big chunks of the machine list or pod details controllers. :(
* Improved action feedback of compose machine button

## QA
* Compose a machine in a KVM and check that the button becomes disabled with a spinner
* Check that the machine shows up automatically in the list below the pod summary.

## Fixes
Fixes canonical-web-and-design/MAAS-squad#1079

## Launchpad Issue
[lp#1859032](https://bugs.launchpad.net/maas/+bug/1859032)

## Screenshot
0 0 0 0_8400_MAAS_ (5)

### DIFF
--- a/legacy/src/app/controllers/pod_details.js
+++ b/legacy/src/app/controllers/pod_details.js
@@ -650,6 +650,18 @@ function PodDetailsController(
         }
       }
     });
+    $scope.$watch("pod.composed_machines_count", () => {
+      // Explicitly refresh machine list by changing search to return identical
+      // results when the number of machines in the pod changes. Symptom of
+      // problems with two-way data binding of machines and search string in the
+      // machine table.
+      if ($scope.loaded) {
+        $scope.machinesSearch = "pod:=" + $scope.pod.name;
+        setTimeout(() => {
+          $scope.machinesSearch = "pod-id:=" + $scope.pod.id;
+        }, 100);
+      }
+    });
     $scope.$watch("action.option", function(now, then) {
       // When the compose action is selected set the default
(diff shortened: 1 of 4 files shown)


squidsoup:
@Caleb-Ellis works just fine, although the newly composed machine can sit beneath the "show all machines" fold, requiring an additional click, which is a bit unfortunate. Could we potentially change this to show all machines in this view?

Caleb-Ellis:
@squidsoup that's a good idea - while QAing I would just automatically show all before composing a pod. I'll add that in now

Caleb-Ellis:
@squidsoup upped the display limit to 100 for the pod list instead of removing it entirely. Juuuuust in case someone has a ton of machines and the KVM details page slows to a crawl.

Caleb-Ellis:
@squidsoup Added another "prop" `defaultSort` to the machine list directive, so now the pod machine list sorts by status by default

squidsoup:
> @squidsoup Added another "prop" `defaultSort` to the machine list directive, so now the pod machine list sorts by status by default. Took me an embarrassingly long time to figure out that to pass a string to the directive it needs to be:
> 
> ```
> <directive prop="'string'"></directive>
> ```
> 
> =[

Awesome, that's great thank you @Caleb-Ellis. I'm sorry angularjs broke your brain :(

PR #641, opened 2020-01-10
Set up real MAAS when running cypress in Circle CI.

### LINKED ISSUE #541: Set up Cypress: Write script for setting up MAAS in CI
1. Use a Docker image that uses the latest Ubuntu, Yarn and Node LTSs, and install MAAS and Cypress at run-time. Alternatively we could create a custom Docker image that includes the requirements for MAAS and has Cypress pre-installed.

2. Install the latest MAAS snap in devmode: `sudo snap install maas --channel=2.7/edge --devmode`

3. Initialise MAAS with first user: `sudo maas init --mode all`, note that name, password **and** email are required. URL should also be set manually.

4. Verify MAAS is running: `sudo maas config`, check that mode is `all` and url has been set correctly.

5. Copy `proxy/env` to `proxy/env.local` and replace MAAS_URL with the URL set above.

6. Run maas-ui and cypress: `yarn test-cypress`

Caleb-Ellis:
## Done
- Set up basic, real MAAS using 2.7/edge snap when running cypress in Circle CI.
- Added a non-stubbed test that could only pass if a real MAAS was running (auth errors returned from api and displayed in UI)
- Removed Circle CI tests that have been migrated to GH actions

## QA
- Check that tests pass in Circle CI

Fixes #541

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -6,23 +6,48 @@ defaults: &defaults
 version: 2
 jobs:
   test-cypress:
-    docker:
-      - image: cypress/base:12.16.1
-        environment:
-          ## this enables colors in the output
-          TERM: xterm
+    machine:
+      image: ubuntu-1604:201903-01
     steps:
       - checkout
-      # Download and cache dependencies
-      - restore_cache:
-          keys:
-            - yarn-{{ checksum "yarn.lock" }}
-      - run: yarn install --frozen-lockfile
-      - save_cache:
-          key: yarn-{{ checksum "yarn.lock" }}
-          paths:
-            - ~/.cache ## cache both yarn and Cypress!
-      - run: yarn test-cypress
+      - run:
+          name: Set environment variables
+          command: |
+            echo 'export IP_ADDRESS=$(hostname -I | cut -d" " -f1)' >> $BASH_ENV
+            echo 'export NVM_DIR="/opt/circleci/.nvm"' >> $BASH_ENV
+            echo ' [ -s "$NVM_DIR/nvm.sh" ] && \. "$NVM_DIR/nvm.sh"' >> $BASH_ENV
+            source $BASH_ENV
+      - run:
+          name: Install system dependencies
+          command: |
+            sudo apt-get update && sudo apt-get install -y curl
+            curl -sS https://dl.yarnpkg.com/debian/pubkey.gpg | sudo apt-key add -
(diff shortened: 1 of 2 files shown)


.circleci/config.yml
```diff
@@ -6,23 +6,48 @@ defaults: &defaults
 version: 2
 jobs:
   test-cypress:
-    docker:
-      - image: cypress/base:12.16.1
-        environment:
-          ## this enables colors in the output
-          TERM: xterm
+    machine:
+      image: ubuntu-1604:201903-01
```
sparkiegeek: That's old...

.circleci/config.yml
```diff
@@ -6,23 +6,48 @@ defaults: &defaults
...
+            node -v
+            nvm alias default v12.14.1
+            yarn install
+            ./node_modules/.bin/cypress verify
+      - run:
+          name: Install and initialise edge MAAS snap
+          command: |
+            echo $IP_ADDRESS
+            sudo systemctl enable snapd
+            sudo snap install maas --channel=2.7/edge
```
sparkiegeek: why `2.7/edge`? master here will land in `latest/edge` and `2.8/edge`

.circleci/config.yml
huwshimi: @sparkiegeek this looks like the most recent version available on CircleCI (https://circleci.com/docs/2.0/configuration-reference/#available-machine-images). Is that going to be an issue?

.circleci/config.yml
squidsoup: @Caleb-Ellis I think once this point is addressed this should be good to land (given we can't do anything about the images provided by Circle).

@sparkiegeek GH provides an 18.04 image, but due to an unfortunate limitation of Github Actions, we can't run CI against PRs that require secrets (the Cypress key in this case) currently. Once we can migrate, we'll move to the the latest LTS.

.circleci/config.yml
squidsoup: Unfortunately the MAAS db initialisation script requires 18.04, and as we can't run the snap in a docker image, we're not left with many options. @sparkiegeek @albertodonato I suspect the issue is that we need the version of postgresql in 18.04? Would adding a PPA with a backport for 16.04 work?

albertodonato: Yes, it should. MAAS currently requires PostgreSQL 10 at least, so you could probably add the appropriate repo from https://wiki.postgresql.org/wiki/Apt

PR #648, opened 2020-01-13
Add disks column to new machine listing.

Caleb-Ellis:
## Done
- Added disks column which shows `machine.physical_disk_count`.
- Copied over scriptStatus enum from legacy which just lists out the different statuses a script can be in (failed, passed, running etc).
- Added ScriptStatus component that generates an icon and tooltip given a certain script type's status (e.g. storage tests have failed = return error icon and tooltip "Machine has failed tests", cpu tests are pending = return pending icon without tooltip, etc).
- Added icons for pending, running and timed out tests.
- Small css and markup fixes for other columns.

## QA
- Check that the disks column correctly displays the physical_disk_count of the machine
- Check that icons and tooltips display for machines that have storage tests that have failed/timed out/pending/running/no tests

## Fixes
Fixes canonical-web-and-design/MAAS-squad#1763

## Screenshot
screenshot

### DIFF
--- a/ui/src/app/base/components/ScriptStatus/ScriptStatus.js
+++ b/ui/src/app/base/components/ScriptStatus/ScriptStatus.js
@@ -1,0 +1,108 @@
+import PropTypes from "prop-types";
+import React from "react";
+
+import { scriptStatus } from "app/base/enum";
+import Tooltip from "app/base/components/Tooltip";
+
+const ScriptStatus = ({
+  children,
+  hidePassedIcon = false,
+  scriptType,
+  tooltipPosition
+}) => {
+  switch (scriptType.status) {
+    case scriptStatus.PASSED: {
+      return hidePassedIcon ? (
+        children
+      ) : (
+        <>
+          <i className="p-icon--success is-inline" />
+          {children}
+        </>
+      );
+    }
+
+    case scriptStatus.DEGRADED:
+    case scriptStatus.FAILED:
+    case scriptStatus.FAILED_APPLYING_NETCONF:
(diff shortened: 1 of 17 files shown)


squidsoup:
Machines with a testing icon in "pending" state don't seem to have a tooltip. 

e.g. tooltip in angular, but not react
image

ui/src/app/base/components/ScriptStatus/ScriptStatus.js
```diff
@@ -0,0 +1,108 @@
...
+    case scriptStatus.RUNNING: {
+      return (
+        <>
+          <i className="p-icon--running is-inline" />
+          {children}
+        </>
+      );
+    }
+
+    case scriptStatus.PENDING: {
```
squidsoup: As mentioned, this gets a tooltip currently, although I think the tooltip is possibly wrong in the angular app, and should indicate that this is pending?

Caleb-Ellis: @squidsoup Yeah I'm pretty sure the angular tooltip is wrong because it assumes any non-passing status == failed. I opted to just remove the tooltip entirely if the status is pending or running because I figured the icon was enough, but maybe my UX-fu is off.

@lilyvidenova any thoughts on if tooltips should show for machines with pending/running tests, and if so what should they say?

Caleb-Ellis: @lilyvidenova oh yeah also I added a warning icon and tooltip if a machine hasn't run any tests of that type yet ("Machine has not run any tests of this type.")

PR #687, opened 2020-01-20
Add status column to new machine listing.

Caleb-Ellis:
## Done
- Added StatusColumn component that handles the logic of what shows up in the status cell.

## QA
- Check that a machine that is deployed just shows the OS + release (only the number + LTS if an Ubuntu release e.g. Ubuntu 18.04 LTS instead of Ubuntu 18.04 LTS Bionic Beaver)
- Check that a machine that is deploying shows `Deploying ${OS + release}`
- Check that if a machine has failed tests and is not new, commissioning or testing, a warning icon and tooltip shows
- Check that when a machine is in a transient state (e.g. deploying), progress text is shown below the status and a spinner shows

## Fixes
Fixes canonical-web-and-design/MAAS-squad#1756

### DIFF
--- a/ui/src/app/machines/views/MachineList/StatusColumn/StatusColumn.js
+++ b/ui/src/app/machines/views/MachineList/StatusColumn/StatusColumn.js
@@ -1,0 +1,131 @@
+import { useSelector } from "react-redux";
+import React from "react";
+import PropTypes from "prop-types";
+
+import {
+  general as generalSelectors,
+  machine as machineSelectors
+} from "app/base/selectors";
+import { nodeStatus, scriptStatus } from "app/base/enum";
+import Tooltip from "app/base/components/Tooltip";
+
+// Node statuses for which the failed test warning is not shown.
+const hideFailedTestWarningStatuses = [
+  nodeStatus.COMMISSIONING,
+  nodeStatus.FAILED_COMMISSIONING,
+  nodeStatus.FAILED_TESTING,
+  nodeStatus.NEW,
+  nodeStatus.TESTING
+];
+
+// Node statuses for which the OS + release is made human-readable.
+const formattedReleaseStatuses = [nodeStatus.DEPLOYED, nodeStatus.DEPLOYING];
+
+// Node statuses that are temporary.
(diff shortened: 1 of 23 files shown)


ui/src/app/machines/views/MachineList/StatusColumn/StatusColumn.js
```diff
@@ -0,0 +1,131 @@
...
+  return "";
+};
+
+const getStatusIcon = machine => {
+  if (transientStatuses.includes(machine.status_code)) {
+    return (
+      <i
+        className="p-icon--spinner u-animation--spin"
+        data-test="status-icon"
+      />
```
huwshimi: Could we use the Loading component for this?

ui/src/app/machines/views/MachineList/StatusColumn/StatusColumn.js
```diff
@@ -0,0 +1,131 @@
...
+  );
+  const osReleases = useSelector(state =>
+    generalSelectors.osInfo.getOsReleases(state, machine.osystem)
+  );
+
+  return (
+    <div className="p-double-row--with-icon">
+      <div className="p-double-row__icon">{getStatusIcon(machine)}</div>
+      <div
+        className="p-double-row__primary-row u-truncate-text"
```
huwshimi: Could we just use `u-truncate` for these?

ui/src/app/machines/views/MachineList/StatusColumn/StatusColumn.js
Caleb-Ellis: @huwshimi Updated. I've had to add "u-no-padding u-no-margin" to the Loader otherwise it popped out of its container - maybe we should add an issue in react-components?

ui/src/app/machines/views/MachineList/StatusColumn/StatusColumn.js
Caleb-Ellis: @huwshimi I had no idea that existed! `u-truncate-text` is in a few places so I'll update them all.

PR #688, opened 2020-01-20
Add storage column to new machine listing.

Caleb-Ellis:
## Done
- Added StorageColumn component, which just shows the total storage of the machine.
- Replace `formatGigabytes` function with `formatBytes` which accepts the unit as the supplied argument, so it should be able to handle any case.

## QA
- Go the new machine list and see that the storage cells look like the old machine list.

## Fixes
Fixes canonical-web-and-design/MAAS-squad#1764


squidsoup:
@Caleb-Ellis bunch of conflicts here sorry.

Caleb-Ellis:
@squidsoup should be all fixed now, and this should be the last of the raw data done! Now all that's left is the one million different actions you can do on the page.

huwshimi:
@Caleb-Ellis in the old machine list it shows two decimal places while this shows three, is that OK do you think?

Caleb-Ellis:
@huwshimi Oh oops I didn't realise, I thought it was 3 decimal places there as well. I don't think it matters but I'll change it to 2 and maybe we decide the UX of it later.

PR #699, opened 2020-01-24
Add double row component

huwshimi:
## Done
- Replace double row elements with a component.
- Use the double row component to display a contextual menu.

## QA
- Visit /r/machines.
- Hover rows, the menu icon should appear next to the owner.
- Clicking the menu icon should display the menu and keep the row highlighted.

## Fixes
Fixes https://github.com/canonical-web-and-design/MAAS-squad/issues/1797.


huwshimi:
@Caleb-Ellis Thanks for the review. I think I've addressed all those issues. Would you mind taking another look?

PR #700, opened 2020-01-24
Implement machine list grouping and sorting with double row headers

Caleb-Ellis:
## Done
- Added a TableHeader component which allows table headers to be clickable to run their own sort functions. Used in MachineList and UserList in settings. This required a bit of custom work to override the default MainTable sort behaviour, so I'm not sure this work should be upstreamed to the component library (if ever). I think once the UX of a generic table between products is finalised we can settle on an implementation, but for now this implements what's in the current angular version.
- Added a GroupSelect component that implements grouping as in the current angular version, plus added options for pool, power state and zone.
- Added placeholder filter dropdown and searchbox.
- Refactored MachineList to better separate grouping, sorting and row generation functions.
- Clicking the FQDN | MAC headers now change the first cell content.
- Drive by styling fixes:
  - Dense table headers (should be released in Vanilla soon)
  - Table cell/header spacing
  - Ensure table widths add up to 100% (oops)

## QA
- Check that you can change the leftmost column to MAC or FQDN.
- Check that clicking the table headers changes the sort parameter, and can be cycled between ascending, descending and none. Check that sorting works for all column headers, including FQDN and MAC.
- Check that you can change the grouping of machines, and that changing the sort changes the order of machines within each group (not the groups themselves)
- Check that the Settings > Users table still sorts as expected.

## Fixes
Fixes canonical-web-and-design/MAAS-squad#1744
Fixes canonical-web-and-design/MAAS-squad#1745
Fixes canonical-web-and-design/MAAS-squad#1746

## Screenshot
0 0 0 0_8400_MAAS_r_machines (1)

### DIFF
--- a/ui/src/app/base/components/TableHeader/TableHeader.js
+++ b/ui/src/app/base/components/TableHeader/TableHeader.js
@@ -1,0 +1,42 @@
+import { Button } from "@canonical/react-components";
+import classNames from "classnames";
+import PropTypes from "prop-types";
+import React from "react";
+
+const TableHeader = ({
+  children,
+  className,
+  currentSort,
+  onClick,
+  sortKey
+}) => {
+  if (!onClick) {
+    return <div className={className}>{children}</div>;
+  }
+
+  return (
+    <Button appearance="link" className={className} onClick={onClick}>
+      <span>{children}</span>
+      {currentSort && currentSort.key === sortKey && (
+        <i
+          className={classNames("p-icon--contextual-menu", {
+            "u-mirror--y": currentSort.direction === "ascending"
+          })}
+        />
+      )}
+    </Button>
+  );
+};
+
+TableHeader.propTypes = {
+  children: PropTypes.node.isRequired,
+  className: PropTypes.string,
+  currentSort: PropTypes.shape({
+    key: PropTypes.string,
+    direction: PropTypes.oneOf(["ascending", "descending", "none"])
+  }),
+  onClick: PropTypes.func,
+  sortKey: PropTypes.string
+};
+
+export default TableHeader;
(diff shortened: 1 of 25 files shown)


huwshimi:
@Caleb-Ellis this seems to have similarities to `p-table-multi-header` found here: https://github.com/canonical-web-and-design/maas-ui/blob/master/ui/src/scss/_patterns_table.scss#L22. Do you think they could be merged somehow or are the cases too different?

Caleb-Ellis:
@huwshimi ah I forgot about that. I think they should be able to be merged no problem.

huwshimi:
@Caleb-Ellis sometimes when clicking on the owner or storage header the header greys out and stops being clickable after the first click.

Caleb-Ellis:
@huwshimi weird, it was only happening for the Storage header for me. Why not all of them? D: Anyway I think the issue was wrapping a Button in a span, so I've removed that and they all seem to work fine for me now.

squidsoup:
@huwshimi since you've had eyes on this, probably best for you to review.

huwshimi:
@Caleb-Ellis one last QA thing, when sorting it doesn't move the groups around (as you say in your QA notes), but when clicking on the header for column that the table is currently grouped by it does nothing. Do you think it should then move the groups around or disable clicking on that header? E.g. if grouped by zone and you click on the zone header to sort the table should it could change the order of the groups or not allow clicking on that header.

Caleb-Ellis:
@lilyvidenova any thoughts on how sorting groups should work, in regards to clicking its header?

lilyvidenova:
@Caleb-Ellis @huwshimi let's keep it as it is, if it is grouped and you click on the header that is already applied as grouping, nothing happens - we will discuss the behaviour in more depth at the react component level with the rest of the UXers and might end up changing the behaviour, however for now I'd like to keep it as is.

Issue #714, opened 2020-01-29
Table menu sometimes appears in the top-left of the page


Caleb-Ellis:
If I click just below the middle of the menu toggle, it appears in the top-left of the page.

![screenshot_9](https://user-images.githubusercontent.com/25733845/73339763-b5fdb300-4279-11ea-99ee-d27d4a747066.png)

PR #717, opened 2020-01-30
Fix contextual menu position

huwshimi:
## Done
- Only calculate styles when the toggle element is visible.

## QA
- Visit /r/machines.
- Click on a contextual menu a bunch of times, it should always appear in the same place (never in the top left of the page).

## Fixes
Fixes https://github.com/canonical-web-and-design/maas-ui/issues/714.

### DIFF
--- a/ui/src/app/base/components/ContextualMenu/ContextualMenu.js
+++ b/ui/src/app/base/components/ContextualMenu/ContextualMenu.js
@@ -57,11 +57,11 @@ const ContextualMenu = ({
 }) => {
   const id = useRef(nanoid());
   const wrapper = useRef(null);
+  const positionStyle = useRef(null);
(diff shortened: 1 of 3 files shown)


ui/src/app/base/components/ContextualMenu/ContextualMenu.js
```diff
@@ -76,6 +76,18 @@ const ContextualMenu = ({
     // redefined on a state update which causes an infinite loop here.
   }, [isOpen]); // eslint-disable-line react-hooks/exhaustive-deps
 
+  useEffect(() => {
```
squidsoup: This could possibly use a comment explaining why it's necessary.

huwshimi: Done, thanks.

PR #737, opened 2020-02-03
Add checkboxes for individual machines.

Caleb-Ellis:
## Done
- Added checkboxes for individual machines in machine list. Selected state is handled by the MachineList component.

## QA
- Log in as an admin.
- Go to the machine list and check there are checkboxes that you can select and unselect.
- Log in as a user.
- Go to the machine list and check that the checkboxes are disabled with a tooltip.

## Fixes
Fixes canonical-web-and-design/MAAS-squad#1741

### DIFF
--- a/ui/src/app/base/components/DoubleRow/DoubleRow.js
+++ b/ui/src/app/base/components/DoubleRow/DoubleRow.js
@@ -15,6 +15,7 @@ const DoubleRow = ({
   primary,
   primaryAriaLabel,
   primaryClassName,
+  primaryTextClassName,
   secondary,
   secondaryAriaLabel,
   secondaryClassName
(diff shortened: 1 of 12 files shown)


huwshimi:
@Caleb-Ellis This is looking great, but the secondary row does not align correctly when there is a checkbox (see below). I'm not sure how much this will change things so I'll do a code review after.

Screen Shot 2020-02-04 at 10 35 58 am

Caleb-Ellis:
@huwshimi You were right to hold off on code reviewing! Anyway the issue was originally I added a little offset to the checkbox. Otherwise the little blue focus ring was cut-off on the left side because of overflow: hidden on the parent. So I've rearranged some scss and added some classes that deal with this case without affecting the contextual menu position.

PR #738, opened 2020-02-04
List pools

squidsoup:
## Done
Add list of pools.

## QA
Visit /pools. A table should display pools correctly, with action buttons in the appropriate state (disabled without permissions, only edit for the default pool).

image

### DIFF
--- a/ui/src/app/pools/views/Pools.js
+++ b/ui/src/app/pools/views/Pools.js
@@ -113,7 +113,7 @@ const Pools = () => {
                 },
                 {
                   content: "Machines",
-                  sortKey: "machine_total_count"
+                  sortKey: "machines"
                 },
                 {
                   content: "Description",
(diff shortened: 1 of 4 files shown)


lilyvidenova:
@squidsoup can we always show the delete action, even if unavailable, similarly to how we do it in users:

image

with a tooltip on hover saying "The default pool cannot be deleted."  (I believe this is the only pool that's undeletable)

Caleb-Ellis:
I'll let @huwshimi review this one after the delete button thing is implemented :)

ui/src/app/pools/views/Pools.js
```diff
@@ -1,5 +1,138 @@
...
+              className="p-table-expanding--light"
+              defaultSortDirection="ascending"
+              headers={[
+                {
+                  content: "Name",
+                  sortKey: "name"
+                },
+                {
+                  content: "Machines",
+                  sortKey: "machine_total_count"
```
huwshimi: I think this key should be `machines`

squidsoup: Good catch, thanks @huwshimi!

PR #741, opened 2020-02-04
Add checkboxes for machine groups.

Caleb-Ellis:
## Done
- Added checkboxes for groups in the machine list

## QA
- Check that clicking a group checkbox selects every machine in that list, unless all of them are selected, in which case it unselects them all

## Fixes
Fixes canonical-web-and-design/MAAS-squad#1740

screenshot_19

### DIFF
--- a/ui/src/app/machines/views/MachineList/MachineList.test.js
+++ b/ui/src/app/machines/views/MachineList/MachineList.test.js
@@ -449,4 +449,122 @@ describe("MachineList", () => {
         .props().checked
     ).toBe(true);
   });
+
+  it("can select all machines in a group if at least one is unselected", () => {
+    const state = { ...initialState };
+    state.machine.items[1].status_code = nodeStatus.DEPLOYED;
+    const store = mockStore(state);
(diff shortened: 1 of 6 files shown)


ui/src/app/machines/views/MachineList/MachineList.test.js
```diff
@@ -449,4 +449,88 @@ describe("MachineList", () => {
...
+      .find("[data-test='group-cell'] input")
+      .at(0)
+      .simulate("change");
+    expect(
+      wrapper.find("[data-test='name-column'] input[checked=true]").length
+    ).toBe(0);
+    expect(
+      wrapper.find("[data-test='group-cell'] input[checked=true]").length
+    ).toBe(0);
+  });
```
huwshimi: Could possibly add a test here that the group checkbox is unchecked if you check all then uncheck one.

Caleb-Ellis: Thanks @huwshimi, added another test

PR #749, opened 2020-02-05
Clear Redux state on log out.

### LINKED ISSUE #629: Logging out from the React side doesn't clear the header
After logging out I see this:

Screenshot_2020-01-08 Login karura MAAS

Clicking an angular link refreshes the page to an empty login page. Clicking a react link (Settings, username, Log out) does nothing.

Caleb-Ellis:
## Done
- Set initial state for each reducer on LOGOUT_SUCCESS. Exceptions are auth and status which were causing authentication issues when cleared.

## QA
- Log in as an admin and go to the machine list.
- Log out and check that the header is cleared.
- Log in as a user while the url is still at the machine list.
- Check that the list of machines changes.
- Log out and check Redux Devtools that on LOGOUT_SUCCESS the state for everything reset.

## Fixes
Fixes #629

### DIFF
--- a/ui/src/__snapshots__/root-reducer.test.js.snap
+++ b/ui/src/__snapshots__/root-reducer.test.js.snap
@@ -1,0 +1,202 @@
+// Jest Snapshot v1, https://goo.gl/fbAQLP
+
+exports[`rootReducer should reset app to initial state on LOGOUT_SUCCESS, except status which
+    resets to authenticating = false 1`] = `
+Object {
+  "config": Object {
+    "items": Array [],
+    "loaded": false,
+    "loading": false,
+    "saved": false,
+    "saving": false,
+  },
+  "controller": Object {
+    "items": Array [],
+    "loaded": false,
+    "loading": false,
+  },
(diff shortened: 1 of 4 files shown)


Caleb-Ellis:
@huwshimi yeah I agree about potentially forgetting to add LOGOUT_SUCCESS cases for new reducers. I originally tried to implement this at the root reducer level but had some issues with logging back in due to weirdness with the status, auth and router reducers.

After looking around online a bit, it seems there's a split in the Redux community about how to deal with actions that affect multiple reducers - either having a top-level "reduced" reducer for exception cases, or like what I've done where every reducer responds individually and is *entirely* responsible for its portion of state.

I'm happy to change, the only thing I would say is that having everything separated makes it a bit easier to reason about what causes state changes, although I can't argue the pragmatism of only implementing a LOGOUT_SUCCESS handler one time.

Caleb-Ellis:
@huwshimi ignore everything I wrote above. A single exception for resetting state is probably totally okay and it's like one thousandth the amount of code.

PR #764, opened 2020-02-09
Add pool form.

squidsoup:
## Done
* Add form for creating new resource pools

## QA
* Visit /pools/add
* Add a pool, it should save with an appropriate success message.


ui/src/app/pools/views/PoolAdd/PoolAdd.js
```diff
@@ -0,0 +1,9 @@
+import React from "react";
+
+import PoolForm from "../PoolForm";
+
+export const PoolAdd = () => {
+  return <PoolForm />;
+};
+
```
squidsoup: This component is a bit anaemic, but will make more sense once PoolEdit is added in a followup.

PR #772, opened 2020-02-11
Add power menu to machine list

huwshimi:
## Done
- Add the power menu to the machines in the machine list

## QA
- Visit /r/machines.
- In the power column you should see different actions depending on the current power state of the machine.

## Fixes
Fixes https://github.com/canonical-web-and-design/maas-squad/issues/1723.

### DIFF
--- a/ui/src/app/base/actions/machine/machine.js
+++ b/ui/src/app/base/actions/machine/machine.js
@@ -53,6 +53,53 @@ machine.setZone = (systemId, zoneId) => {
   };
 };
 
+machine.turnOff = systemId => {
+  return {
+    type: "TURN_MACHINE_OFF",
+    meta: {
+      model: "machine",
+      method: "action"
+    },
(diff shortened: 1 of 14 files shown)


huwshimi:
Thanks @Caleb-Ellis, these should all be addressed except I've removed the spinner for checking power as we don't have a reliable way of knowing when the action is complete. Sound OK? Would you mind taking another look?

huwshimi:
@Caleb-Ellis I've added this issue about the power checking: https://github.com/canonical-web-and-design/maas-ui/issues/781.

Caleb-Ellis:
@huwshimi nevermind I see what the problem is now - there's no link between the update notify and the original check power request. In the angular version the result from the check power action mutates machine.power_state directly, which we probably shouldn't replicate hahaha.

huwshimi:
Thanks @Caleb-Ellis

PR #778, opened 2020-02-11
Improve perceived performance of machine list.

### LINKED ISSUE #777: Improve perceived performance of machine list.
Navigating to the machine list while everything is loaded into state has some noticeable lag while rendering to the DOM. Nowhere near as much as the angular version, but it's still present. We should improve the perceived performance by rendering something small at first before rendering the table rows.

Caleb-Ellis:
## Done
- Added useComponentDidMount hook which runs a function exactly once after a component is initially rendered.
- Used useComponentDidMount hook to show a spinner in the machine list on first render. I think this makes the app feel faster because there's no delay between when you click and when the screen changes.
- Moved spinner to below the table headers to minimise "jumpiness".
- Fixed the logic for when to show a loading spinner. Originally it was shown while machines were loading, but that would mean nothing else would show while loading a lot of machines (e.g. >1000). It's now only shown if machines have not all loaded.

## QA
- Go to the machine list and click between the pools and machines tabs.
- Check that a loading spinner shows up every time you navigate to the machines tab, and there is no lag between pages switching.

## Fixes
Fixes #777

### DIFF
--- a/ui/src/app/base/hooks.js
+++ b/ui/src/app/base/hooks.js
@@ -2,7 +2,7 @@ import { __RouterContext as RouterContext } from "react-router";
 import { notificationTypes } from "@canonical/react-components";
 import { useContext } from "react";
 import { useDispatch, useSelector } from "react-redux";
-import { useEffect, useRef } from "react";
+import { useEffect, useRef, useState } from "react";
 import { useFormikContext } from "formik";
 
 import { sendAnalyticsEvent } from "analytics";
@@ -155,4 +155,16 @@ export const useSendAnalytics = (
       sendAnalyticsEvent(eventCategory, eventAction, eventLabel);
     }
   }, [sendCondition, eventCategory, eventAction, eventLabel]);
+};
+
+/**
+ * Returns whether the component has undergone initial render,
+ * @returns {Boolean} rendered - Component has rendered.
+ */
+export const useHasRendered = () => {
+  const [rendered, setRendered] = useState(false);
+  useEffect(() => {
+    setRendered(true);
+  }, []);
+  return rendered;
 };
--- a/ui/src/app/machines/views/MachineList/MachineList.js
+++ b/ui/src/app/machines/views/MachineList/MachineList.js
@@ -27,7 +27,7 @@ import {
 import { groupAsMap, simpleSortByKey } from "app/utils";
 import { machine as machineSelectors } from "app/base/selectors";
 import { nodeStatus } from "app/base/enum";
-import { useWindowTitle } from "app/base/hooks";
+import { useHasRendered, useWindowTitle } from "app/base/hooks";
 import CoresColumn from "./CoresColumn";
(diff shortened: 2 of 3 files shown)


ui/src/app/base/hooks.js
```diff
@@ -156,3 +156,10 @@ export const useSendAnalytics = (
     }
   }, [sendCondition, eventCategory, eventAction, eventLabel]);
 };
+
+/**
+ * Runs a function exactly once after component has undergone initial render,
+ * i.e. the component has mounted.
+ * @param {Function} func - Function to run after component mounts.
+ */
+export const useComponentDidMount = func => useEffect(func, []);
```
squidsoup: Can't say I like this very much, componentDidMount is an old lifecycle event that doesn't have much meaning in the context of hooks.

useEffect(func, []) is the component is the idiomatic way to do this, I don't think we need this hook.

ui/src/app/machines/views/MachineList/MachineList.js
```diff
@@ -516,6 +519,23 @@ const MachineList = () => {
     showMAC
   };
 
+  let rows;
+  if (!rendered) {
```
squidsoup: Did you try setting default rows to render if `!machinesLoaded`?

ui/src/app/base/hooks.js
Caleb-Ellis: @squidsoup that's fair! Originally I had it as `useHasRendered` which returned a boolean of whether or not the component has undergone initial render, but I think I generalised it a bit too far in the end.

ui/src/app/machines/views/MachineList/MachineList.js
Caleb-Ellis: >  We should be responding to state changes, not whether react has rendered yet if possible.

I agree with that in principle, but if we take this example to a logical extreme, say the browser has to render 1000 table rows on initial render and it takes 3s while the screen appears frozen. Wouldn't that warrant a *little* skirt around best practices for a better user experience?

> If we render the table with some default data, and remove the machinesLoaded guard on l553, that should achieve what we want here I think? 

I don't think so, because once those machines have loaded into state the browser would still render all the machines at once every time you navigate to the machine list. The central issue is the amount of time the browser takes to repaint the screen, which I was hoping to address with the "rendered" check.

I will say that this has only been an issue I've noticed in development, and could totally be a non-issue in production.

squidsoup: > I will say that this has only been an issue I've noticed in development, and could totally be a non-issue in production.

I think we should definitely confirm that the issue exists in production first @Caleb-Ellis when you have some time - you should be able to just use the test factories to create a bunch of machines on a local maas.

Caleb-Ellis:
Moving to backlog for now - will investigate if slowness happens in production at a later time.

huwshimi:
I'm going to close this for now and we can reopen in the future if necessary.

PR #782, opened 2020-02-12
Initial redux-toolkit spike, refactoring actions.

squidsoup:
## Done
* Refactored common CRUD action creators to use a new function, `createStandardActions`. This makes use of `createAction` from the [redux-toolkit](https://redux-toolkit.js.org/), and will be followed up with further refactoring of reducers using `createReducer`, which further reduce some boilerplate.
* Use reselect's `useSelector` provided and re-exported by redux-toolkit, and remove explicit `reselect` dependency.

## QA
* n/a tests should pass


ui/src/app/base/actions/dhcpsnippet/dhcpsnippet.test.js
```diff
@@ -57,7 +57,7 @@ describe("dhcpsnippet actions", () => {
   });
 
   it("can handle cleaning dhcp snippets", () => {
-    expect(dhcpsnippet.cleanup({ name: "kookaburra" })).toEqual({
```
squidsoup: This test erroneously passed an object to `cleanup`, which takes no arguments.

PR #788, opened 2020-02-12
Replace the run script with the dotrun snap

anthonydillon:
## Done
* Remove the `./run` script
* Document how to run the project via the dotrun snap
* Add `start` NPM script to support the snap 

## QA
* Run `./run` and see it fails
* Follow the README and install the snap
* Join the VPN
* Ensure the config is pointing at karura
* Run `dotrun` from the root of the project
* Check it installs correctly
*  Go to http://localhost:8400/MAAS/r
* Login with the karura login details
* See that its working from within the docker container
* Swich to a local maas and check that works too


hatched:
Using both the `candidate` and `edge` versions of the snap I'm unable to build the project.

```
:dotrun
- Checking yarn dependencies ... changes detected

[ $ yarn install ]

yarn install v1.16.0
[1/4] Resolving packages...
[2/4] Fetching packages...
error An unexpected error occurred: "https://registry.yarnpkg.com/error-ex/-/error-ex-1.3.2.tgz: Request failed \"404 Not Found\"".
```

hatched:
Turns out I needed to do a `./run clean` first before trying the `dotrun` then it built successfully.

anthonydillon:
Why am I assigned to my own PR?

squidsoup:
@anthonydillon is it possible to have the dotrun snap run `yarn clean-all`, as with existing `node_modules` it seems this can be in a weird state where I'm seeing babel errors. Starting from a clean directory works fine.

anthonydillon:
Are you sure you want ever run serve to clean all deps? This surely would slow everyone down.

`dotrun` simply calls a set number of top level  scripts in package.json. Each script common to all projects. Which `clean-all` is not. If you want to run `clean-all` manually you can run `dotrun exec yarn clean-all` or you could add `yarn clean-all` to the `serve` command.

squidsoup:
>  or you could add `yarn clean-all` to the `serve` command.
We definitely do not want to do that.

After a manual install, are you able to confirm that the initial build from dotrun fails?

anthonydillon:
When testing this out I wondered what is the use case switching between the way of running the project and not expecting to clean? If `dotrun` is working well we should all be using that.

squidsoup:
> When testing this out I wondered what is the use case switching between the way of running the project and not expecting to clean? If `dotrun` is working well we should all be using that.

The project is still just a standard js app, I think its reasonable to expect that someone might want to run it without dotrun.

anthonydillon:
@squidsoup done.

Issue #793, opened 2020-02-13
Generalise table menu link creation


Caleb-Ellis:
> I wonder if this whole section of creating the menu links could be generalised somehow? Like a custom hook that you just pass in the allowed actions, something like `const menuLinks = useTableMenuLinks(["acquire", "release"])`. :man_shrugging:

_Originally posted by @Caleb-Ellis in https://github.com/canonical-web-and-design/maas-ui/pull/791_

huwshimi:
This should also include the title from general.machineActions.data as note here: https://github.com/canonical-web-and-design/maas-ui/pull/796#discussion_r379306223

PR #796, opened 2020-02-14
Add the status column actions to the machine list

huwshimi:
## Done
- Add the status column actions to the machine list.

## QA
- Visit /r/machines.
- In the status column you should get a menu of different actions depending on the state of the machine.

## Fixes
Fixes https://github.com/canonical-web-and-design/maas-squad/issues/1722.

### DIFF
--- a/ui/src/app/machines/views/MachineList/StatusColumn/StatusColumn.js
+++ b/ui/src/app/machines/views/MachineList/StatusColumn/StatusColumn.js
@@ -1,8 +1,9 @@
 import { Loader } from "@canonical/react-components";
-import { useSelector } from "react-redux";
+import { useDispatch, useSelector } from "react-redux";
 import React from "react";
 import PropTypes from "prop-types";
 
+import { machine as machineActions } from "app/base/actions";
 import {
   general as generalSelectors,
   machine as machineSelectors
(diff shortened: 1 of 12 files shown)


ui/src/app/machines/views/MachineList/StatusColumn/StatusColumn.js
```diff
@@ -99,17 +100,63 @@ const getStatusIcon = machine => {
...
+    ["override-failed-testing", "overrideFailedTesting"],
+    ["release", null],
+    ["rescue-mode", "rescueMode"],
+    ["test", null],
+    ["unlock", null]
+  ]);
+  Array.from(actionTypes.keys()).forEach(action => {
+    if (machine.actions.includes(action)) {
+      actionLinks.push({
+        children: `${action}...`,
```
Caleb-Ellis: In `general.machineActions.data` each action has a display name unhelpfully called `title`. Can we use that here instead of the raw action name?

huwshimi: That's a good idea, I will do that as a follow up as part of https://github.com/canonical-web-and-design/maas-ui/issues/793.

huwshimi:
@Caleb-Ellis I had presumed that it would do the right thing without the extra object, but I've updated it to include it so that the shape matches that of the angular client.

PR #797, opened 2020-02-14
Migrate some reducers to createReducer.

squidsoup:
## Done
* Migrate the majority of reducers with the same shape to redux-toolkit, generating 'standard' CRUD reducers using the `createStandardActions` utility.

## QA
* Ensure the app works as usual, without console errors.

## Fixes
Fixes canonical-web-and-design/MAAS-design#878

## Note
I've created [an additional card in the backlog](https://github.com/canonical-web-and-design/maas-ui/issues/801) with work that remains to fully complete the migration.

### DIFF
--- a/ui/src/app/utils/redux.js
+++ b/ui/src/app/utils/redux.js
@@ -1,4 +1,4 @@
-import { createAction } from "@reduxjs/toolkit";
+import { createAction, createReducer } from "@reduxjs/toolkit";
 
 /**
  * Generates async actions for 'start', 'error' and 'success' events,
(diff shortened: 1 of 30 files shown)


ui/src/app/utils/redux.js
```diff
@@ -61,10 +65,83 @@ export const createStandardActions = name => {
...
 
   action.cleanup = createAction(`CLEANUP_${name.toUpperCase()}`);
 
-  createAsyncActions(name, action);
+  createStandardAsyncActions(name, action);
 
   return action;
 };
+
+export const setStart = state => {
```
Caleb-Ellis: Do these functions need to be separated out? I feel like `createStandardReducer` should give the full story of how an object in state can change, without any further abstractions. And since `fetch` doesn't follow the pattern it's a little confusing.

PR #802, opened 2020-02-17
Display action menu in machine list.

Caleb-Ellis:
## Done
- Built non-interactive action menu in machine list. It displays a list of actions that the selected machines can perform along with a count of how many machines can perform it. If the count is 0 the action is filtered, unless it's a lifecycle action in which case it will show up disabled.
- Moved the `selectedMachines` state (and tests) up from `MachineList` to `Machines` so it can be shared with `HeaderStrip`. I'm not sure if it would be better for it to live in Redux. @squidsoup @huwshimi any thoughts?
- Removed `Routes` component so routing is now handled by `Machines`.

## QA
- Select a bunch of machines and click the take action menu.
- Check that the actions and counts match what's found in the angular version.

## Fixes
Fixes canonical-web-and-design/MAAS-squad#1778
Fixes canonical-web-and-design/MAAS-squad#1792

### Disabled
0 0 0 0_8400_MAAS_r_machines (11)

### Machines selected
0 0 0 0_8400_MAAS_r_machines (12)

### Opened
0 0 0 0_8400_MAAS_r_machines (13)


Caleb-Ellis:
Thanks @squidsoup, I've added a little bit of margin-left to the count which helps give a little breathing room. Also just to note down what I said at stand up - I removed the `Routes` component because with `selectedMachines` state now higher it would mean passing it from `Machines` to `Routes` to `MachineList` which seemed a bit long-winded. Most of the top-level view components like `Machines` kinda function as route components with a little styling wrapper anyway. I'm not married to either way though.

Issue #809, opened 2020-02-18
Power action indicator is too close to the text


lilyvidenova:
In the React implementation the caret down is too close to the text:
![image.png](https://images.zenhubusercontent.com/5af976724b5806bc2bd30860/6bb04395-7681-40f9-a322-ae35e3cf214f)

vs the Angular version:
![image.png](https://images.zenhubusercontent.com/5af976724b5806bc2bd30860/16f6eb93-4191-43b5-a9a3-33d8b5f02b7e)

PR #819, opened 2020-02-19
Truncate the power text

huwshimi:
## Done
- Correctly truncate the power text for small screens or when the action icon is visible.

## QA
- Open /r/machines.
- Reduce the size of your screen a bit until some of the power statuses don't have room. They should truncate with an ellipsis.
- Hover of the statuses, the menu chevron should appear and the statuses should further truncate.

## Fixes
Fixes: https://github.com/canonical-web-and-design/maas-ui/issues/809.

## Screenshots

Screen Shot 2020-02-20 at 10 42 03 am

### DIFF
--- a/ui/src/app/machines/views/MachineList/MachineList.scss
+++ b/ui/src/app/machines/views/MachineList/MachineList.scss
@@ -43,11 +43,11 @@
   }
   // Power
   &:nth-child(2) {
-    @include breakpoint-widths(23%, 15%, 12%, 10%, 8%);
+    @include breakpoint-widths(25%, 17%, 14%, 12%, 10%);
   }
   // Status
   &:nth-child(3) {
-    @include breakpoint-widths(32%, 30%, 20%, 20%, 17%);
+    @include breakpoint-widths(30%, 28%, 18%, 18%, 15%);
   }
   // Owner
   &:nth-child(4) {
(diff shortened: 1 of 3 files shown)


lilyvidenova:
Could we make the column a bit wider so the caret sits outside of the longest string (Unknown) similarly to how we do it in the angular version:

image

huwshimi:
@lilyvidenova after a bit of investigation it looks like the power column is set to the same size as the angular column, but a few months ago the tables in Vanilla added some left padding to table cells.

This change makes the table a little less dense, but also leaves less room for the content. Would you like me to overwrite that left padding for the machine table or try and make the cell slightly larger to account for the padding?

lilyvidenova:
@huwshimi this is more of a question for @lyubomir-popov - could you take a look and advice what we should do?

lyubomir-popov:
@lilyvidenova @huwshimi I would increase the power column width, maybe art the expense of the Status column?. The cell padding in maas is already quite tight at .5rem.

lyubomir-popov:
@huwshimi We're currently writing rules about icons in tables, might be relevant to this discussion. In short, icons should be .5rem away from adjacent related text (as here), and a fixed 1rem + any extra flexible space from unrelated text in adjacent columns.

huwshimi:
Thanks @lilyvidenova and @lyubomir-popov, I've increased the power column by 2% (and the status down 2%) and it seems to fit nicely at most screen sizes. Would you mind taking another look?

Screen Shot 2020-02-24 at 10 26 50 am

@lyubomir-popov it does seem like our icon spacing matches those rules so nothing to be gained there unfortunately.

lyubomir-popov:
@huwshimi I can't verify locally for some reason, but trying those width percentage in the browser looks good.

lilyvidenova:
Looks great, thanks @huwshimi 🚀

huwshimi:
Thanks @lilyvidenova!

Issue #813, opened 2020-02-18
No spinner when checking power


lilyvidenova:
The power icon should be changed to the spinner icon when checking power for a machine when using the in-table actions

huwshimi:
Related to: https://github.com/canonical-web-and-design/maas-ui/issues/781

Caleb-Ellis:
Still valid

Caleb-Ellis:
Actually a duplicate of #781

Issue #781, opened 2020-02-12
No way to determine if checking power is complete


huwshimi:
When performing a check power action from the machine list there is no way to determine when that action has completed.

We send the following over the websocket to do the power check:

```javascript
{
  method: "machine.check_power",
  type: 0,
  params: {system_id: "8acnxm"},
  request_id: 31
}
```

Which immediately responds with something like this (or an error):

```javascript
{
  type: 1,
  request_id: 31,
  rtype: 0,
  result: null
}
```

And then when the power update completes (I presume) a machine update is sent:

```javascript
  type: 2
  name: "machine",
  action: "update",
  data: {id: 23, system_id: "8acnxm", ... }
```

This final message does not have any details about which request it is from, so it could be from our request or from any other action that might have updated a machine (e.g. an action performed via the CLI or somewhere else in the web UI). It also may include the same data that we already have in the redux store.

We have this same problem with many parts of the UI (e.g. forms) but in those cases we assume that data changes are related to the user's action (and can do so in a somewhat more reliable fashion than this case).

Caleb-Ellis:
Still valid

PR #825, opened 2020-02-21
Display a spinner when checking power

huwshimi:
## Done
- Track state when waiting for a machine update.
- Display a spinner when awaiting a machine update.
Note: this may see the spinner prematurely stop as a machine update may be received from other actions. This issue is tracked here: https://github.com/canonical-web-and-design/maas-ui/issues/781.

## QA
- Load /r/machines.
- Click 'Check power' in the power menu.
- You should see a spinner.

## Fixes
Fixes: https://github.com/canonical-web-and-design/maas-ui/issues/813.

### DIFF
--- a/ui/src/app/base/reducers/machine/machine.js
+++ b/ui/src/app/base/reducers/machine/machine.js
@@ -35,6 +35,14 @@ const machine = produce(
             break;
           }
         }
+        draft.awaitingUpdate = false;
+        break;
+      case "CHECK_MACHINE_POWER_START":
+        draft.awaitingUpdate = true;
+        break;
+      case "CHECK_MACHINE_POWER_ERROR":
+        draft.awaitingUpdate = false;
+        draft.errors = action.error;
         break;
       case "ACQUIRE_MACHINE_ERROR":
       case "RELEASE_MACHINE_ERROR":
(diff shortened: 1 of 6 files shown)


ui/src/app/base/reducers/machine/machine.js
```diff
@@ -35,6 +35,14 @@ const machine = produce(
             break;
           }
         }
+        draft.awaitingUpdate = false;
+        break;
+      case "CHECK_MACHINE_POWER_START":
+        draft.awaitingUpdate = true;
+        break;
+      case "CHECK_MACHINE_POWER_ERROR":
```
Caleb-Ellis: `"CHECK_MACHINE_POWER_SUCCESS` case is missing

huwshimi: Interesting, are you not getting an `UPDATE_MACHINE_NOTIFY` when the power check completes?

huwshimi:
@Caleb-Ellis looking at the MAAS codebase it looks like "unkown" is returned when the machine reports a power type that MAAS doesn't know about (it gets an error status if it just failed when requesting the power status). Maybe it's disabled because there's no point requesting the power type again if MAAS doesn't know how to handle it?

ui/src/app/base/reducers/machine/machine.js
Caleb-Ellis: Nope, no `UPDATE_MACHINE_NOTIFY`. Maybe this only happens if  `CHECK_MACHINE_POWER` returns the same power state that the machine is in? i.e. MAAS thinks the machine is on, you check the power and it is indeed on = no update.

huwshimi: Ah OK, I think that shows that we can't show this spinner after all, thanks for double checking.

PR #829, opened 2020-02-21
Add extra macs fields to add machine form.

### LINKED ISSUE #794: Add "Extra MACs" button to Add machine form

Caleb-Ellis:
## Done
- Added extra macs button and extra macs fields to add machine form. Getting it to work with FormikField was tricky because of it being an optional field with variable size, so I've opted to instead build a more custom/imperative approach. I don't think this sort of pattern shows up anywhere else in MAAS.

## QA
- Go to /r/machines/add
- Check that you can add and remove extra MAC fields
- Submit the form with extra MACs
- Check that the network request includes an array of `extra_macs` and the machine is added successfully

## Fixes
Fixes #794 

## Screenshot
0 0 0 0_8400_MAAS_r_machines_add (1)

### DIFF
--- a/ui/src/app/machines/views/AddMachine/AddMachineFormFields/AddMachineFormFields.js
+++ b/ui/src/app/machines/views/AddMachine/AddMachineFormFields/AddMachineFormFields.js
@@ -1,5 +1,5 @@
-import { Col, Row, Select } from "@canonical/react-components";
-import React from "react";
+import { Button, Col, Input, Row, Select } from "@canonical/react-components";
+import React, { useState } from "react";
 import { useFormikContext } from "formik";
 import { useSelector } from "react-redux";
 
@@ -20,8 +20,10 @@ export const AddMachineFormFields = () => {
   const resourcePools = useSelector(resourcePoolSelectors.all);
   const zones = useSelector(zoneSelectors.all);
 
+  const [extraMACs, setExtraMACs] = useState([]);
+
   const formikProps = useFormikContext();
-  const { values } = formikProps;
+  const { setFieldValue, values } = formikProps;
 
   const architectureOptions = [
     {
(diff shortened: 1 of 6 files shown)


ui/src/app/machines/views/AddMachine/AddMachineFormFields/AddMachineFormFields.js
```diff
@@ -124,6 +126,49 @@ export const AddMachineFormFields = () => {
...
+            />
+            <Button
+              className="p-close-input"
+              hasIcon
+              onClick={() => {
+                const newExtraMACs = extraMACs.filter((_, j) => j !== i);
+                setExtraMACs(newExtraMACs);
+                setFieldValue("extra_macs", newExtraMACs);
+              }}
+              type="button"
```
huwshimi: Does this need a type?

ui/src/app/machines/views/AddMachine/AddMachineFormFields/AddMachineFormFields.js
```diff
@@ -124,6 +126,49 @@ export const AddMachineFormFields = () => {
...
+            >
+              <i className="p-icon--close" />
+            </Button>
+          </div>
+        ))}
+        <div className="u-align--right">
+          <Button
+            data-test="add-extra-mac"
+            hasIcon
+            onClick={() => setExtraMACs([...extraMACs, ""])}
```
huwshimi: Because new macs are set to an empty string these get included with the submitted data. It appears the API ignores them, but it might be nice to filter them out before the data is submitted.

ui/src/app/machines/views/AddMachine/AddMachineFormFields/AddMachineFormFields.js
Caleb-Ellis: Yes, otherwise I get the following error from Formik:
```
Warning: You submitted a Formik form using a button with an unspecified `type` attribute
Most browsers default button elements to `type="submit"`.
If this is not a submit button, please add `type="button"`
```

ui/src/app/machines/views/AddMachine/AddMachineFormFields/AddMachineFormFields.js
Caleb-Ellis: Good idea, adding a filter in now

PR #850, opened 2020-02-27
Reset add machine/chassis form if "add another" button used.

### LINKED ISSUE #844: Clear form when using "Save and add another" submit
The normal submit method on form cards redirects back to the "main" page (e.g "Save machine" redirects from `/machines/add` to `/machines`) however the "Save and add another" submit method does not redirect. In this case it would be nice for the form to reset to initial values.

Implementing `resetForm` after form submit is pretty straightforward with formik (`resetForm({ values: initialValues })`) however the form is reset regardless of any errors returned from the api. I imagine it's more annoying to have to re-enter form fields than clearing stale ones, so the current implementation doesn't reset the form at all.

We should implement a method to clear the form on *successful* submit when using the "Save and add another" button.

Caleb-Ellis:
## Done
- Reset add machine/chassis form if user successfully submits with the "Save and add another" button.
- Reworded some FormikForm props from `redirectOnSave` to `resetOnSave` (and consequently reversed logic). Seems a bit clearer for me to highlight resetting a form than handling redirects in the form logic.

## QA
- Go to `/MAAS/r/machines/add`
- Fill in the fields and add a machine using the "Save and add another" button.
- Check the form resets on successful submit and a notification confirms the machine is added.
- Add another machine using values that will return an error from the backend (i.e. duplicate hostname/MAC, unknown virsh address, etc.)
- Submit the form with "Save and add another"
- Check that an error notification shows up and the form does not get reset.
- Do the same at `/MAAS/r/machines/chassis/add`

## Fixes
Fixes #844

### DIFF
--- a/ui/src/app/base/components/FormikForm/FormikForm.js
+++ b/ui/src/app/base/components/FormikForm/FormikForm.js
@@ -17,6 +17,7 @@ const FormikForm = ({
   onSaveAnalytics = {},
   onSubmit,
   onValuesChanged,
+  resetOnSave,
   saving,
   saved,
   savedRedirect,
@@ -56,7 +57,9 @@ const FormikForm = ({
         allowAllEmpty={allowAllEmpty}
         buttons={buttons}
         errors={errors}
+        initialValues={initialValues}
         onValuesChanged={onValuesChanged}
+        resetOnSave={resetOnSave}
         saving={saving}
         saved={saved}
         secondarySubmit={secondarySubmit}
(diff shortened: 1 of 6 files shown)


huwshimi:
This all worked except the extra mac addresses didn't get cleared:

Screen Shot 2020-02-28 at 9 44 39 am

Caleb-Ellis:
Thanks @huwshimi, should be fixed now :)

PR #855, opened 2020-03-03
Add machine searching

huwshimi:
## Done
- Ported search utilities from the angular client. Includes bug fixes, additional tests, es6ificiation and cleanupification.

## QA
- Visit /r/machines.
- Search using the search box. You should be able to free search against the machine attributes and also specify specific values e.g. hostname:cow.

## Fixes
Fixes: https://github.com/canonical-web-and-design/maas-squad/issues/1488.

### DIFF
--- a/ui/src/app/machines/filter-nodes.js
+++ b/ui/src/app/machines/filter-nodes.js
@@ -1,0 +1,179 @@
+import { getCurrentFilters } from "./search";
+
+// Helpers that convert the pseudo field on node to an actual
+// value from the node.
+const mappings = {
+  cpu: node => node.cpu_count,
+  cores: node => node.cpu_count,
+  ram: node => node.memory,
+  mac: node => [node.pxe_mac, ...node.extra_macs],
+  zone: node => node.zone.name,
+  pool: node => node.pool.name,
+  pod: node => node.pod && node.pod.name,
+  "pod-id": node => node.pod && node.pod.id,
+  power: node => node.power_state,
+  release: node =>
+    node.status_code === 6 || node.status_code === 9
+      ? node.osystem + "/" + node.distro_series
+      : "",
+  hostname: node => node.hostname,
+  vlan: node => node.vlan,
+  rack: node => node.observer_hostname,
+  subnet: node => node.subnet_cidr,
+  numa_nodes_count: ({ numa_nodes_count, numa_nodes }) => {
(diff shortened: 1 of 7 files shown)


ui/src/app/machines/filter-nodes.js
```diff
@@ -0,0 +1,179 @@
+import { getCurrentFilters } from "./search";
```
huwshimi: This file is ported from the angular client.

ui/src/app/machines/filter-nodes.test.js
```diff
@@ -0,0 +1,410 @@
+import filterNodes from "./filter-nodes";
```
huwshimi: This file is ported from the angular client.

ui/src/app/machines/search.js
```diff
@@ -0,0 +1,124 @@
+// Holds all stored filters.
```
huwshimi: This file is ported from the angular client.

ui/src/app/machines/search.test.js
```diff
@@ -0,0 +1,259 @@
+import {
```
huwshimi: This file is ported from the angular client.

ui/src/app/machines/views/MachineList/MachineList.js
```diff
@@ -458,10 +465,6 @@ const MachineList = ({ selectedMachines = [], setSelectedMachines }) => {
     dispatch(zoneActions.fetch());
   }, [dispatch, machinesLoaded]);
 
-  useEffect(() => {
```
huwshimi: This is replaced with `useMemo` above. It seemed that this was only in state to prevent unnecessary calls, but as this was updating state it meant that any state updates to machines was causing an infinite loop. Replaced with useMemo as it is probably the correct way of preventing reruns.

squidsoup:
Works great @huwshimi, thanks. Are we able to begin the search after the user has typed three characters rather than four, given we can have machine names of three characters?

huwshimi:
Thanks a lot @squidsoup. There isn't anything that should delay the search until a certain number of characters has been entered. The search matches across all machine attributes so I wonder if somehow your searches have been matching multiple machines until 4 characters are entered. I'm going to land this branch to unblock my in-progress branch, but I've filed an issue to confirm: https://github.com/canonical-web-and-design/maas-ui/issues/868.

PR #885, opened 2020-03-10
Enable production source maps.

### LINKED ISSUE #857: Re-enable sourcemaps on production builds
Helpful for the engineers (and us) to debug

squidsoup:
## Done
Enable production source maps, allowing for more useful sentry reports. Clients will not normally download `.map` files, so this should have no affect on production (unless someone opens the devtools).

## QA
* Run `yarn build` for `ui`, should build with source maps.

## Fixes
Fixes #857


huwshimi:
@squidsoup is there any way to enable the maps for the angular client as well?

squidsoup:
> @squidsoup is there any way to enable the maps for the angular client as well?

Yes, but we need separate dev/prod builds for the angular client first (we have a card for that).

PR #886, opened 2020-03-10
Drop old circleci jobs.

squidsoup:
## Done
* Drop circleci jobs we've migrated to github actions.

Config validated:

```
 $ circleci config validate 
Config file at .circleci/config.yml is valid.
```

## QA
* None, tests should pass.


huwshimi:
Thanks @squidsoup. I've change the PR rules to not require the jobs you've removed and also to require the new actions job.

PR #887, opened 2020-03-10
Fix machine list checkboxes highlighting when action cannot be performed.

### LINKED ISSUE #871: Colours for machine list checkboxes have disappeared
Once upon a time there were blue *and* yellow checkboxes for when an action and multiple machines were selected:
Blue = selected machine can perform selected action
Yellow = selected machine cannot perform selected action
The same logic would apply to the group checkboxes and all checkbox, i.e. if at least one machine in a group cannot perform the action, the group checkbox becomes yellow.

Caleb-Ellis:
## Done
- Fixed Vanilla regression in which checkboxes would turn yellow if the action could not be performed on the selected machine.

## QA
- Select a bunch of machines that can't all perform an action.
- Select that action and see that the warning to update selection shows up, and the machines that can't perform the action have a yellow checkbox instead of a blue one. The group(s) and header checkboxes should also be yellow if the machine inside can't perform the action.

## Fixes
Fixes #871 

## Screenshot
0 0 0 0_8400_MAAS_ (7)


legacy/src/app/directives/tests/test_machines_table.js
```diff
@@ -1117,4 +1115,42 @@ barbaz`); // Has to be formatted this way for tooltip
...
+      machines[2].actions = ["action1"];
+      scope.actionOption = { name: "action2" };
+
+      expect(scope.nodeSelectionNotActionable(machines)).toEqual(true);
+    });
+
+    it("returns false if all nodes in selection can perform action", () => {
+      const directive = compileDirective();
+      const scope = directive.isolateScope();
+      const machines = Array.from(Array(3)).map(makeMachine);
```
sparkiegeek: Am I being dumb? Why make a new copy of an empty array that you already have?
```suggestion
      const machines = Array(3).map(makeMachine);
```

Caleb-Ellis: You're not being dumb, I'm just being copy+pastey. `Array(3).map(makeMachine)` won't work though because the array is actually empty - a spooky ghost array that has a length of 3 *and nothing else*, so it can't actually be mapped. I think that's what the original code was trying to get around, but it's probably easier and more readable for it to be `Array(3).fill().map(makeMachine)` which fills it with something iterable first.

PR #889, opened 2020-03-10
Convert component scss into mixins + import all styles into index.scss file.

### LINKED ISSUE #747: Remove instances of importing scss directly into components
While trying to find out what was causing regressions in the Vanilla 2.6.0 upgrade branch, @lyubomir-popov and I have discovered it is incredibly difficult to determine where styles originate. It appears that settings defined in the root `scss/_settings.scss` file are being overwritten, but it is unclear by what. Also, the use of sass-loader could potentially be leading to duplicated scss when importing Vanilla files directly.

Currently we are handling scss in a few different ways:
- Sass files in the project root for components (e.g. `_patterns_double-row.scss`);
- Sass files in the project root for styles that are used across different components (e.g. `_patterns_table.scss`);
- Sass files in the project root for global styles and importing vanilla (e.g. `base.scss`, `_settings.scss`);
- Some components import custom scss directly into the JS file, which may or may not import Vanilla files directly (e.g. `SideNav.scss`).

The confusion is compounded when taking into account we are also importing components from the react-components library, some of which import custom scss directly.

We should determine and document the best way of handling scss in MAAS in a way that is consistent and not prone to duplicated styles.

Caleb-Ellis:
## Done
- Converted all component scss into `_index.scss`'s (for nicer import paths) and removed them from being directly imported into React components.
- Wrapped all "global" styles in mixins.
- Created a top-level `index.scss` file where all scss in the project is listed. I think this is nicer because now we only have to import vanilla+settings only once, instead of as needed per-component (which I believe was also causing some settings overrides).
- Upgraded to Vanilla 2.8.0 and fixed regressions + removed overrides (mainly to do with contextual menu positioning and form spacing).
- Removed some "easy win" redundant styles that were either fixed in the latest Vanilla or copied unnecessarily from legacy, but we should definitely go over what we have more thoroughly.

## QA
- Click around everywhere in the React pages and check that styling still looks ok everywhere.
- See if you agree with the folder structure

## Fixes
Fixes #747


huwshimi:
Thanks for doing all these changes. I've noticed a couple of things, it's possible that some are on master but worth fixing (could be in a followup if needed).

Generally the forms that are in a card are missing space between the last input and the horizontal line:
Screen Shot 2020-03-11 at 10 45 37 am

The forms that aren't in a card are generally OK but I noticed this one in the user preferences:
Screen Shot 2020-03-11 at 10 46 31 am

The DHCP snippets have a bit of extra space below them:
Screen Shot 2020-03-11 at 10 49 59 am

ui/src/app/base/components/ContextualMenu/ContextualMenu.js
```diff
@@ -9,7 +9,11 @@ const getPositionStyle = (position, wrapper, constrainPanelWidth) => {
   if (!wrapper || !wrapper.current) {
     return undefined;
   }
-  const { bottom, left, width } = wrapper.current.getBoundingClientRect();
+  const {
+    bottom,
+    left,
+    width
+  } = wrapper.current.children[0].getBoundingClientRect();
```
huwshimi: Instead of accessing the children would it be possible to move the ref to the correct node?

ui/src/scss/_vanilla-overrides.scss
```diff
@@ -4,17 +4,6 @@
...
 
-@import "~vanilla-framework/scss/settings_colors";
-@import "~vanilla-framework/scss/settings_spacing";
-@import "~vanilla-framework/scss/patterns_icons";
-@import "~vanilla-framework/scss/global_functions";
-
-// 19-08-2019 Caleb: Form submit button too close to input when $multi = 1
-// https://github.com/canonical-web-and-design/vanilla-framework/issues/2497
-.p-form__group:last-child {
-  margin-bottom: $spv-outer--medium;
```
huwshimi: As noted in the QA notes maybe this one can't be removed yet?

Caleb-Ellis:
> Generally the forms that are in a card are missing space between the last input and the horizontal line:
> Screen Shot 2020-03-11 at 10 45 37 am

@huwshimi I think this looks fine given that the spacing is supposed to be dense. Having collated the custom scss we have in the project (and seen how much it is) I'd rather let Vanilla define the spacing, and only have an override if it looks truly awful.



> The forms that aren't in a card are generally OK but I noticed this one in the user preferences:

Fixed by adding a little spacer class to the div. I think that's fine given this is a bit of a special case i.e. it's not an input in the form.

screenshot_8



> The DHCP snippets have a bit of extra space below them:

Thanks. Fixed.

screenshot_9

ui/src/app/base/components/ContextualMenu/ContextualMenu.js
Caleb-Ellis: Hmm I can't pass it to the `Button` because you can only assign `ref`s to normal DOM elements, and there isn't currently a `useForwardRef` hook. I've added a guard in the function so that if a child isn't present it just gets the dimensions of the span, but otherwise it gets the first child's dimensions. Would that be ok for now?

ui/src/scss/_vanilla-overrides.scss
Caleb-Ellis: Upgrading to Vanilla 2.8.0 meant this rule broke a few other things. :( I've left it removed but I've added another issue for a slightly different regression.

ui/src/app/base/components/ContextualMenu/ContextualMenu.js
huwshimi: Yep, that's better I reckon.

huwshimi:
> > Generally the forms that are in a card are missing space between the last input and the horizontal line:
> > Screen Shot 2020-03-11 at 10 45 37 am
> 
> @huwshimi I think this looks fine given that the spacing is supposed to be dense. Having collated the custom scss we have in the project (and seen how much it is) I'd rather let Vanilla define the spacing, and only have an override if it looks truly awful.

That's fine, it just differs from the designs, but maybe check with them if you're concerned.

Issue #881, opened 2020-03-09
Action menu - handle action progress


Caleb-Ellis:
This could entail a custom hook that determines whether none/some/all of the selected machines have had the selected action complete, and only close the dropdown if everything is completed successfully.

In the angular version a spinner shows with the text `x of y machines are transitioning to z`.

Caleb-Ellis:
This is likely to be made possible by #928

PR #890, opened 2020-03-11
Build form components for actions that do not require fields.

Caleb-Ellis:
## Done
- Updated `ActionForm` to dispatch actions for forms that don't have fields, which are the following:
  - Acquire
  - Release
  - Abort
  - Power on
  - Power off
  - Lock
  - Unlock
  - Rescue mode
  - Exit rescue mode
  - Mark broken
  - Mark fixed
  - Delete (red button)
- Added a string util function `kebabToCamelCase` for converting the websocket method names into our redux action names.

## QA
- Select some machines and check that you can successfully perform the above actions, taking note that the delete action should remove the machine from `items` state and `selected` state.
- Check that the delete button is red instead of green.
- Note that errors and progress will be tackled in #881 - this PR merely makes it possible to perform these actions in batch. 

## Fixes
Fixes canonical-web-and-design/MAAS-squad#1783

### DIFF
--- a/ui/src/app/base/actions/machine/machine.js
+++ b/ui/src/app/base/actions/machine/machine.js
@@ -56,10 +56,10 @@ machine.setZone = (systemId, zoneId) =>
     zone_id: zoneId
   });
 
-machine.turnOff = systemId =>
+machine.off = systemId =>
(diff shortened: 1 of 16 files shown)


huwshimi:
This one has a conflict @Caleb-Ellis

PR #898, opened 2020-03-16
Improve search DSL and add selected filtering

huwshimi:
## Done
- Clean up DSL destructuring.
- Implement selected filtering.
- Fix DSL bugs.

## QA
- Visit /r/machines.
- Select some machines.
- In the search box search for `in:selected`, you should only see your selected machines.
- In the search box search for `in:!selected`, you should not see your selected machines.
- Check that you can search in other ways according to: https://maas.io/docs/interactive-search

## Fixes
Fixes: https://github.com/canonical-web-and-design/maas-squad/issues/1799.

### DIFF
--- a/ui/src/app/base/selectors/machine/machine.js
+++ b/ui/src/app/base/selectors/machine/machine.js
@@ -69,12 +69,16 @@ machine.getBySystemId = (state, id) =>
 machine.errors = state => state.machine.errors;
 
 /**
- * Gets the search terms.
+ * Gets the search terms and selected machines.
  * @param {Object} state - The redux state.
- * @param {Array} terms - The search terms.
- * @returns {Array} The search terms.
+ * @param {String} terms - The search string.
+ * @param {Array} selectedIDs - The selected machine ids.
+ * @returns {Array} The search terms and selected machines.
  */
-machine._getTerms = (state, terms) => terms;
+machine._getSearchParams = (state, terms, selectedIDs) => ({
+  terms,
+  selectedIDs
+});
 
 /**
  * Get machines that match terms.
@@ -83,12 +87,12 @@ machine._getTerms = (state, terms) => terms;
  * @returns {Array} A filtered list of machines.
  */
 machine.search = createSelector(
-  [machine.all, machine._getTerms],
-  (items, terms) => {
+  [machine.all, machine._getSearchParams],
+  (items, { terms, selectedIDs }) => {
     if (!terms) {
       return items;
     }
-    return filterNodes(items, terms);
+    return filterNodes(items, terms, selectedIDs);
   }
 );
 
--- a/ui/src/app/machines/filter-nodes.js
+++ b/ui/src/app/machines/filter-nodes.js
@@ -36,7 +36,7 @@ const matches = (value, lowerTerm, exact, negate) => {
   return negate ? !match : match;
 };
 
-const filterNodes = (nodes, search) => {
+const filterNodes = (nodes, search, selectedIDs) => {
   if (
     typeof nodes === "undefined" ||
     typeof search === "undefined" ||
@@ -45,6 +45,10 @@ const filterNodes = (nodes, search) => {
     return nodes;
   }
   const filters = getCurrentFilters(search);
+  if (!filters) {
+    // No matching filters were found.
+    return [];
+  }
   Object.entries(filters).forEach(([attr, terms]) => {
     if (terms.length === 0) {
       // If this attribute has no associated terms then skip it.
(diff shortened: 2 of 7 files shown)


ui/src/app/machines/filter-nodes.js
```diff
@@ -54,15 +58,15 @@ const filterNodes = (nodes, search) => {
...
-        nodes.forEach(node => {
-          if (node.$selected && term.toLowerCase() === "selected") {
-            matched.push(node);
-          } else if (!node.$selected && term.toLowerCase() === "!selected") {
-            matched.push(node);
-          }
-        });
-        nodes = matched;
+        if (term.toLowerCase() === "selected") {
+          nodes = nodes.filter(({ system_id }) =>
```
hatched: You're redefining the function arguments, typically this is a no-no. I'm surprised the linter didn't catch it...

I probably just don't have enough context but this seems odd...So for each `term` you're redefining `nodes` with a `filter` result? So any subsequent `term` will override any previous values. Then once `terms` is exhausted it just returns that value? It looks like you could possibly return the result from a `terms.filter` on line 60.

Or you an disregard all of this if I'm way off base 😄

ui/src/app/base/selectors/machine/machine.js
```diff
@@ -69,12 +69,16 @@ machine.getBySystemId = (state, id) =>
...
+ * Gets the search terms and selected machines.
  * @param {Object} state - The redux state.
- * @param {Array} terms - The search terms.
- * @returns {Array} The search terms.
+ * @param {String} terms - The search string.
+ * @param {Array} selectedIDs - The selected machine ids.
+ * @returns {Array} The search terms and selected machines.
  */
-machine._getTerms = (state, terms) => terms;
+machine._getSearchParams = (state, terms, selectedIDs) => ({
```
hatched: This is interesting way to curry arguments into selectors. You may prefer how we do it on the jaas-dashboard https://github.com/canonical-web-and-design/jaas-dashboard/blob/master/src/app/selectors.js#L448

ui/src/app/machines/search.js
```diff
@@ -10,55 +10,39 @@ export const getCurrentFilters = search => {
...
-          filter.push(termItem);
-        }
-      });
-      // If the term contains the ending ')' then it's the last in the group.
-      if (term.includes(")")) {
-        processingFilter = null;
+  // Match filters with parens e.g. 'status:(new,deployed)'.
+  // Then: match filters without parens e.g. 'status:new,deployed' or 'status'
+  const filterMatchingRegex = search.matchAll(/(\b\w+:!?\([^)]+\))|(\w+\S+)/g);
+  Array.from(filterMatchingRegex).forEach(([group]) => {
```
hatched: You could use the spread operator to make this a bit shorter if you'd like.
```JavaScript
[...filterMatchingRegex]
```

cristinadresch:
@hatched do you want to give it a go reviewing this one?

hatched:
@cristinadresch I have already....and no updates have been made since my comments were submitted.

cristinadresch:
@huwshimi spoke to Jeff, he said he didn't want to out code -1 because he doesn't have enough context. but if you are making any changes and this is not ready for another review, could you remove the labels? just so we know what it's happening in our standup.

cc @hatched

ui/src/app/machines/search.js
huwshimi: Done.

ui/src/app/base/selectors/machine/machine.js
huwshimi: I decided against this approach as it makes using the selector more complicated (looks like you wrap the selector creator in `useMemo` before you use `useSelector` on the resulting function.

ui/src/app/machines/filter-nodes.js
huwshimi: Fixed (and a few bugs found along the way too).

huwshimi:
This required significant reworking after @hatched's comments. I've just finished those so ready for another review.

@hatched your comments have been very helpful, I'd be happy with your review if you have time to do another one.

@lilyvidenova maybe you could QA this one to save @hatched having to get a MAAS set up. Would that be OK?

ui/src/app/base/selectors/machine/machine.js
hatched: Sure no problem, maybe we can convert the dashboard to use this approach instead. Something to investigate.

lilyvidenova:
@huwshimi terribly sorry for still blocking this, however I can't get my maas-ui to see the internet through multipass. Ant kindly added the ./run back to the ui today, could you maybe include that bit into this branch so I can run it on my mac in order to QA it?

huwshimi:
@lilyvidenova this has now been updated with the `./run` script.

lilyvidenova:
Looks great, thanks @huwshimi

Issue #910, opened 2020-03-24
UI fails when no access to MAAS


huwshimi:
**Describe the bug**
When maas-ui can't proxy requests to a MAAS it doesn't display anything on the page, but in the browser console there are the following errors:

```
Failed to load resource: the server responded with a status of 504 (Gateway Timeout)
index.js:1 A non-serializable value was detected in an action, in the path: `error`. Value: TypeError: Cannot read property 'includes' of null
    at handlePromise (http.js:30) 
Take a look at the logic that dispatched this action:  Object 
(See https://redux.js.org/faq/actions#why-should-type-be-a-string-or-at-least-serializable-why-should-my-action-types-be-constants)
console.<computed> @ index.js:1
index.js:1 A non-serializable value was detected in the state, in the path: `status.error`. Value: TypeError: Cannot read property 'includes' of null
    at handlePromise (http.js:30) 
Take a look at the reducer(s) handling this action type: LOGIN_ERROR.
(See https://redux.js.org/faq/organizing-state#can-i-put-functions-promises-or-other-non-serializable-items-in-my-store-state)
```

And the shell shows:

```
[2] [HPM] Error occurred while trying to proxy request /MAAS/accounts/login/ from dev.local:8400 to http://192.168.64.4:5240/ (ECONNREFUSED) (https://nodejs.org/api/errors.html#errors_common_system_errors)
[2] [HPM] Error occurred while trying to proxy request /MAAS/accounts/login/ from dev.local:8400 to http://192.168.64.4:5240/ (ECONNREFUSED) (https://nodejs.org/api/errors.html#errors_common_system_errors)
```

**To Reproduce**
- Open `proxy/.env.local` and change `MAAS_URL` to URL/IP that doesn't exist.
- Load the ui.
- You should be given the login page, with errors in the console.

**Expected behavior**
This should probably display a error message to the user.

PR #911, opened 2020-03-25
Handle server down errors

huwshimi:
## Done
- Handle errors when checking if the user is authenticated and the server is down.

## QA
- Open proxy/.env.local and change MAAS_URL to a URL/IP that DOESN'T exist.
- Load the UI.
- You should see a message about not being able to access the server.
- Open proxy/.env.local and change MAAS_URL to URL/IP that DOES exist.
- Load the UI.
- You should see the login page or a logged in page (if you were previously logged in).

## Fixes
Fixes: https://github.com/canonical-web-and-design/maas-ui/issues/910.


huwshimi:
Hmm this branch breaks Cypress as it no longer shows the login page if there is no server to log into. I think this should be resolved by https://github.com/canonical-web-and-design/maas-ui/pull/641.

squidsoup:
Thanks for addressing this @huwshimi - aware that this has been an issue for a long time!

PR #912, opened 2020-03-27
Link from pools to machines

huwshimi:
## Done
- Link from pools to the machine list.

## QA
- Visit /r/pools.
- Add a pool (using the Add pool button in the top right) and add 1 or more machines to that pool (click on a machine from the machine list and then click on Configuration and change 'Resource pool').
- Add another pool and leave it empty.
- Go back to the pool list and the pool with no machines should have "Empty pool" in the machines column.
- For a pool with machines you should see something like "0 of 1 ready". Click on that link. You should be take to the machine list with only the machines in that pool visible.

## Fixes
Fixes: https://github.com/canonical-web-and-design/maas-squad/issues/1807.

### DIFF
--- a/ui/src/app/pools/views/Pools.js
+++ b/ui/src/app/pools/views/Pools.js
@@ -1,6 +1,7 @@
 import React, { useEffect, useState } from "react";
 import { useDispatch, useSelector } from "react-redux";
 import { Col, Loader, MainTable, Row } from "@canonical/react-components";
+import { Link } from "react-router-dom";
 
 import TableActions from "app/base/components/TableActions";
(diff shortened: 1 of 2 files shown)


lilyvidenova:
@huwshimi when the filter is applied this way, it doesn't get reflected in the URL and it should. Other than that works well!

squidsoup:
@lilyvidenova I've QA'd this in both Chrome and FF and the link from pools definitely has the correct filter term in the querystring. Maybe the branch you QA'd from was out of sync somehow?

image

Going to mark this QA+1, but if the problem persists for some reason, please let @huwshimi know @lilyvidenova.

PR #937, opened 2020-04-03
Fix recursive header rendering.

### LINKED ISSUE #872: Investigate slowness (infinite loop) when using the UI for a while
The UI can become unresponsive in production.

## Steps to replicate
Switch repeatedly between any two views.

squidsoup:
## Done
* Fix recursive header rendering by ensuring routeChangeSuccess handler is only setup once on app init.

Note: only relevant changes are 368, 389, the rest prettier touched. 

## QA
* Load the app, and click between _angular_ routes for 20s or so. The app should not get progressively slower and crawl to a standstill.

## Fixes
Fixes #872

### DIFF
--- a/legacy/src/app/entry.js
+++ b/legacy/src/app/entry.js
@@ -34,7 +34,7 @@ import {
   filterLinkModes,
   filterEditInterface,
   filterSelectedInterfaces,
-  filterVLANNotOnFabric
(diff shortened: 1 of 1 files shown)


lilyvidenova:
@squidsoup should we backport this to 2.7?

PR #944, opened 2020-04-03
Build test form in action menu dropdown.

Caleb-Ellis:
## Done
- Built TestForm component
- Allow TagSelector to include descriptions

## QA
- Select some machines that can be tested and choose "Test" from the action dropdown
- Check that you can select from all the testing scripts in the maas db
- Choose a script that requires a URL (by default the only one is `internet-connectivity`) and check that an extra field shows up for entering the URL
- Submit the form and check that the selected machines run the selected tests (go to Machine details > Hardware tests and the selected tests should be either pending or running)

## Fixes
Fixes canonical-web-and-design/MAAS-squad#1785

### Default
Screenshot_2020-04-03 Machines bolla MAAS

### Expanded
Screenshot_2020-04-03 Machines bolla MAAS(1)

### With URL test
Screenshot_2020-04-03 Machines bolla MAAS(2)

### DIFF
--- a/ui/src/app/base/actions/machine/machine.js
+++ b/ui/src/app/base/actions/machine/machine.js
@@ -92,8 +92,12 @@ machine.deploy = (systemId, extra = {}) =>
 machine.abort = (systemId) =>
   generateMachineAction("ABORT_MACHINE", "abort", systemId);
 
-machine.test = (systemId) =>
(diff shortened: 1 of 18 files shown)


squidsoup:
@Caleb-Ellis when all scripts have been selected, the "select scripts" input should be disabled.

PR #948, opened 2020-04-06
Manage the list of loaded models in the saga module

huwshimi:
## Done
- To deal with multiple components requesting the same data at once the loaded models are stored in the saga module.
- Drive-by to disconnect from the websocket when logging out.

## QA
- View /r/machines.
- In your network tab check the websocket for machine.list. You should only see one.
- Log out and back in, your models should load again.

### DIFF
--- a/ui/src/app/App.js
+++ b/ui/src/app/App.js
@@ -123,6 +123,7 @@ export const App = () => {
         )}
         location={location}
         logout={() => {
+          dispatch(statusActions.websocketDisconnect());
           dispatch(statusActions.logout());
         }}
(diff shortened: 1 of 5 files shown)


ui/src/app/App.js
```diff
@@ -123,6 +123,7 @@ export const App = () => {
         )}
         location={location}
         logout={() => {
+          dispatch(statusActions.websocketDisconnect());
```
huwshimi, line 126: This is a drive-by to disconnect from the websocket when logging out. This ws connection should not be left open if you're not  authenticated.

PR #949, opened 2020-04-06
Run tests before build in CI

huwshimi:
## Done
- Run the tests before build-all in CI so it fails sooner (build-all takes multiple minutes).

## QA
None, CI should pass.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -22,7 +22,7 @@ jobs:
           node-version: ${{ matrix.node-version }}
       - run: CYPRESS_INSTALL_BINARY=0 yarn install
       - run: yarn lint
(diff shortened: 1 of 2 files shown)


squidsoup:
@huwshimi ah, there was a reason for that, we need to build the shared library first. Perhaps a compromise would be to build the shared library first, run the test suite, and then run build-all.

huwshimi:
@squidsoup I've made building components part of the test script so command can never run without the required files. I seem to remember you hit that issue after clean-all locally.

squidsoup:
> @squidsoup I've made building components part of the test script so command can never run without the required files. I seem to remember you hit that issue after clean-all locally.

Good solution, thanks @huwshimi

PR #956, opened 2020-04-07
Add power state and owner count to pod list.

Caleb-Ellis:
## Done
- Added power state if that data exists (which it should for a LXD host)
- Removed Virsh address + truncated host type (e.g. "Virsh (virtual systems)" becomes "Virsh")
- Added owner count
- Removed Actions column
- Rewrote some code so that the pod object returned from the websocket is not mutated directly.
- Fixed some code that only handled rsd or virsh pod types.
- Various styling fixes (long storage pool name in popover, chart with >40 cores,

## QA
- Check that the KVM list matches the [design](https://app.zeplin.io/project/5e538f1156434564ebb73903/screen/5e8da399a001e0235c277f1d), noting that the OS will be added in #1003

## Fixes
Fixes canonical-web-and-design/MAAS-squad#1873

### Loading
Screenshot_2020-04-20 Loading…

### Loaded
Screenshot_2020-04-20 KVM bolla MAAS(1)


legacy/src/scss/_patterns_cores-chart.scss
```diff
@@ -7,6 +7,7 @@ $bar-height: 11px;
 @mixin maas-p-cores-chart {
   maas-cores-chart {
```
anthonydillon, line 8: Huh is this used? Neither a placeholder or class?

Caleb-Ellis: It's weird, but yeah this is used in the code. It's kinda like css specific for Angular's version of web components.

PR #962, opened 2020-04-08
Fix inconsistent header strip sizes in angular pages.

### LINKED ISSUE #961: Header strip changes height depending on the page

Caleb-Ellis:
## Done
- Fixed inconsistent header strip sizes in angular pages.

## QA
- Navigate between angular pages (and react pages if you want) and check that the header strip does not change height, with the exception of pages with tab strips (e.g. machine list, node details).
- Reload the page and check that the transition from Loading... > Connecting... > Page title does not change the header height.

## Fixes
Fixes #961

### DIFF
--- a/legacy/src/app/partials/directives/error_overlay.html
+++ b/legacy/src/app/partials/directives/error_overlay.html
@@ -1,11 +1,10 @@
-<header id="error-header" class="page-header" data-ng-show="show()">
+<header id="error-header" class="p-strip--light is-shallow page-header" data-ng-show="show()">
     <div class="row">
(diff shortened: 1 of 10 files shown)


squidsoup:
@Caleb-Ellis looks like we have some broken tests. 👀

Caleb-Ellis:
@squidsoup yeah bit of a weird one. Updating the error_overlay.html markup broke all the tests in error_overly.js. To fix it I've used `ng-hide` instead of `ng-if` and I'm not entirely sure why this works!

PR #963, opened 2020-04-08
Fix form validation when adding machine with multiple MACs.

### LINKED ISSUE #905: Cannot add mutiple MAC addresses in GUI
[LP bug 1865122](https://bugs.launchpad.net/maas/+bug/1865122)

When I add MAC addresses in GUI does not allow to enter MAC addresses randomly.
Reordering addresses helped me.

The screenshot shows that the same address is validated in one field, but not in another.

image

Caleb-Ellis:
## Done
- Fixed form validation for extra MACs fields when adding new machine.

## QA
- Go to angular machine list > Add hardware > Machine and add extra macs
- Type into the macs fields and check that they validate correctly (just a red border in Vanilla 2.0.1)

## Fixes
Fixes #905

### DIFF
--- a/legacy/src/app/partials/nodelist/add-machine.html
+++ b/legacy/src/app/partials/nodelist/add-machine.html
@@ -90,28 +90,32 @@
             </div>
           </div>
           <div class="p-form__group row" data-ng-repeat="mac in machine.macs">
-            <label for="mac-address" class="p-form__label col-2" ng-class="{ 'is-required': mac === machine.macs[0] }">
-              <span data-ng-if="mac === machine.macs[0]">MAC Address</span>&nbsp;
+            <label
+              class="p-form__label col-2"
+              for="mac-address-{$ $index $}"
+              ng-class="{ 'is-required': $first }"
+            >
+              <span ng-if="$first">MAC Address</span>
+              <span class="u-hide" ng-if="!$first">Extra MAC {$ index $}</span>
             </label>
             <div class="col-4 p-form__control p-search-box u-no-margin--bottom">
               <input
-                type="text"
-                id="mac-address"
-                placeholder="00:00:00:00:00:00"
-                maxlength="17"
-                data-ng-class="{ 'has-error': mac.error }"
-                data-ng-model="mac.mac"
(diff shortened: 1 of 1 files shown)


legacy/src/app/partials/nodelist/add-machine.html
```diff
@@ -90,18 +90,17 @@ <h3 class="page-header__dropdown-title">Add machine</h3>
             </div>
           </div>
           <div class="p-form__group row" data-ng-repeat="mac in machine.macs">
-            <label for="mac-address" class="p-form__label col-2" ng-class="{ 'is-required': mac === machine.macs[0] }">
-              <span data-ng-if="mac === machine.macs[0]">MAC Address</span>&nbsp;
+            <label for="mac-address-{$ $index $}" class="p-form__label col-2" ng-class="{ 'is-required': mac === machine.macs[0] }">
+              <span data-ng-if="mac === machine.macs[0]">MAC Address</span>
```
sparkiegeek, line 94: perhaps I'm misunderstanding this, but won't this make an empty <label> for the second and subsequent macs?

also instead of checking for === the first machine.macs can't we just use `$first`? (re: https://docs.angularjs.org/api/ng/directive/ngRepeat )

anthonydillon: I agree I don't think we should have this conditional in the label content. It looks like it just determines if the field should be required or not.

Caleb-Ellis: > perhaps I'm misunderstanding this, but won't this make an empty <label> for the second and subsequent macs?

That's exactly what it does!

Screenshot_2020-04-09 Machines bolla MAAS

> also instead of checking for === the first machine.macs can't we just use $first? (re: https://docs.angularjs.org/api/ng/directive/ngRepeat )

That's definitely the smarter way to do this. Generally with the angular pages I try to only change the things that are actually causing bugs (e.g. the input validation, not the labels), because often fixing/improving the current code leads to completely unrelated things breaking. This seems pretty innocuous though.

PR #966, opened 2020-04-09
Point all machine listing routes to the react machine listing.

squidsoup:
## Done
* Point all machine listing routes to the react machine listing.

## QA
* Run app, login. You should be redirected to `/r/machines` by default.
* Ensure links back to the machine listing are routed to the react app.


squidsoup:
> Should these be using the `REACT_BASENAME` environment variables?

They *could* be absolute paths, do you think that would be better?

PR #975, opened 2020-04-14
Prevent user from deleting pools that contain machines.

### LINKED ISSUE #918: Prevent user from deleting pools with machines in it
In the React resource pools table (`/r/pools`) you can attempt to delete a resource pool with machines in it, but nothing happens. The angular version allows you to attempt to delete the pool, but a warning shows up below the row: `Pool has machines in it, it cannot be deleted`. I think we can just disable the delete button with a tooltip `Cannot delete a pool that has machines in it` or something to that effect.

Ping @lilyvidenova @amylily1011 for thoughts on the wording/interaction.

Caleb-Ellis:
## Done
- Disabled delete button for pools that contain at least one machine.
- Show resourcepool errors in state in a notification on the pool list

## QA
- Go to `/r/pools` and check that resource pools with machines in them have disabled delete buttons that have a tooltip on hover
- Remove the `deleteDisabled` prop from `<TableActions>` in `Pools.js`
- Attempt to delete a pool that you otherwise couldn't and check that a an error notification shows up that you can clear

## Fixes
Fixes #918

### DIFF
--- a/ui/src/app/pools/views/Pools.js
+++ b/ui/src/app/pools/views/Pools.js
@@ -1,6 +1,12 @@
+import {
+  Col,
+  Loader,
+  MainTable,
+  Notification,
+  Row,
+} from "@canonical/react-components";
 import React, { useEffect, useState } from "react";
 import { useDispatch, useSelector } from "react-redux";
-import { Col, Loader, MainTable, Row } from "@canonical/react-components";
 import { Link } from "react-router-dom";
 
 import TableActions from "app/base/components/TableActions";
@@ -12,6 +18,7 @@ import {
 import { useAddMessage, useWindowTitle } from "app/base/hooks";
 import { resourcepool as resourcePoolSelectors } from "app/base/selectors";
 import { filtersToQueryString } from "app/machines/search";
+import { formatErrors } from "app/utils";
 
 const getMachinesLabel = (row) => {
   if (row.machine_total_count === 0) {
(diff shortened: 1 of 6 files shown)


ui/src/app/pools/views/Pools.js
```diff
@@ -105,51 +118,66 @@ const Pools = () => {
...
-        </div>
-      </Col>
-    </Row>
+    <>
+      {errorMessage ? (
+        <Row>
+          <Col size={12}>
+            <Notification
+              close={() => dispatch(resourcePoolActions.cleanup())}
+              timeout={5000}
```
anthonydillon, line 127: I don't think we should timeout a negative notification. I thought timeout notifications were just for informational and confirmation of user action?

ui/src/app/pools/views/Pools.js
```diff
@@ -105,51 +118,66 @@ const Pools = () => {
...
+      <Row>
+        <Col size={12}>
+          <div>
+            {poolsLoading && (
+              <div className="u-align--center">
+                <Loader text="Loading..." />
+              </div>
+            )}
+            {poolsLoaded && (
+              <MainTable
```
anthonydillon, line 144: Why not wrap the conditional around both sections?

squidsoup: they're different guards, `poolsLoading` vs `poolsLoaded`.

ui/src/app/pools/views/Pools.js
squidsoup: @anthonydillon a question for UX, I believe we do this in a number of places. @amylily1011 @lilyvidenova?

anthonydillon: We did have this guidelines when we originally created the notification component but that was a little while ago now.

Caleb-Ellis: I think you're right, we have temporary timeouts for the "information" notifications that just confirm what a user has done and the "negative" notifications persist until the user navigates away. I'll remove the timeout.

PR #983, opened 2020-04-16
Add warning to login page if no users detected.

### LINKED ISSUE #969: Login form should display warning to user that no admin exists
**Describe the bug**
Apparently this used to exist in the angular login form, but we seem to have lost it along the way.

**MAAS version**
? (try master)

**To Reproduce**
1. Login without an admin user

**Expected behavior**
This needs clarification (we should grep the code for the old warning).

Caleb-Ellis:
##  Done
- Added warning to login page if no users detected.

## QA
- Create a MAAS using the latest edge snap but don't create an admin
- Try to log in and check that a message shows up asking you to create an admin
- Click Retry and check that the page is refreshed as normal
- Create an admin using `sudo maas createadmin` and then click the Retry button again
- Check that the Login page shows up

## Fixes
Fixes #969 

## Launchpad fix
https://code.launchpad.net/~ack/maas/+git/maas/+merge/382269

## Screenshot
Screenshot_2020-04-16 Login MAAS

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -56,14 +56,15 @@ jobs:
           command: |
             sudo maas init \
               --maas-url=http://$IP_ADDRESS:5240/MAAS \
-              --admin-username=admin \
-              --admin-password=test \
-              --admin-email=fake@email.com \
               --mode region+rack \
               --database-host=localhost \
               --database-name=$MAAS_DBNAME \
               --database-user=$MAAS_DBUSER \
               --database-pass=$MAAS_DBPASS
+            sudo maas createadmin \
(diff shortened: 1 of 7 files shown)


Caleb-Ellis:
@amylily1011 @lilyvidenova I've added a "UX needed" label to check the styling/wording. I've just copied over what was there from the old code, so if you wanna improve it NOW'S THE TIME! :O

amylily1011:
I don't have anything to add. I think the message looks clear and it provides an action.  😄   :shipit:

lilyvidenova:
As long as the copy button *actually* works +1! :D

PR #997, opened 2020-04-17
Fix lock icon background-image to be a padlock.

### LINKED ISSUE #996: Lock icon is a grey power icon instead of a padlock
Definitely some bad copy+pasting from the angular code over to the react code. OOPS

Screenshot_2020-04-17 Machines bolla MAAS

Caleb-Ellis:
## Done
- Fixed bug where lock icon was a grey power icon instead of a padlock.
- Ran padlock svg through svgo and converted to background-image css that doesn't rely on the ubuntu asset server

## QA
- Lock a machine and check that a padlock icon shows up next to it

## Fixes
Fixes #996

Screenshot_2020-04-17 Machines bolla MAAS(1)

### DIFF
--- a/ui/src/scss/_patterns_icons.scss
+++ b/ui/src/scss/_patterns_icons.scss
@@ -22,6 +22,10 @@
     )+"'%3E%3C/path%3E%3C/g%3E%3C/g%3E%3C/g%3E%3C/svg%3E");
 }
 
+@mixin maas-icon-lock($color) {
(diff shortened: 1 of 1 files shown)


ui/src/scss/_patterns_icons.scss
```diff
@@ -30,7 +30,7 @@
 
   .p-icon--locked {
     @extend %icon;
-    @include maas-icon-power($color-mid-dark);
+    background-image: url("data:image/svg+xml,%3Csvg xmlns='http://www.w3.org/2000/svg' height='16' width='12'%3E%3Cpath d='M6 0C3.793 0 2 1.793 2 4v3H0v9h12V7h-2V4c0-2.207-1.793-4-4-4zm0 1c1.67 0 3 1.33 3 3v3H3V4c0-1.67 1.33-3 3-3zm1 8.5V14H5v-4l2-.5z' fill='%23888' fill-rule='evenodd'/%3E%3C/svg%3E");
```
anthonydillon, line 33: I wonder if naming a icon mixin would be more clear and help make icons more manageable in the project. Would also be consistent with the use of the removed icon mixin.

Caleb-Ellis: Good call! I've added a `maas-icon-lock` mixin and included it here

PR #1007, opened 2020-04-20
WIP: Debounce machine filtering.

### LINKED ISSUE #1005: Debounce filter function

Caleb-Ellis:
## Done
- Debounce machine filter function so that it only runs if there's been at least a 500ms interval. In the meantime, a spinner is displayed.

## QA
- Type in the filter box and check that a spinner shows up immediately.
- Check that the machine list doesn't update until you stop typing for at least 500ms.

### DIFF
--- a/ui/src/app/machines/views/MachineList/MachineList.js
+++ b/ui/src/app/machines/views/MachineList/MachineList.js
@@ -2,6 +2,7 @@ import {
   Button,
   Col,
   Input,
+  Loader,
   MainTable,
   Notification,
   Row,
@@ -11,7 +12,7 @@ import {
 import { useDispatch, useSelector } from "react-redux";
 import { useStorageState } from "react-storage-hooks";
 import classNames from "classnames";
-import React, { useEffect, useMemo, useState } from "react";
+import React, { useEffect, useMemo, useRef, useState } from "react";
 import pluralize from "pluralize";
 
 import {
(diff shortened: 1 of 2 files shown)


huwshimi:
@Caleb-Ellis I'm not sure about hiding the list. I wonder if people will want to refer to the list as they type? In a funny way it makes the search feel slower as the spinner makes me think it's trying to load new data.

I wonder if @lilyvidenova or @amylily1011 have any thoughts.

Caleb-Ellis:
@huwshimi after having played around with both options I think I prefer the spinner method better. IMO the instant feedback of the spinner provides a sense that the user is in control of what's happening on the screen, otherwise it kinda feels like we've just added 500ms of lag with the debounce.

For example with your suggestion, if I add a filter from the dropdown, the text in the searchbox immediately becomes `status:(=Allocated)` but the list doesn't update until the 500ms debounce interval passes *and* the DOM re-rendering occurs. To the end user all they are experiencing is an unresponsive page for 500ms longer, but with the spinner there's an indication that something is happening even if the amount of time it takes for the list to be filtered is identical.

Maybe there's a middle ground? We could show a spinner in the searchbox (or somewhere) and keep the list?

squidsoup:
> Maybe there's a middle ground? We could show a spinner in the searchbox (or somewhere) and keep the list?

That would be worth experimenting with, hiding the list would be a step backwards.

squidsoup:
We want to debounce changing the input state, that means that there's a perceivable lag when you start typing in the input - the user expects to see their text appear immediately. We do want to debounce the actual filtering/searching. Does that make sense?

Caleb-Ellis:
@huwshimi @squidsoup I've updated the PR so that the list stays and a spinner shows up in the search box. It's incredibly hacky (absolutely position `<i>`) but I just want to get peoples opinions on whether it feels right. The previous commit shows how it works with the list disappearing for comparison.

lilyvidenova:
I really like the spinner - it is quite subtle but it is enough to capture your attention and to be reassuring that something is in fact happening.

Caleb-Ellis:
Superseded by #1019

PR #1009, opened 2020-04-21
Add deprecation notices to NotificationList.

### LINKED ISSUE #1008: Build deprecation messages
[Design](https://github.com/canonical-web-and-design/MAAS-design/issues/897)

Caleb-Ellis:
## Done
- Added actions, reducers and selectors for fetching deprecation notices.
- Display deprecation notices in state at the top of NotificationList.

## QA
- Run this branch on bolla and check that a deprecation notice shows up at the top of the section.

## Fixes
Fixes #1008 

## Screenshot
Screenshot_2020-04-21 Machines bolla MAAS

### DIFF
--- a/ui/src/app/base/components/DeprecationNotices/DeprecationNotices.js
+++ b/ui/src/app/base/components/DeprecationNotices/DeprecationNotices.js
@@ -1,0 +1,42 @@
+import { Notification } from "@canonical/react-components";
+import { useDispatch, useSelector } from "react-redux";
+import React, { useEffect } from "react";
+
+import { general as generalActions } from "app/base/actions";
+import { general as generalSelectors } from "app/base/selectors";
+
+const DeprecationList = () => {
+  const dispatch = useDispatch();
+  const deprecationNotices = useSelector(
+    generalSelectors.deprecationNotices.filterByMinorVersion
+  );
+
+  useEffect(() => {
+    dispatch(generalActions.fetchVersion());
+    dispatch(generalActions.fetchDeprecationNotices());
+  }, [dispatch]);
+
+  return (
+    <>
+      {deprecationNotices.map((notice) => (
+        <Notification
+          data-test={`deprecation-${notice.id}`}
+          key={notice.id}
+          type="caution"
+        >
+          {notice.description}
+          <br />
+          {notice.url && (
+            <a className="p-link--external" href={notice.url}>
+              {notice["link-text"]
+                ? `${notice["link-text"]}...`
+                : "Learn more..."}
+            </a>
+          )}
+        </Notification>
+      ))}
+    </>
+  );
+};
+
+export default DeprecationList;
--- a/ui/src/app/base/selectors/general/version.js
+++ b/ui/src/app/base/selectors/general/version.js
@@ -6,4 +6,13 @@ import { generateGeneralSelector } from "./utils";
 
 const version = generateGeneralSelector("version");
 
+version.minor = (state) => {
+  const { data } = state.general.version;
+  const splitVersion = data.split(".");
+  if (splitVersion[0] && splitVersion[1]) {
+    return `${splitVersion[0]}.${splitVersion[1]}`;
+  }
+  return "";
+};
+
 export default version;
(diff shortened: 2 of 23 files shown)


Caleb-Ellis:
@lilyvidenova @amylily1011 I've added UX needed to check a few things:
- The copy in this branch is different to [the design](https://github.com/canonical-web-and-design/MAAS-design/issues/897#issuecomment-616438523). This will need to be changed in the api. Also, I've made the link text generic here because the api doesn't differentiate between what should be normal text and what should be link text. Let @albertodonato know if you want to update the copy and to allow the link text to be specific.
- On which pages do we want to display the deprecation notices? Just the machine list? Or is it every page? I just want to know if I should be doing any work on the angular pages too.
- Do we want to be able to close deprecation notices?

lilyvidenova:
> @lilyvidenova @amylily1011 I've added UX needed to check a few things:
> 
> * The copy in this branch is different to [the design](https://github.com/canonical-web-and-design/MAAS-design/issues/897#issuecomment-616438523). This will need to be changed in the api. Also, I've made the link text generic here because the api doesn't differentiate between what should be normal text and what should be link text. Let @albertodonato know if you want to update the copy and to allow the link text to be specific.

He's updating it as we speak. He's leaving out the "..." at the end of the link, could we add it in on our end, please? 

> * On which pages do we want to display the deprecation notices? Just the machine list? Or is it every page? I just want to know if I should be doing any work on the angular pages too (which, as always with the angular stuff, could potentially take a while).

Just machine listing, please. 

> * Do we want to be able to close deprecation notices?

How much work would it be to add the ability to re-show the notification after a week if it's still there and the user has closed it?

albertodonato:
@Caleb-Ellis https://code.launchpad.net/~ack/maas/+git/maas/+merge/382675 adds the `"link-text"` field to the websocket result entries for the text to display.

squidsoup:
> How much work would it be to add the ability to re-show the notification after a week if it's still there and the user has closed it?

That's a question for the core team @lilyvidenova, so you may want to raise that with them (they might not see your query here).

Caleb-Ellis:
@lilyvidenova I've made the changes so that the link text will show + "..." (or fallback to "Learn more...") and so that it only shows up in the machine list. For now it's not closeable, but we could add that functionality in as a follow up, either in the form of an api method (ideal) or adding a field to localStorage on the front end (a bit hacky).

lilyvidenova:
After talking to engineers, we decided that for the moment we will keep the notification as a non-dismissible warning.

ui/src/app/base/selectors/general/version.js
```diff
@@ -6,4 +6,13 @@ import { generateGeneralSelector } from "./utils";
...
 
+version.minor = (state) => {
+  const { data } = state.general.version;
+  const splitVersion = data.split(".");
+  if (splitVersion[0] && splitVersion[1]) {
+    return `${splitVersion[0]}.${splitVersion[1]}`;
+  }
+  return "";
+};
+
```
squidsoup, line 17: This could use some tests.

ui/src/app/base/components/DeprecationNotices/DeprecationNotices.js
```diff
@@ -0,0 +1,42 @@
...
+  return (
+    <>
+      {deprecationNotices.map((notice) => (
+        <Notification
+          data-test={`deprecation-${notice.id}`}
+          key={notice.id}
+          type="caution"
+        >
+          {notice.description}
+          <br />
```
squidsoup, line 28: Is this a sin?

Caleb-Ellis: @lyubomir-popov -sanctioned!
https://github.com/canonical-web-and-design/MAAS-design/issues/897#issuecomment-616438523

lilyvidenova:
> A bit disappointed that this wasn't discussed with the team, I'm not sure this is a good solution. I would have preferred to see _one_ place where we handle server side notifications, including these.
> 
> That being said, it looks like the decision was already made, and this branch is fine.

Can you elaborate, Kit? I feel like I'm missing the issue that you would have liked discussed - we were looking for a quick solution for this exact deprecation message, if there is a bigger issue you'd like discussed, let's have a session next week? @amylily1011 @cristinadresch

PR #1011, opened 2020-04-21
Action menu progress

huwshimi:
## Done
- Enabled status to be shown in the action menu form.
- Update the pool menu to use the status.

## QA
- Open the machine list.
- Select some machines.
- In the Action menu select the option to set the pool.
- Either choose a pool or create a new one in the form.
- Save the form.
- You should see (maybe very quickly) the count of machines increase until they're all processed.
- The form should then close.

## Fixes
Fixes: https://github.com/canonical-web-and-design/maas-ui/issues/881.

### DIFF
--- a/ui/src/app/base/selectors/general/machineActions.js
+++ b/ui/src/app/base/selectors/general/machineActions.js
@@ -1,3 +1,5 @@
+import { createSelector } from "@reduxjs/toolkit";
+
 /**
  * Selector for all possible machine actions.
  */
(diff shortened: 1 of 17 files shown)


ui/src/app/base/selectors/general/machineActions.js
```diff
@@ -6,4 +8,22 @@ import { generateGeneralSelector } from "./utils";
...
+
+/**
+ * Get a machine action by name.
+ * @param {Object} state - The redux state.
+ * @param {String} name - The name of a machine action.
+ * @returns {Object} A machine action.
+ */
+machineActions.getByName = createSelector(
+  [machineActions.get, getParams],
+  (items, [name]) => items.find((item) => item.name === name)
```
Caleb-Ellis, line 26: `items` is too generic, I think `actions` or similar would work better. Also I'm finding this `getParams` pattern confusing and hard to read (though I guess the `createSelector` api is a little weird to begin with). I'm wondering if it's too early to abstract it out, or if we want to abstract it out at all. Is it possible for it to be something like:
``` js
machineActions.getByName = createSelector(
  [machineActions.get, (state, name) => name],
  (actions, name) => actions.find((action) => action.name === name)
);
```

Actually does it even need to be its own selector to begin with? Couldn't you just use `machineActions.get ()`then `.find()` the action in the component? I'm not actually sure if there are any benefits to using selectors if they have dynamic arguments.

huwshimi: The benefit of doing it this way is we can use `createSelector` and it'll memoize. Unfortunately we have to do it this way because we don't use props to get the data into the components so the selectors can't use those for params.

PR #1035, opened 2020-04-24
Fix Loader in paragraph error

huwshimi:
## Done
- Upgrade React Components.
- Replace Loader with Spinner.

## QA
- Visit /r/machines.
- Select some machines.
- Using the take action menu select the option to change the pool.
- Select a pool (or create a new one).
- A spinner should appear with the status of the machines.
- In your console there shouldn't be an error about a `div` not being allowed in a `p`.

### DIFF
--- a/ui/package.json
+++ b/ui/package.json
@@ -4,7 +4,7 @@
   "private": true,
   "dependencies": {
     "@canonical/macaroon-bakery": "0.3.0",
-    "@canonical/react-components": "0.6.0",
+    "@canonical/react-components": "0.7.0",
     "@maas-ui/shared": "0.1.0",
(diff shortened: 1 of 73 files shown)


anthonydillon:
Code looks good. I am not seeing a spinner when changing a pool. It takes about 5 seconds and then the pool changes in the list. Am I missing something?

anthonydillon:
I'm seeing a lot of warnings in the console though:
```
ImmutableStateInvariantMiddleware took 103ms, which is more than the warning threshold of 32ms. 
```
Separate from this PR but the selection of the machines is taking a long time.

anthonydillon:
Sorry missed the clean and rebuild step. I see it now :+1:

PR #1040, opened 2020-04-27
Determine chassis power types from api.

### LINKED ISSUE #1012: Chassis power types should be dynamically listed
All the chassis power types are listed in ui/src/app/machines/views/AddChassis/chassisPowerTypes.js. If we add a new chassis power types, we have to change the backend and the ui. A better way is to use the settings not on scope SETTING_SCOPE.NODE registered in each driver and available in /MAAS/api/2.0/machines/?op=power_parameters

Caleb-Ellis:
## Done
- Replaced hardcoded list of chassis power types with dynamically filtered list from api.

## QA
- Go to `/r/machines/add` and check that all the power types can be selected from the dropdown.
- Go to `/r/machines/chassis/add` and check that the list of power types is filtered by `chassis: true` according to `general.powerTypes.data` in state.

## Fixes
Fixes #1012


ui/src/app/base/selectors/general/powerTypes.js
```diff
@@ -2,8 +2,14 @@
  * Selector for power types.
  */
 
+import { createSelector } from "@reduxjs/toolkit";
+
 import { generateGeneralSelector } from "./utils";
 
 const powerTypes = generateGeneralSelector("powerTypes");
 
+powerTypes.chassis = createSelector([powerTypes.get], (powerTypes) =>
```
huwshimi, line 11: I think we usually add doc strings for these selectors.

PR #1047, opened 2020-04-28
Memoise MachineListTable

squidsoup:
## Done
Use `React.memo` with `MachineListTable`.

With 105 machines, this saves ~500ms of rendering time in development mode.

## QA
`MachineListTable` should not render twice when switching from Pools to Machines.

### DIFF
--- a/ui/src/app/machines/views/MachineList/MachineListTable/MachineListTable.js
+++ b/ui/src/app/machines/views/MachineList/MachineListTable/MachineListTable.js
@@ -406,334 +406,337 @@ const generateGroupRows = ({
   return rows;
 };
 
-const MachineListTable = React.memo(
-  ({ filter, grouping, hiddenGroups, setHiddenGroups }) => {
-    const dispatch = useDispatch();
-    const selectedMachines = useSelector(machineSelectors.selected);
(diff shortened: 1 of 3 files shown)


ui/src/app/machines/views/MachineList/MachineListTable/MachineListTable.js
```diff
@@ -406,337 +406,334 @@ const generateGroupRows = ({
...
-  hiddenGroups,
-  setHiddenGroups,
-}) => {
-  const dispatch = useDispatch();
-  const selectedMachines = useSelector(machineSelectors.selected);
-  const selectedIDs = useSelector(machineSelectors.selectedIDs);
-  const machines = useSelector((state) =>
-    machineSelectors.search(state, filter, selectedIDs)
-  );
+const MachineListTable = React.memo(
```
huwshimi, line 409: The way I've seen this done is to instead change the export below, something like:
```js
export default React.memo(MachineListTable);
```

The docs show both ways: https://reactjs.org/docs/react-api.html#reactmemo

I think one of the benefits of applying the memo to the default export is that we can still export the un-memoized component for testing if need be.

PR #1048, opened 2020-04-28
Fix KVM checkbox when switching from other OS/Release to Ubuntu 18.04.

Caleb-Ellis:
## Done
- Fixed KVM checkbox being disabled when switching from any other OS/Release back to Ubuntu 18.04 if 18.04 is the first release in state.

## QA
- Make sure Ubuntu 18.04 is the earliest release available and you have at least 2 releases that aren't Ubuntu (bolla is set up this way).
- Select deployable machines and open the deploy form
- Change selection to Ubuntu 18.04 if not the default, and check that the KVM checkbox is enabled.
- Change the OS and release then change back to Ubuntu
- Check that the checkbox is enabled

## Fixes
Fixes canonical-web-and-design/MAAS-squad#1904


ui/src/app/machines/components/HeaderStrip/ActionFormWrapper/DeployForm/DeployFormFields/DeployFormFields.js
```diff
@@ -25,6 +26,15 @@ export const DeployFormFields = () => {
   const canBeKVMHost =
     values.oSystem === "ubuntu" && values.release === "bionic";
 
+  // Whenever the selected OS changes, set the release value to the first OS's
+  // release in state.
+  const previousOS = usePrevious(values.oSystem);
+  useEffect(() => {
```
huwshimi, line 32: I can't quite figure out why this is in a `useEffect` and not on the field's `onChange`. Is there a reason? If so, maybe a comment would be useful here for future developers (probably me) wondering the same thing.

Caleb-Ellis: Basically because the list of releaseOptions is determined by the hook:

``` js
const releaseOptions = useSelector((state) =>
  generalSelectors.osInfo.getOsReleases(state, values.oSystem)
);
```
And since you can't have hooks in callbacks I've moved the logic up into the main component. I guess I could just have get all releaseOptions and then filter based on values.oSystem *outside* of the hook instead though.

PR #1050, opened 2020-04-28
Add tooltips to truncated content

huwshimi:
## Done
- Add tooltips to the cell content in the machine list.

## QA
- View the machine list.
- Resize the window to different sizes and look for content that gets truncated.
- Check that the top and bottom rows of the cell content have tooltips .

## Fixes
Fixes: https://github.com/canonical-web-and-design/MAAS-squad/issues/1913.

### DIFF
--- a/ui/src/app/machines/views/MachineList/MachineListTable/NameColumn/NameColumn.js
+++ b/ui/src/app/machines/views/MachineList/MachineListTable/NameColumn/NameColumn.js
@@ -40,7 +40,10 @@ const generateIPAddresses = (machine) => {
 
   if (ipAddresses.length) {
     let ipAddressesLine = (
-      <span data-test="ip-addresses" title={ipAddresses[0]}>
+      <span
+        data-test="ip-addresses"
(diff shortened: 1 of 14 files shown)


ui/src/app/machines/views/MachineList/MachineListTable/NameColumn/NameColumn.js
```diff
@@ -40,7 +40,7 @@ const generateIPAddresses = (machine) => {
 
   if (ipAddresses.length) {
     let ipAddressesLine = (
-      <span data-test="ip-addresses">
+      <span data-test="ip-addresses" title={ipAddresses[0]}>
```
Caleb-Ellis, line 43: I think the title could be conditional on there being less than 2 IPs, otherwise you kinda double up with the browser and js tooltips:

screenshot

huwshimi: Thanks, fixed.

PR #1055, opened 2020-04-29
Add RSD form to Add hardware dropdown.

Caleb-Ellis:
## Done
- Added RSD form to Add hardware dropdown
- Added actions/reducers/selectors for pods

## QA
- Go to the machine list, click Add hardware then RSD and check that the form shows up
- Fill in the form, submit, and check that you get an error (unless you've done the setup required to get an RSD working... which you probably haven't).
- Check the network request is the same as if you'd done it at /#/rsd, *except* for `default_pool` which as far as I can tell from the api doesn't actually do anything because if a pool isn't sent the api assumes 0 anyway.

## Fixes
Fixes canonical-web-and-design/MAAS-squad#1901

## Screenshot
Screenshot_2020-04-29 Add pod bolla MAAS


ui/src/app/base/actions/pod/pod.js
```diff
@@ -0,0 +1,5 @@
+import { createStandardActions } from "app/utils/redux";
+
+const pod = createStandardActions("pod");
```
huwshimi, line 3: So nice to be able to do this!

ui/src/app/machines/components/HeaderStrip/AddHardwareMenu/AddHardwareMenu.js
```diff
@@ -1,47 +1,30 @@
...
-      to: "/machines/add",
-    },
-    {
-      children: "Chassis",
-      element: Link,
-      to: "/machines/chassis/add",
-    },
-  ];
-
-  return navigationOptions.rsd
```
huwshimi, line 23: Does this option become available after the RSD is set up using this form?

PR #1074, opened 2020-05-01
Fix processing state

huwshimi:
## Done
- Prevent the "update your selection" notice from appearing once machines have begun to be processed. This was causing a bug as when the machines begin to be changed they can change the allowable actions from that machine, resulting in that machine being shown.

## QA
- Selecting a number of new machines (at least 5).
- Using the take action menu select "Commission" and submit the form.
- The machines should transition to commissioning and the "update your selection" warning should not appear.
- You might have to do this a few times (you can abort once they transition to commissioning) to be satisfied this is fixed.

## Fixes
Fixes: https://github.com/canonical-web-and-design/MAAS-squad/issues/1897.

### DIFF
--- a/ui/src/app/machines/components/HeaderStrip/ActionFormWrapper/ActionFormWrapper.test.js
+++ b/ui/src/app/machines/components/HeaderStrip/ActionFormWrapper/ActionFormWrapper.test.js
@@ -1,4 +1,3 @@
-import { act } from "react-dom/test-utils";
 import { MemoryRouter } from "react-router-dom";
 import { mount } from "enzyme";
 import { Provider } from "react-redux";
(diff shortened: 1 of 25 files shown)


ui/src/app/machines/components/HeaderStrip/ActionFormWrapper/ActionFormWrapper.test.js
```diff
@@ -45,6 +83,35 @@ describe("ActionFormWrapper", () => {
...
+          <ActionFormWrapper
+            selectedAction={{ name: "commission" }}
+            setSelectedAction={jest.fn()}
+            _processing={true}
+          />
+        </MemoryRouter>
+      </Provider>,
+      { context: store }
+    );
+    // act(() => wrapper.find("CommissionForm").props().setProcessing(true));
```
Caleb-Ellis, line 109: Probably need to uncomment this

huwshimi: Whoops, thanks, removed.

PR #1086, opened 2020-05-05
Use createSelector for all selectors that can use it

huwshimi:
## Done
- Update all selectors that are fetching and manipulating state to use `createSelector`.
- This resulted in an up to about 50ms improvement on the first render of the machine list and went from 11 rerenders to 4.

### Before:
Screen Shot 2020-05-05 at 10 10 41 am

### After:
Screen Shot 2020-05-05 at 10 12 20 am

## QA
- These should all be internal changes to the selectors so the tests should pass, but you could poke around the machine list/prefs/settings to make sure everything works.

## Fixes
Fixes: https://github.com/canonical-web-and-design/maas-squad/issues/1872.

### DIFF
--- a/ui/src/app/base/selectors/user/user.js
+++ b/ui/src/app/base/selectors/user/user.js
@@ -1,17 +1,13 @@
+import { createSelector } from "@reduxjs/toolkit";
+
 const user = {};
 
 /**
  * Returns all users
  * @param {Object} state - The redux state.
- * @param {Number} batch - Number of users to return.
  * @returns {Array} A list of all users.
  */
-user.get = (state, batch) => {
-  if (batch) {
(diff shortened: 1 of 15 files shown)


ui/src/app/base/selectors/user/user.js
```diff
@@ -1,32 +1,28 @@
+import { createSelector } from "@reduxjs/toolkit";
+
 const user = {};
 
 /**
  * Returns all users
  * @param {Object} state - The redux state.
- * @param {Number} batch - Number of users to return.
```
huwshimi, line 6: This is the only select where there any api changes. We're not using this batch param in favour of handling the pagination using the table.

PR #1089, opened 2020-05-06
Change LXD KVM password input to type="password".

Caleb-Ellis:
## Done
- Changed LXD KVM password input to type="password".

## QA
- Go to /kvm and open the Add KVM form
- Select LXD type and check that the password input has `type="password"`

## Fixes
Fixes canonical-web-and-design/MAAS-squad#1920

## Launchpad Issue
[lp#1876921](https://bugs.launchpad.net/maas/+bug/1876921)


huwshimi:
@Caleb-Ellis the ipmi select boxes have double up/down arrows. It's because the selects are getting type="text" and should not have a type param. I wonder if you could fix that while you're modifying those field types?

Issue #1123, opened 2020-05-13
Fix styling between apps


anthonydillon:
The versions of Vanilla are different between legacy and modern so we need to either namespace them or align them.

PR #1124, opened 2020-05-13
Apply style namespacing to the ui views

anthonydillon:
## Done
This seemed the safest way of scoping the styling. If Vanilla is not namespaced we could get strange leakage by applying two versions of Vanilla to the legacy views. If we namepsace Vanilla we lose the HTML or root level styling. But that was the lesser evil in this case.

* Added a class to the HTML element when triggering a ui view
* Apply the `ui-view` style namespacing to the ui views
* Edited some parent-child inheritance in the local patterns
* Ported the HTML element styling from Vanilla directly into the namespace for now

## QA
* Ensure you're on the VPN. 
* Pull down this branch
* Run `yarn clean` 
* Then ` cd proxy`
* Followed by `yarn` then `yarn start`
* When the browser tab loads change the URL to http://0.0.0.0:8400/MAAS/r/machines
* Open the inspector and see there is the `ui-view` class on the `html` element.
* See that the machine list looks as it does [on bolla](http://bolla.internal:5240/MAAS/r/machines)
_If you don't have access to bolla check you are on the VPN. If you are please ask to be added on #maas_
* Click on Subnets to enter a angular view
* In the inspector check the `ui-view` class has been removed
* Check that this view matches the view on bolla
* Refresh the page and see the page looks the same
* Click on Settings to switch to a React view
* In the inspector see that the class has been added again
* Check the view looks the same as bolla
* Click through the pages and menus and check there is nothing out of place 

## Fixes
Fixes https://github.com/canonical-web-and-design/maas-ui/issues/1123

## Launchpad Issue
N/A

### DIFF
--- a/legacy/src/app/routes.js
+++ b/legacy/src/app/routes.js
@@ -21,6 +21,9 @@ import zonesListTmpl from "./partials/zones-list.html";
 
 /* @ngInject */
 const configureRoutes = ($stateProvider, $urlRouterProvider) => {
+  // Remove class on the HTML element when initialed to aid styling separation.
+  document.getElementsByTagName('html')[0].classList.remove("ui-view");
+
   $stateProvider
     .state("master", {
       abstract: true,
--- a/ui/src/index.js
+++ b/ui/src/index.js
@@ -38,6 +38,9 @@ export const store = configureStore({
 sagaMiddleware.run(rootSaga);
 
 const Root = () => {
+  // Apply class on the HTML element when initialed to aid styling separation.
+  document.getElementsByTagName("html")[0].classList.add("ui-view");
+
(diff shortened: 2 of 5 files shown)


anthonydillon:
You are going to need to namespace even if you upgrade legacy. Having both sets of styles applied to both applications is going to lead to issues that will be very difficult to spot in review.

The only issues that could arise is the use of `& &` in Vanilla. There is one case of it with `.row .row` but it's very unlikely to occur again in the time we still have the legacy app. Also, changes to the styling for the HTML element which there hasn't been for years.

I don't think spending any time on upgrading Vanilla in the legacy app is worthwhile as we are in the process of replacing it and could cause issues this close to release.

Caleb-Ellis:
@anthonydillon is it alright to close this one?

anthonydillon:
Yeah sure

PR #1141, opened 2020-05-19
Fix interface constraint construction in VM host compose form.

### LINKED ISSUE #1107: UI does not send interface constraints when composing a machine
I tried to compose a machine with two interfaces using the UI. The request went through with no errors however the region never received any interface constraints. Below is what was sent over the websocket. As you can see "interfaces" is set to ""

Note this effects all Pod types and this bug may exist previous versions of MAAS.

{
  "type": 0,
  "request_id": 24,
  "method": "pod.compose",
  "params": {
    "storage": "0:4(maas),1:4(maas)",
    "requests": [
      {
        "poolId": "e9c55659-125f-4b50-abf7-7ed631b3602c",
        "size": 8,
        "available": 11
      }
    ],
    "interfaces": "",
    "domain": 0,
    "zone": 1,
    "cores": "2",
    "memory": "2048",
    "id": 78
  }
}

https://bugs.launchpad.net/maas/+bug/1876180

Caleb-Ellis:
## Done
- Fixed interface constraint logic in VM host compose form.
- Slightly updated subnet list styling to be less space-y.

## QA
- Go to the KVM details page of a KVM that has attached VLANs (e.g. bolla LXD) and open the machine compose dropdown.
- Confirm that you can still compose a machine with default interface constraints. Check the websocket request has `params.interfaces : ""`.
- Open the compose dropdown again and add a couple of custom interface constraints. Check that the first available subnet is already selected by default, and the IP address field has a placeholder saying "Leave empty to auto-assign"
- For the first interface just select a subnet, and for the second select a subnet and enter an IP address.
- Submit the form and check that the websocket request correctly adheres to [the api](https://maas.io/docs/api#pod). You might get some cryptic errors if you input an IP address that is not part of the selected subnet.

## Fixes
Fixes #1107 

## Launchpad Issue
[lp#1876180](https://bugs.launchpad.net/maas/+bug/1876180)

## Screenshot
Screenshot_2020-05-19 MAAS UI

### DIFF
--- a/legacy/src/app/controllers/pod_details.js
+++ b/legacy/src/app/controllers/pod_details.js
@@ -441,19 +441,24 @@ function PodDetailsController(
 
     // Create the interface constraint.
     // <interface-name>:<key>=<value>[,<key>=<value>];...
-    var interfaces = [];
-    angular.forEach($scope.compose.obj.interfaces, function(iface) {
-      let row = "";
-      if (iface.ipaddress) {
-        row = `${iface.name}:ip=${iface.ipaddress}`;
-      } else if (iface.subnet) {
-        row = `${iface.name}:subnet_cidr=${iface.subnet.cidr}`;
-      }
-      if (row) {
-        interfaces.push(row);
-      }
-    });
-    params.interfaces = interfaces.join(";");
+    const interfaceConstraints = $scope.compose.obj.interfaces
+      .filter((iface) => iface.ipaddress || iface.subnet)
+      .map((iface) => {
+        const constraints = [];
+        if (iface.ipaddress) {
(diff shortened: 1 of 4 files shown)


Caleb-Ellis:
@cassiocassio @lilyvidenova I've added UX needed and tagging you here because from the code it looks like fields were being deliberately disabled depending on whether an ip address was typed or a subnet was selected.

This PR makes it so the first subnet is automatically selected so I've had to remove the disable-y work -- now you can enter both, but you'll get an error from the api if you do it wrong.

Caleb-Ellis:
Removing UX needed - this can be considered in the overall UX review and I'd like to merge it so I can handle conflicts with another branch

PR #1147, opened 2020-05-21
Autofill VM compose form with defaults + add overcommit sliders for LXD hosts.

### LINKED ISSUE #1144: UI should show default VM composition values
https://bugs.launchpad.net/maas/+bug/1879772

### LINKED ISSUE #1145: UI does not show overcommit configuration for LXD Pods
https://bugs.launchpad.net/maas/+bug/1879774

Caleb-Ellis:
## Done
- Backend now provides core, memory and storage defaults for pod types. Now they're being used in the VM compose form - the form is autofilled with defaults, and clearing them displays the default as a placeholder.
- Added overcommit sliders for LXD hosts.
- Tidied up websocket request preprocessor. It was originally sending a bunch of irrelevant garbage - now it's a bit more explicit what's actually being sent.
- Changed memory help text from GiB to MiB - it seemed weird for the form value (and the backend) to be in MiB and to have to calculate from MiB to GiB and back, when we could just use MiB.

## QA
- **Note:** Something has happened with bolla and now I can't seem to compose machines or edit configuration for either the LXD or virsh host - the requests just hang until they time out. This happens even on the production build. I think the best we can do for QA in the meantime is make sure the websocket request is correct.
- Go to a LXD pod details page and check that there are two sliders for CPU and memory overcommit. Submit the form and check that the websocket request includes `cpu_over_commit_ratio` and `memory_over_commit_ratio` and the numbers are correct.
- Open the compose form and check that cores, memory and storage are autofilled.
- Clear the cores, memory and storage fields and check that the placeholder displays the default
- Submit the form and check that the websocket request includes the defaults (even though the fields were empty) and there are only parameters that are allowed by [the api](https://maas.io/docs/api#pod).

## Fixes
Fixes #1144 
Fixes #1145 

## Launchpad Issue
[lp#1879772](https://bugs.launchpad.net/maas/+bug/1879772)
[lp#1879774](https://bugs.launchpad.net/maas/+bug/1879774)

### DIFF
--- a/legacy/src/app/controllers/pod_details.js
+++ b/legacy/src/app/controllers/pod_details.js
@@ -266,6 +266,17 @@ function PodDetailsController(
     iface.showOptions = true;
   };
 
+  $scope.validateRequest = (requested, available) => {
+    // Empty implies default, so is valid.
+    if (requested === "") {
+      return true;
+    }
+    if ((requested && isNaN(requested)) || (available && isNaN(available))) {
+      return false;
+    }
+    return Number(requested) > 0 && Number(available) >= Number(requested);
+  };
+
   $scope.validateMachineCompose = () => {
     const { availableWithOvercommit, compose, pod } = $scope;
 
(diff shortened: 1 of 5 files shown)


squidsoup:
QA is good @Caleb-Ellis, but could we fix the disabled styles for overcommit?

image

PR #1150, opened 2020-05-24
Use html5mode for routeProvider.

squidsoup:
## Done
Use html5mode route provider. This both fixes the production singlespa legacy bootstrapping issue, and is just generally a more contemporary approach to routing!

Legacy routes are now in the form `/MAAS/l/<route>` and the `#` routes are gone entirely.

## QA
* Ensure the app loads, and navigating between apps works as expected.


legacy/src/app/controllers/tests/test_nodes_list.js
```diff
@@ -277,7 +277,7 @@ describe("NodesListController", function() {
     expect($scope.devices).toBe(DevicesManager.getItems());
     expect($scope.pools).toBe(ResourcePoolsManager.getItems());
     expect($scope.controllers).toBe(ControllersManager.getItems());
-    expect($scope.osinfo).toBe(GeneralManager.getData("osinfo"));
+    expect($scope.osinfo).toStrictEqual(GeneralManager.getData("osinfo"));
```
squidsoup, line 280: Why this test is now failing is a bit of a mystery...

legacy/src/app/routes.js
```diff
@@ -19,6 +19,9 @@ import vlanDetailsTmpl from "./partials/vlan-details.html";
 import zoneDetailsTmpl from "./partials/zone-details.html";
 import zonesListTmpl from "./partials/zones-list.html";
 
+const prefixRoute = (route) =>
```
squidsoup, line 22: Unfortunately, afaict, angularjs provides no way to programatically specifiy a basepath when using the locationProvider's html5mode (it expects to be defined in `index.html` which is not possible in the singlespa context, which is unfortunately why this code exists.

PR #1158, opened 2020-05-27
Allow root prod build to configure PUBLIC_PATH from env.

squidsoup:
## Done
Allow root prod build to configure PUBLIC_PATH from env. This is required as maas-ui is served in two different contexts in MAAS (twisted/django, and from nginx in the snap).

## QA
Run `PUBLIC_PATH='/WHATEVS/' yarn build-all`. Ensure root builds with the specified public path (checking root/dist/index.html is easiest).

### DIFF
--- a/root/webpack.prod.js
+++ b/root/webpack.prod.js
@@ -4,7 +4,7 @@ const merge = require("webpack-merge");
 
 const common = require('./webpack.common.js');
 
-const publicPath = '/MAAS/r/';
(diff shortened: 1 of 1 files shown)


squidsoup:
This may not be necessary for the snap build after all, so putting on this on hold until my LP branch lands.

squidsoup:
Closing as this wasn't necessary.

Issue #1174, opened 2020-05-28
The Take action menu has deteriorated with the last update.


amylily1011:
On version 2.8.0 ~ beta4 (8502-g.942cfc7e9). The take action menu has deteriorated with the last update.
See the attached part.
Thank's.

From: https://bugs.launchpad.net/maas/+bug/1881143

![alt text](https://launchpadlibrarian.net/481790043/Screenshot_2020-05-28%20Machines%20maas-home-lab%20MAAS.png)

lilyvidenova:
Does this happen to any other of the menus?

amylily1011:
Only `take action` and `add hardware` on the first page.

`take action` menu on other pages are fine.

lilyvidenova:
![image](https://user-images.githubusercontent.com/17980375/83167045-a5febe80-a107-11ea-8d33-4703020d5f32.png)

amylily1011:
Looks like all dropdown menus on the first page is affected.

Issue #1172, opened 2020-05-28
Broken CSS in take action dropdown


squidsoup:
**Describe the bug**
We seem to have some wonky css in the take action dropdowns in react now.

**maas-ui version**
@maas-ui/maas-ui 1.1.1 (db20378).

**To Reproduce**
1. Select a machine on the machine listing
2. Click take action

**Screenshots**
![image](https://user-images.githubusercontent.com/130286/83128141-18cf6000-a12f-11ea-83a5-1aff59b5c1c3.png)

PR #1176, opened 2020-05-28
Copy the contextual styling from legacy to ui

anthonydillon:
## Done
Copy the contextual styling from legacy to ui.

## QA
- Join VPN
- Run `dotrun`
- Go to http://0.0.0.0:8400/MAAS/r/machines
- Check the Add hardware and Take action dropdowns look ok
- Check the in-table actions look ok

## Fixes
Fixes https://github.com/canonical-web-and-design/maas-ui/issues/1174
Fixes https://github.com/canonical-web-and-design/maas-ui/issues/1172

### DIFF
--- a/ui/src/scss/_patterns_cta.scss
+++ b/ui/src/scss/_patterns_cta.scss
@@ -1,0 +1,70 @@
+@mixin maas-p-cta {
+  %p-cta-icon {
+    background-position-y: center;
+    background-repeat: no-repeat;
+    background-size: map-get($icon-sizes, default);
+    content: "";
+    height: map-get($icon-sizes, default);
+    position: absolute;
+    width: map-get($icon-sizes, default);
+  }
+
+  .p-cta {
+    @extend .p-contextual-menu;
+  }
+
+  .p-cta__toggle {
+    margin-bottom: 0;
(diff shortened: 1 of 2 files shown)


Caleb-Ellis:
This QAs fine but I think the fix is a false positive. The `p-cta` classes aren't used in the react app at all. The actual issue is the Vanilla classes are out of order, because it's being included at the root level as well as each app. So the bug is occurring because the button classes are now occurring after the contextual menu classes, overwriting the overwrite, e.g. `menu` should override `button` but now `button` overrides `menu` overrides `button`.

I think what's happening here is that the styles copied over from legacy extend `.p-contextual-menu`, so since the mixin is being included after buttons it just so happens to put the vanilla styles back in the right order: `menu` overrides `button` overrides `menu` overrides `button`.

I have a fix that makes each app only includes some base styles and extend placeholders so we should keep the original Vanilla order: https://github.com/canonical-web-and-design/maas-ui/pull/1179

anthonydillon:
@Caleb-Ellis cool, I was going to switch to using the correct classnames but thought we wanted to get this issue fixed quickly before release. Your PR looks good.

PR #1177, opened 2020-05-28
Add hashroute redirect with machine listing fallthrough.

squidsoup:
## Done
Add redirect for legacy hash-based routes, with a fall through to the machine listing.

## QA
* Visit a deprecated route such as `/MAAS/#/controllers`, you should be redirected to `/MAAS/l/controllers`. 
* Visit an invalid route such as `/MAAS/l/foo`, you should be redirected to the machine listing.

### DIFF
--- a/legacy/src/app/routes.js
+++ b/legacy/src/app/routes.js
@@ -148,8 +148,14 @@ const configureRoutes = ($stateProvider, $urlRouterProvider) => {
       template: podDetailsTmpl,
       controller: "PodDetailsController",
     })
-    .state("pods", { url: prefixRoute("/pods"), redirectTo: prefixRoute("/kvm") })
-    .state("podDetails", { url: prefixRoute("/pod/:id"), redirectTo: prefixRoute("/kvm/:id") })
+    .state("pods", {
+      url: prefixRoute("/pods"),
+      redirectTo: prefixRoute("/kvm"),
+    })
+    .state("podDetails", {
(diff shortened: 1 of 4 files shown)


legacy/src/app/routes.js
```diff
@@ -148,8 +148,14 @@ const configureRoutes = ($stateProvider, $urlRouterProvider) => {
...
-    .state("pods", { url: prefixRoute("/pods"), redirectTo: prefixRoute("/kvm") })
-    .state("podDetails", { url: prefixRoute("/pod/:id"), redirectTo: prefixRoute("/kvm/:id") })
+    .state("pods", {
+      url: prefixRoute("/pods"),
+      redirectTo: prefixRoute("/kvm"),
+    })
+    .state("podDetails", {
+      url: prefixRoute("/pod/:id"),
+      redirectTo: prefixRoute("/kvm/:id"),
+    })
```
squidsoup, line 158: just reformatting here.

legacy/src/app/routes.js
```diff
@@ -227,6 +233,26 @@ const configureRoutes = ($stateProvider, $urlRouterProvider) => {
       template: dashboardTmpl,
       controller: "DashboardController",
     });
+
+  $urlRouterProvider.otherwise(($injector, $location) => {
```
squidsoup, line 237: unfortunately `$rootScope` isn't available in the angularjs `config` function, so I can't reuse caleb's nice navigation functions.

Issue #1185, opened 2020-06-01
UI stuck in error state after error loading machine


amylily1011:
From https://bugs.launchpad.net/maas/+bug/1881555

When attempting to view a machine that either doesn't exist, or that the current user doesn't have access to, the UI displays an error message with an action to 'Reload'.

All interactions within the (legacy) AngularJS app from that point refuse to update the UI. If one happens to hit a React route, then because the app switches, it behaves as one might expect.

Steps to reproduce:
1. http://bolla.internal:5240/MAAS/l/machine/this-does-not-exist
2. Now try to navigate to any Angular routes (e.g. Devices, Controllers, KVM, Images, DNS, AZs)
3. Note the browser doesn't move, and you're stuck with this error message

Expected
1. Nice 'Machine not found' page, offer link back to Machines list
2. No button to reload (if in prod, fine for dev)

MAAS: 2.8.0~beta4 (8512-g.1897d06c8)

PR #1194, opened 2020-06-03
Hide the error overlay when navigating away

huwshimi:
## Done
- Hide the error overlay when navigating away.

## QA
- Visit an entity url that doesn't exist, e.g. /MAAS/l/device/bob.
- You should see an error screen.
- In the header click on another angular route.
- The error screen should close and you should see the page you clicked on.

## Fixes
Fixes: https://github.com/canonical-web-and-design/maas-ui/issues/1185.

## Launchpad Issue
https://bugs.launchpad.net/maas/+bug/1881555

### DIFF
--- a/legacy/src/app/services/error.js
+++ b/legacy/src/app/services/error.js
@@ -17,6 +17,11 @@ function ErrorService() {
       this._error = error;
     }
   };
+
+  // Clear the current error.
+  this.clearError = function() {
+    this._error = null;
+  };
 }
 
 export default ErrorService;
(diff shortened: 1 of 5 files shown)


legacy/src/app/services/error.js
```diff
@@ -17,6 +17,12 @@ function ErrorService() {
       this._error = error;
     }
   };
+
+  // Clear the current error.
+  this.clearError = function() {
+    console.log('clearError');
```
sparkiegeek, line 23: stray `console.log`?
```suggestion
```

huwshimi: Oops! Thanks, fixed.

PR #1208, opened 2020-06-04
scripts history order is descending now.

xtrusia:
scripts history order is descending now.
so slice(0) access latest one

now, it shows always the oldest one after script created. even if script is updated with cli

## Done
* Itemised list of what was changed by this PR.

## QA
* Steps for QA.

## Fixes
Fixes # .

## Launchpad Issue
Related Launchpad maas issue in the form `lp#1879986`.


squidsoup:
Thanks for catching this @xtrusia! I've updated your PR to simply be:

```
const lastHistory = script.history[0];
```

as there's no need for destructuring now that the most recent script history is always at index 0. Thanks for the contribution!

PR #1209, opened 2020-06-04
Only log maas-ui version once.

### LINKED ISSUE #1162: Version string logged to console every time app mounts
**Describe the bug**
Given the app mounts and dismounts repeatedly with singlespa, logging the version should only occur once. I think there's a singlespa event for this we can use.

**maas-ui version**
master

**MAAS version**
n/a

squidsoup:
## Done
* Add GIT_SHA envvar to root app, and only log version once on `single-spa:first-mount`.

## QA
* Run app from either a react or angular route, ensure app version is only logged once.

## Fixes
Fixes #1162

### DIFF
--- a/root/src/root-application.js
+++ b/root/src/root-application.js
@@ -1,4 +1,5 @@
 import { registerApplication, start } from "single-spa";
+import { name as appName, version as appVersion } from "../../ui/package.json";
 
 import "./scss/base.scss";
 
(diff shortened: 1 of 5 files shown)


root/src/root-application.js
```diff
@@ -9,6 +10,10 @@ const showLoading = () => {
   }
 };
 
+window.addEventListener("single-spa:first-mount", () => {
+  console.info(`${appName} ${appVersion} (${process.env.GIT_SHA}).`);
```
huwshimi, line 14: This `console.info` could just be inside this module and will be printed when the module is initialised as it doesn't actually need to wait for the mount to happen, but inside the listener is fine too.

Issue #1211, opened 2020-06-05
Machine listing got messed up


amylily1011:
From: https://bugs.launchpad.net/maas-ui/+bug/1881275

This is with MAAS 2.8.0~beta4 (8510-g.f5a42ccea)

I'm not sure what happened, but the machine listing got in a weird state:

  https://private-fileshare.canonical.com/~bjorn/too-many-machines.png

There should be 19 machines and 2 resource pools, but it instead showed
57 machines and 4 resource pools.

This happened on two browser tabs that both had the machine listing open.
I'm not sure what triggered it. But I see that the MAAS snap was refreshed
and thus caused a restart to happen, so it might be related to that.

In the logs I see that MAAS was updated from revision 6683 to 6723.

After debugging, it seems that the UI is not refreshing its state when it loses connection to the server and reconnects again.

huwshimi:
Fixed by https://github.com/canonical-web-and-design/maas-ui/pull/1234

PR #1214, opened 2020-06-08
Update state entity on CREATE_<MODEL>_NOTIFY if exists.

squidsoup:
## Done
In the event that the server erroneously attempts to create an entity that already exists in state (e.g. a machine), ensure that the entity is updated in state instead.

## QA
Point maas-ui at a local maas snap. Restart the maas service (`sudo snap restart maas.supervisor`), wait a bit and you should not see duplicate machines in the machine list.

## Fixes
Relevant issue #1211 (will not close until websocket fix has landed).

## Launchpad Issue
[lp#1881275](https://bugs.launchpad.net/maas-ui/+bug/1881275) (note, while this change should effectively fix this bug, the UI still needs to refresh state when the websocket disconnects which will be addressed in a later branch).

### DIFF
--- a/ui/src/app/base/reducers/machine/machine.js
+++ b/ui/src/app/base/reducers/machine/machine.js
@@ -150,8 +150,8 @@ const machine = createNextState(
         draft.saving = false;
         break;
       case "CREATE_MACHINE_NOTIFY":
-        // In the event that the server erroneously attempts to create an existing machine
-        // ensure we update instead.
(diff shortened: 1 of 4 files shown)


sparkiegeek:
[LP: #1881275 - Machine listing got messed up](https://trello.com/c/K6GEMyko/369-lp-1881275-machine-listing-got-messed-up)

PR #1231, opened 2020-06-10
Update images page to allow selecting architectures on a per-release basis.

### LINKED ISSUE #1159: MAAS "Images" page seems to imply that i386/20.04 is available, but it's not
https://bugs.launchpad.net/maas/+bug/1877585

Caleb-Ellis:
##  Done
- Updated images page to allow selecting architectures on a per-release basis. Originally, selecting releases and architectures would download every possible combination.
- Disable release/arch combinations that are unsupported.
- Rearranged markup in boot_images.html to make a bit more sense.

## QA
- Run a local MAAS using [this branch](https://code.launchpad.net/~ltrager/maas/+git/maas/+merge/385331), which updates the websocket handler to allow for per-release selections.
- Go to /MAAS/l/images and check that you can select architectures per release, and the table below displays the summary correctly.
- Check that 20.04/i386 is disabled with a tooltip.
- Click "Save selection" and check that the downloads start correctly.
- Check that once the downloads are complete you can also remove them correctly.

## Fixes
Fixes #1159
Fixes canonical-web-and-design/maas-design#910

## Launchpad Issue
[lp#1877585](https://bugs.launchpad.net/maas/+bug/1877585)

## Screenshot
0 0 0 0_8400_MAAS_l_intro_user

### DIFF
--- a/legacy/src/app/directives/boot_images.js
+++ b/legacy/src/app/directives/boot_images.js
@@ -6,6 +6,7 @@
 import angular from "angular";
 
 import bootImagesTmpl from "../partials/boot-images.html";
+import { BootResourceType } from "../enum";
 
 /* @ngInject */
 export function maasBootImagesStatus(BootResourcesManager) {
@@ -60,9 +61,13 @@ export function maasBootImages(
 
   /* @ngInject */
   function BootImagesController($scope) {
+    const DEFAULT_RELEASE = "bionic";
+    const DEFAULT_ARCH = "amd64";
+
     $scope.loading = true;
     $scope.saving = false;
     $scope.saved = false;
+    $scope.selectedUbuntuRelease = "";
     $scope.stopping = false;
     $scope.design = $scope.design || "page";
     $scope.bootResources = BootResourcesManager.getData();
@@ -79,10 +84,12 @@ export function maasBootImages(
       keyring_data: "",
       releases: [],
       arches: [],
+      osystems: [],
       selections: {
         changed: false,
         releases: [],
         arches: [],
+        osystems: [],
       },
     };
     $scope.ubuntuCoreImages = [];
@@ -169,10 +176,12 @@ export function maasBootImages(
         }
         $scope.source.releases = [];
         $scope.source.arches = [];
+        $scope.source.osystems = [];
         $scope.source.selections = {
           changed: false,
           releases: [],
           arches: [],
+          osystems: [],
         };
       }
       $scope.updateSource();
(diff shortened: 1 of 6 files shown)


legacy/src/app/partials/intro-user.html
```diff
@@ -2,10 +2,10 @@
     <div class="row">
         <div class="p-card--highlighted">
             <div class="p-card__content u-no-margin--top">
-                <h2 class="p-card__title p-heading--four">
+                <h4 class="p-card__title">
```
Caleb-Ellis, line 5: Driveby fixes for icons in titles - icon only matches heading size according to the tag, not the class. The images partial is used in the images page (obviously) as well as the intro page, with the only differences being these heading have icons and the page is wrapped in a highlighted card instead of a plain div.

legacy/src/app/directives/boot_images.js
```diff
@@ -216,17 +221,51 @@ export function maasBootImages(
...
+    // objects that can be used in the save_ubuntu websocket handler.
+    $scope.convertResourcesToOSystems = (resources) => {
+      if (!resources) {
+        return [];
+      }
+      return resources.reduce((osystems, resource) => {
+        const split = resource.name.split("/");
+        const osName = split[0];
+        const osRelease = split[1];
+        if (resource.rtype === 0) {
```
squidsoup, line 246: Would be nice to have a resource type enum to make this more meaninful.

legacy/src/app/directives/boot_images.js
```diff
@@ -322,23 +366,79 @@ export function maasBootImages(
...
+    // Toggle the selection of the release/architecture combination.
+    $scope.toggleSelectedOS = (releaseName, archName) => {
+      const selectedOSs = $scope.source.selections.osystems;
+      let newSelected = [];
+      const osystemInSelected = selectedOSs.find(
+        (osystem) => osystem.release === releaseName
+      );
+      if (osystemInSelected) {
+        if (osystemInSelected.arches.includes(archName)) {
+          if (osystemInSelected.arches.length === 1) {
```
squidsoup, line 386: Perhaps consider factoring this out if you can, three levels of `if else` statements are a bit hard to read. If you moved this logic to a function outside of toggleSelectedOS, perhaps you could return `newSelected` early which might help with readability, e.g. something like:

```
if (!osystemInSelected) {
        return newSelected = [
          ...selectedOSs,
          { osystem: "ubuntu", release: releaseName, arches: [archName] },
        ];
}
.... etc
```

legacy/src/app/directives/boot_images.js
```diff
@@ -216,17 +221,50 @@ export function maasBootImages(
...
-      });
-      angular.forEach($scope.source.arches, function (arch) {
-        if (arch.name === "amd64") {
-          $scope.source.selections.arches.push(arch);
-        }
-      });
+    $scope.selectDefaults = () => {
+      const source = $scope.source;
+      if (
+        source.releases.some((release) => release.name === "bionic") &&
```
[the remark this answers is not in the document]
Caleb-Ellis: Done!

legacy/src/app/directives/boot_images.js
Caleb-Ellis: Done!

legacy/src/app/directives/boot_images.js
Caleb-Ellis: Good call - I think I was lost in the sauce a bit when writing it. I've factored out the newSelected part to its own function so it returns out earlier, and flattened the ifs.

PR #1232, opened 2020-06-10
Fix form card column sizes.

Caleb-Ellis:
## Done
- Fixed col sizes in form cards.
- Put col sizes into a constants file so we hopefully don't get regressions like this again. Ideally we should be using these constants wherever we have a Col component but for now I've just used them wherever a sidebar might be present.

## QA
- Go to settings, see that the add user form card and delete user expanded rows look good.

## Fixes
Fixes canonical-web-and-design/maas-design#907

## Screenshot
0 0 0 0_8400_MAAS_r_settings_users

Also updated DHCP snippets form:


squidsoup:
One potentially related issue, in contrast to the user form, the dhcp snippets form doesn't extend its fields to the end of the container:

image

Does that look right to you?

Issue #1240, opened 2020-06-11
CSS misstyled when accessing angular pages from redirect 


amylily1011:
From: https://bugs.launchpad.net/maas/+bug/1883132

MAAS version: 2.8.0~rc3 (8547-g.90d154c40)
@maas-ui/maas-ui 1.3.1 (7c2e4f4).

When accessing a page that ends up redirecting, the CSS can be missing/confused.

Example going from a 2.7 URL to a 2.8 RC3 one,

Visit http://karura.internal:5240/MAAS/#/kvm
Change the URL to test the redirect so it's http://bolla.internal:5240/MAAS/#/kvm
Note URL frenzy of redirects, and end up with a funkily styled page

## Steps to reproduce
- Be connected to the VPN
- Open this link in a new tab: http://bolla.internal:5240/MAAS/#/kvm
- See that when the page loads it looks like the image below
![image](https://user-images.githubusercontent.com/1413534/84443927-80f66980-ac38-11ea-91c9-60fdf0147580.png)

## Further redirect issue
- You URL should be: http://bolla.internal:5240/MAAS/l/kvm
- Change the `l` to `#` and you land at http://bolla.internal:5240/MAAS/r/machines

PR #1242, opened 2020-06-11
Handle styling to legacy redirects

anthonydillon:
## Done
* If both apps do not load then it is likely a legacy redirect to a legacy view, therefore, load the legacy styles as a fallback. 

## QA
* I couldn't replicate from the UI alone so Adam is QA'ing

## Fixes
Fixes https://github.com/canonical-web-and-design/maas-ui/issues/1240

## Launchpad Issue
[lp1883132](https://bugs.launchpad.net/maas/+bug/1883132)

### DIFF
--- a/root/src/root-application.js
+++ b/root/src/root-application.js
@@ -13,6 +13,7 @@ const showLoading = () => {
 console.info(`${appName} ${appVersion} (${process.env.GIT_SHA}).`);
 
 window.addEventListener("single-spa:before-app-change", (evt) => {
+
   const {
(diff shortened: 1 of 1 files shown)


sparkiegeek:
LGTM, confirmed unable to reproduce the original error with it built and installed on bolla

squidsoup:
I've marked this don't merge while investigating further. I'm not sure this is the right fix.

huwshimi:
Superseded by https://github.com/canonical-web-and-design/maas-ui/pull/1244

PR #1245, opened 2020-06-12
Fix subnets hightlight

huwshimi:
## Done
- Highlight the subnets nav item.

## QA
- Click on the subnets item in the main nav, it should stay highlighted.
- Click on other items, they should still highlight.

### DIFF
--- a/shared/src/components/Header/Header.js
+++ b/shared/src/components/Header/Header.js
@@ -65,7 +65,7 @@ export const Header = ({
       window.ga("set", "dimension2", uuid);
 
       const sendPageview = () => {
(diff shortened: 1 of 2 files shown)


shared/src/components/Header/Header.js
```diff
@@ -65,7 +65,7 @@ export const Header = ({
       window.ga("set", "dimension2", uuid);
 
       const sendPageview = () => {
-        const path = window.location.pathname + window.location.hash;
```
huwshimi, line 68: We use /l/ instead of /#/ so I've removed that as part of the url.

PR #1256, opened 2020-06-15
Break down KVM list into header, table and cell components.

Caleb-Ellis:
## Done
- Broke down KVMList component into logical sub-components:
  - KVMListHeader, which will include the action forms eventually,
  - KVMListTable, which is basically a prop'd MainTable,
    - Each cell is also split into its own component, like with the machine list.
- Power and OS cells just return "Unknown" for now. They will be handled as a separate issue because they require a bit of extra work to find the relevant host info.
- Adjusted the meter label padding so it matches default line break spacing.

## QA
- Go to /MAAS/r/kvm.
- Check that the cell data matches the data in /MAAS/l/kvm, except for Power and OS.

## Fixes
Fixes canonical-web-and-design/MAAS-squad#1951


ui/src/app/kvm/views/KVMList/KVMListTable/KVMListTable.js
```diff
@@ -0,0 +1,90 @@
...
+          headers={[
+            { content: "FQDN" },
+            { content: "Power" },
+            { content: "KVM Host Type" },
+            {
+              className: "u-align--right",
+              content: (
+                <>
+                  <span>
+                    VM<span style={{ textTransform: "none" }}>s</span>
```
huwshimi, line 61: Should this maybe be a utility class?

Caleb-Ellis: @huwshimi Yeah probably. I'll update it now

PR #1257, opened 2020-06-15
Migrate to TypeScript.

squidsoup:
## Done
* Add TypeScript compiler, targeting ES6.
* Add eslint overrides to allow for separate eslint configuration for js and ts files (this requires the explicit addition of libraries provided by CRA)
* Convert a simple utility function to TypeScript as a test (capitaliseFirst).

## QA
* Ensure app runs, and builds as expected.

This might be a good time to ensure your IDE works well with TypeScript and they you see inline errors, have intellisense etc.


package.json
```diff
@@ -13,7 +13,7 @@
     "cypress-open": "cd proxy && yarn cypress-open",
     "cypress-run": "cd proxy && yarn cypress-run",
     "copy-build": "mkdir -p build && yarn copy-root && yarn copy-legacy-assets && yarn copy-ui-css",
-    "copy-legacy-assets": "cp -Ru legacy/dist/assets build/",
+    "copy-legacy-assets": "cp -R legacy/dist/assets build/",
```
squidsoup, line 16: Driveby to allow the build to work on macos (`-u` is not a valid `cp` flag on macos). I think this should be fine to remove, but do let me know if it was needed for a particular reason.

ui/src/app/machines/components/HeaderStrip/ActionFormWrapper/OverrideTestForm/OverrideTestForm.js
```diff
@@ -187,7 +187,7 @@ export const OverrideTestForm = ({
                             );
                           }}
                         >
-                          Machine > Hardware tests
+                          Machine &gt; Hardware tests
```
squidsoup, line 190: The TypeScript compiler noticed that this should be an html entity.

legacy/.eslintrc.js
```diff
@@ -1,37 +1,42 @@
...
+    sourceType: "module",
+  },
+  plugins: ["react"],
+  rules: {
+    "no-unused-vars": [2, { args: "none" }],
+  },
+  settings: {
+    react: {
+      version: "detect",
     },
```
squidsoup, line 40: This is almost all formatting, I've added `settings: { react: { version: "detect" } }` here to suppress the annoying warnings.

PR #1259, opened 2020-06-16
Build KVM table Power and OS cells.

Caleb-Ellis:
## Done
- Built KVM table Power and OS cells.
- Converted the other cells (except for the Meters) to use the DoubleRow component to be consistent with the MachineListTable.
- Note that the secondary row text is slightly misaligned in the DoubleRow component - it should be exactly aligned with the Meter labels. I initially tried to fix it in this PR but it ended up breaking the MachineList rows, so I'll look into it as a follow up.
- Also note that making the table responsive will be handled in canonical-web-and-design/MAAS-squad#2025 so the OS text is likely to get truncated at the moment.

## QA
- Go to /MAAS/r/kvm
- Check that the Power and OS text matches what you see on /MAAS/l/kvm

## Fixes
Fixes canonical-web-and-design/MAAS-squad#2022

## Screenshot
0 0 0 0_8400_MAAS_r_kvm (4)

### DIFF
--- a/ui/src/app/base/selectors/pod/pod.test.js
+++ b/ui/src/app/base/selectors/pod/pod.test.js
@@ -77,4 +77,46 @@ describe("pod selectors", () => {
       id: 222,
     });
   });
+
+  it("can get a pod's host machine", () => {
+    const state = {
+      controller: {
+        items: [{ name: "fat-controller", system_id: "qwerty" }],
+      },
+      machine: {
+        items: [
+          { name: "mean-bean-machine", system_id: "abc123" },
+          { name: "lean-cuisine-machine", system_id: "def456" },
+        ],
+      },
+      pod: {
+        items: [{ host: "abc123", name: "pod-1" }],
+      },
+    };
+    expect(pod.getHost(state, { host: "abc123" })).toStrictEqual({
+      name: "mean-bean-machine",
+      system_id: "abc123",
+    });
+  });
+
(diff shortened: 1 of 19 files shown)


ui/src/app/base/selectors/pod/pod.test.js
```diff
@@ -77,4 +77,46 @@ describe("pod selectors", () => {
       id: 222,
     });
   });
+
+  it("can get a pod's host machine", () => {
+    const state = {
+      controller: {
+        items: [],
```
squidsoup, line 84: Would it be useful to add a controller here, so that the test also validates that a machine is returned by default?

ui/src/app/utils/getPowerIcon.js
```diff
@@ -0,0 +1,9 @@
+export const getPowerIcon = (machine, loading) => {
```
squidsoup, line 1: Some jsdoc would be nice.

ui/src/app/utils/getStatusText.js
```diff
@@ -0,0 +1,30 @@
+import { nodeStatus } from "app/base/enum";
+
+// Node statuses for which the OS + release is made human-readable.
+const formattedReleaseStatuses = [nodeStatus.DEPLOYED, nodeStatus.DEPLOYING];
+
+export const getStatusText = (machine, osReleases) => {
```
squidsoup, line 6: jsdoc again

PR #1263, opened 2020-06-17
Convert KVM React code to TypeScript

Caleb-Ellis:
## Done
- Converted all files under `app/kvm/` to TypeScript.
- Converted reducer and selector files that are relevant to KVM page to TypeScript (pod, resourcepool, zone).
- Added `app/base/types.ts` file which currently holds all the types used in more one place.

## QA
- See if you like the initial structure of the code.
- Check that the app runs.

Fixes #1263


ui/src/app/base/components/DoubleRow/DoubleRow.js
```diff
@@ -88,6 +88,7 @@ DoubleRow.propTypes = {
   className: PropTypes.string,
   icon: PropTypes.node,
   iconSpace: PropTypes.bool,
+  iconTitle: PropTypes.string,
```
Caleb-Ellis, line 91: TS picked up on a missing proptype here

ui/src/app/kvm/views/KVMList/KVMListTable/CPUColumn/CPUColumn.tsx
```diff
@@ -3,9 +3,14 @@ import { useSelector } from "react-redux";
...
 import { pod as podSelectors } from "app/base/selectors";
 import Meter from "app/base/components/Meter";
+import RootState from "app/base/reducers/types";
 
-const CPUColumn = ({ id }) => {
-  const pod = useSelector((state) => podSelectors.getById(state, id));
+type Props = { id: number };
+
+const CPUColumn = ({ id }: Props): JSX.Element => {
+  const pod = useSelector((state: RootState) =>
```
Caleb-Ellis, line 11: All of these selectors hooks were causing TS issues, namely `Expected 1 arguments, but got 2.` To fix it I've converted the selector files to valid TS, but maybe there's a more progressive way to do it.

ui/src/app/kvm/views/KVMList/KVMListTable/CPUColumn/CPUColumn.tsx
```diff
@@ -3,9 +3,14 @@ import { useSelector } from "react-redux";
...
+    podSelectors.getById(state, id)
```
Caleb-Ellis, line 12: All of these selectors hooks were causing TS issues, namely `Expected 1 arguments, but got 2.` To fix it I've converted the selector files to valid TS, but maybe there's a more progressive way to do it.

ui/src/app/base/reducers/pod/pod.test.ts
```diff
@@ -2,7 +2,7 @@ import pod from "./pod";
 
 describe("pod reducer", () => {
   it("should return the initial state", () => {
-    expect(pod(undefined, {})).toEqual({
+    expect(pod(undefined, { type: "" })).toEqual({
```
Caleb-Ellis, line 5: react-redux types complain if an action does not have a type parameter

ui/src/app/base/selectors/resourcepool/resourcepool.ts
```diff
@@ -1,58 +1,69 @@
...
  * @returns {Object} Resource pool that matches id.
  */
-resourcepool.getById = createSelector(
-  [resourcepool.all, (state, id) => id],
+const getById = createSelector(
+  [all, (_: RootState, id: number) => id],
   (pools, id) => pools.find((pool) => pool.id === id)
 );
 
+const resourcepool = {
```
squidsoup, line 59: Assigning the selectors to the exported object at the end is fine, but we should be consistent and add a card perhaps if we'd rather do it this way. It is less verbose, but it also means you can potentially forget to export a selector.

ui/src/app/kvm/views/KVMList/KVMListTable/CPUColumn/CPUColumn.tsx
squidsoup: No worries, there's actually less code here than I though there would be, 50+ files sounded scary last night :)

ui/src/app/kvm/views/KVMList/KVMListTable/KVMListTable.tsx
```diff
@@ -10,6 +10,7 @@ import {
   resourcepool as poolActions,
   zone as zoneActions,
 } from "app/base/actions";
+import { Pod } from "app/base/reducers/pod/types";
```
squidsoup, line 13: As before, this suggests `types` are in the wrong place I think.

ui/src/app/base/reducers/pod/types.ts
```diff
@@ -0,0 +1,51 @@
+export type PodHint = {
```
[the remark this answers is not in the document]
Caleb-Ellis: @squidsoup Yeah I was certain this was not the right place. Should I just lump them all in `app/base/types.ts` for now, then as a follow up try moving everything to `app/store/`?

ui/src/app/kvm/views/KVMList/KVMListTable/PoolColumn/PoolColumn.tsx
```diff
@@ -7,13 +7,18 @@ import {
   zone as zoneSelectors,
 } from "app/base/selectors";
 import DoubleRow from "app/base/components/DoubleRow";
+import RootState from "app/base/reducers/types";
 
-const PoolColumn = ({ id }) => {
-  const pod = useSelector((state) => podSelectors.getById(state, id));
-  const pool = useSelector((state) =>
+type PoolColumnProps = { id: number };
```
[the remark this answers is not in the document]
Caleb-Ellis: @squidsoup Yeah I think that's fine since adding the name of the component to the prop name is pretty redundant

ui/src/app/base/reducers/pod/types.ts
```diff
@@ -0,0 +1,51 @@
...
+  cpu_speed: number;
+  created: string;
+  default_macvlan_mode: string;
+  default_storage_pool: string;
+  hints: PodHint & {
+    cpu_speed: number;
+    iscsi_storage: number;
+    iscsi_storage_gb: string;
+    local_disks: number;
+  };
```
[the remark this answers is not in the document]
Caleb-Ellis: @squidsoup But then I have to come up with a good name for the type! D:

ui/src/app/base/reducers/pod/types.ts
```diff
@@ -0,0 +1,51 @@
...
+  id: number;
+  ip_address: number | string;
+  memory_over_commit_ratio: number;
+  name: string;
+  permissions: string[];
+  pool: number;
+  power_address: string;
+  power_pass: string;
+  owners_count: number;
+  storage_pools: unknown[];
```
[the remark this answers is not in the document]
Caleb-Ellis: Gotcha

ui/src/app/base/selectors/resourcepool/resourcepool.ts
Caleb-Ellis: The reason I've changed it is because TS was raising errors about trying to assign parameters to an empty object. I guess I could instead create the object in one go, so we don't have to worry about adding it to the export e.g.
``` typescript
const pod = {
  all: (state: RootState): Pod[] => state.pod.items,
  loaded: (state: RootState): boolean => state.pod.loaded,
  ...
};

File: ui/src/app/base/reducers/pod/types.ts (hunk 9 above)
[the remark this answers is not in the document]
Comment (squidsoup) [reply ui/src/app/base/reducers/pod/types.ts:24]: > @squidsoup But then I have to come up with a good name for the type! D:

Hah yes, I think that's half the battle.

File: ui/src/app/base/selectors/resourcepool/resourcepool.ts (hunk 5 above)
Comment (squidsoup) [reply ui/src/app/base/selectors/resourcepool/resourcepool.ts:59]: > The reason I've changed it is because TS was raising errors about trying to assign parameters to an empty object. I guess I could instead create the object in one go, so we don't have to worry about adding it to the export e.g.
> 
> ```ts
> const pod = {
>   all: (state: RootState): Pod[] => state.pod.items,
>   loaded: (state: RootState): boolean => state.pod.loaded,
>   ...
> };
> 
> export default pod;
> 

That looks good, let's do that!

File: ui/src/app/base/reducers/pod/types.ts (hunk 7 above)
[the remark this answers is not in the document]
Comment (squidsoup) [reply ui/src/app/base/reducers/pod/types.ts:1]: `app/base/types.ts` is probably okay for now, then we can figure out a more ideal project structure.

File: ui/src/app/base/selectors/resourcepool/resourcepool.ts (hunk 5 above)
Comment (Caleb-Ellis) [reply ui/src/app/base/selectors/resourcepool/resourcepool.ts:59]: ahh actually that won't work for any composed selectors because they relies on those methods actually existing first, nevermind

File: ui/src/app/base/selectors/pod/pod.test.js [hunk 11]
```diff
@@ -72,10 +72,6 @@ describe("pod selectors", () => {
       name: "podrick",
       id: 222,
     });
-    expect(pod.getById(state, "222")).toStrictEqual({
```
Caleb-Ellis, line 75: Removed this test because it was only useful when using `useParams` which returned the pod id as a string. The id type should always be number when running this function.

ui/src/app/base/selectors/resourcepool/resourcepool.ts
squidsoup: Right okay, this is fine, but we should remember to update the other selectors with this shape when we convert them to ts.

ui/src/app/base/types.ts
```diff
@@ -0,0 +1,104 @@
+export type TSFixMe = any; // eslint-disable-line @typescript-eslint/no-explicit-any
+
+export type ErrorState = TSFixMe;
```
squidsoup, line 3: so I think we just want the errors to have the type TSFixMe, so it's clear that we haven't defined a type for them yet, I don't think we need `ErrorState`.

ui/src/app/base/types.ts
```diff
@@ -0,0 +1,104 @@
...
+  tags: string[];
+  total: PodHint;
+  type: string;
+  updated: string;
+  used: PodHint;
+  zone: number;
+};
+
+export type PodState = {
+  errors: ErrorState;
```
squidsoup, line 51: This should just be `TSFixMe`.

ui/src/app/base/types.ts
Caleb-Ellis: Yeah that's fine - I figured at some point we would define a type for errors in state so we would only need to make the change in one place. But maybe it's better to highlight the fact that it's *not* currently typed wherever it's used.

PR #1278, opened 2020-06-19
Add sorting to KVMListTable.

Caleb-Ellis:
## Done

- Added sorting to KVMListTable. A bit of extra work was required to be able to sort by things that aren't parameters of the KVMs themselves (power, OS, resource pool).
- Added types for Controller, ControllerState, Machine, MachineState and TestStatus.

### MAAS deployment

- [Bolla](/HACKING.md#bolla)

### Running the branch

You can run this branch by:

### QA steps

- Go to /MAAS/r/kvm and check that you can sort by every table header.

## Fixes

Fixes: canonical-web-and-design/MAAS-squad#2024

### DIFF
--- a/ui/src/app/kvm/views/KVMList/KVMListTable/KVMListTable.tsx
+++ b/ui/src/app/kvm/views/KVMList/KVMListTable/KVMListTable.tsx
@@ -1,7 +1,8 @@
 import { Col, MainTable, Row } from "@canonical/react-components";
-import React, { useEffect } from "react";
+import React, { useEffect, useState } from "react";
 import { useDispatch, useSelector } from "react-redux";
 
+import { getStatusText } from "app/utils";
 import {
   controller as controllerActions,
   general as generalActions,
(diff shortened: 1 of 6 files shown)


ui/src/app/kvm/views/KVMList/KVMListTable/KVMListTable.tsx
```diff
@@ -52,49 +139,147 @@ const KVMListTable = (): JSX.Element => {
     dispatch(zoneActions.fetch());
   }, [dispatch]);
 
+  // Update sort parameters depending on whether the same sort key was clicked.
+  const updateSort = (newSortKey: Sort["key"]) => {
```
huwshimi, line 143: Do you think this function or any other bits of functionality could be shared via utils or hooks with other tables like the machine list?

Caleb-Ellis: Definitely! I initially wanted it to be generic but in the machine list all the data needed for sorting exists on the machines themselves, which isn't the case for the KVM list. I'll try and figure out a good way to abstract out the sorting functions in a way that can work for both.

huwshimi: Happy for that to be an update if you'd like to land all this work first. This branch is already rather large.

Caleb-Ellis: Yeah that works for me. Maybe it would be a good idea to do that once selecting is done too we we can create hooks/utils for both.

huwshimi: @Caleb-Ellis yep that sounds reasonable to me, maybe just make a card to remember.

Caleb-Ellis: Opened an issue here: #1279

PR #1337, opened 2020-07-06
Hoist shared properties to SimpleNode and Node types.

squidsoup:
## Done

- Add `Node` and `SimpleNode` types to simplify `Device`, `Machine` and `Controller` types.
- Update test factories to reflect changes to types.
- Add a few new types, `ModelRef`, `Vlan`, `Device` and `NodeStatus`.

Note: I wasn't able to find a way to extend a test factory from multiple types, so had to abandon the `DeviceMachineNode` type in favour of some duplication between `Device` and `Machine` types.

### MAAS deployment

- [Bolla](/HACKING.md#bolla)

### Running the branch

You can run this branch by:

### QA steps

- n/a

## Backports

### DIFF
--- a/ui/src/app/base/types.ts
+++ b/ui/src/app/base/types.ts
@@ -139,17 +139,17 @@ export type Model = {
 /**
  * A named foreign model reference, e.g. machine.domain
  */
-export type ModelRef = {
+export type ModelRef = Model & {
   name: string;
-} & Model;
+};
 
-export type Vlan = {
+export type Vlan = Model & {
   name: string;
   fabric_id: number;
   fabric_name: string;
-} & Model;
+};
 
-export type User = {
+export type User = Model & {
   completed_intro: boolean;
   email: string;
   global_permissions: string[];
@@ -157,9 +157,9 @@ export type User = {
   last_name: string;
   sshkeys_count: number;
   username: string;
-} & Model;
+};
 
-export type Zone = {
+export type Zone = Model & {
   controllers_count: number;
   created: string;
   description: string;
@@ -167,9 +167,9 @@ export type Zone = {
   machines_count: number;
   name: string;
   updated: string;
-} & Model;
+};
 
-export type Pod = {
+export type Pod = Model & {
   architectures: string[];
   available: PodHint;
   capabilities: string[];
@@ -196,7 +196,7 @@ export type Pod = {
   updated: string;
   used: PodHint;
   zone: number;
-} & Model;
+};
 
 export type PodHint = {
   cores: number;
@@ -216,7 +216,7 @@ export type PodHintExtras = {
 /**
  * SimpleNode represents the intersection of Devices, Machines and Controllers
  */
-export type SimpleNode = {
+export type SimpleNode = Model & {
   actions: string[];
   domain: ModelRef;
   hostname: string;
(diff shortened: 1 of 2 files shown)


ui/src/app/base/types.ts
```diff
@@ -200,105 +215,131 @@ export type PodHintExtras = {
...
+  link_speeds: number[];
+  owner: string;
+  parent?: string;
+  primary_mac: string;
+  spaces: string[];
+  subnets: string[];
+  zone: ModelRef;
+} & SimpleNode;
+
+type NodeStatus =
```
huwshimi, line 247: I wonder if we need to define all these. It's probably fine, but as it's API data I wonder if we'd catch any errors by having these values at compile time...

squidsoup: Nice for tests when we're setting `<node>.status` as we'll get valid autocompletions:

image

ui/src/app/base/types.ts
```diff
@@ -104,12 +104,14 @@ export type ResourcePoolState = {
...
-  status: number;
-  pending: number;
-  running: number;
-  passed: number;
-  failed: number;
+  status: TestResult;
+  pending: TestResult;
+  running: TestResult;
+  passed: TestResult;
+  failed: TestResult;
```
Caleb-Ellis, line 114: I don't think this is right. `status` can be any one of the following, according to `app/base/enum.js`:
``` js
export const scriptStatus = {
  NONE: -1,
  PENDING: 0,
  RUNNING: 1,
  PASSED: 2,
  FAILED: 3,
  TIMEDOUT: 4,
  ABORTED: 5,
  DEGRADED: 6,
  INSTALLING: 7,
  FAILED_INSTALLING: 8,
  SKIPPED: 9,
  APPLYING_NETCONF: 10,
  FAILED_APPLYING_NETCONF: 11,
};
```

And the rest of the parameters should just be `number` because they're simply counts of how many tests are in `pending`/`running`/etc state

ui/src/app/base/types.ts
```diff
@@ -136,6 +138,19 @@ export type Model = {
   id: number;
 };
 
+/**
+ * A named foreign model reference, e.g. machine.domain
+ */
+export type ModelRef = {
+  name: string;
+} & Model;
```
Caleb-Ellis, line 146: I wonder if there's a convention for type unions, i.e. should it be `Model & { extra: "stuff" }` or `{ extra: "stuff" } & Model`?

ui/src/app/base/types.ts
squidsoup: Ah quite right, thanks @Caleb-Ellis. There's no way to nicely represent a range in typescript, so I might just revert this and keep them all numbers.

ui/src/app/base/types.ts
squidsoup: I think I prefer the former actually, for long type definitions, it makes it clearer having the existing type in the intersection first.

ui/src/app/base/types.ts
huwshimi: I guess that's a case that the data doesn't really change, but we should probably be careful of not adding too much dynamic data that might give us issues and a maintenance burden down the track.

PR #1340, opened 2020-07-06
Add segmented prop to Meter component.

Caleb-Ellis:
## Done

- Added `segmented` prop to Meter component, which separates each discrete value in a meter with a 1px separator (whose colour can be customised with the `separatorColor` prop).
- Updated Meter labels and CPU header label in KVM list
- Updated `formatBytes` util to be able to convert to a specific unit, instead of just the nearest unit in which the value would be between 1 - 1023. This was needed to update the Meter labels so that you could do e.g. "0.1 of 10 GB assigned" instead of "100 MB of 10 GB assigned".
- Updated loading logic for KVMList cells that rely on host data. Now if the `pod` object does not have a `host` key, the cell value automatically goes to "Unknown" instead of waiting for all the controllers/machines to load first. 

### MAAS deployment

- [Bolla](/HACKING.md#bolla)

### Running the branch

You can run this branch by:

### QA steps

- Go to /MAAS/r/kvm and check that the meters match the [design](https://app.zeplin.io/project/5ed922c192bf80b9c37cb4fb/screen/5efa229d3f80d013734e9b21)
- Check that the table looks good on all screen sizes

## Fixes

Fixes canonical-web-and-design/MAAS-squad#1949

## Screenshot
0 0 0 0_8400_MAAS_r_kvm (14)


ui/src/app/base/components/Meter/Meter.tsx
```diff
@@ -19,18 +52,41 @@ const Meter = ({
...
   const valueSum = data.reduce((sum, datum) => sum + datum.value, 0);
   const maximum = max || valueSum;
   const widths = data.map((datum) => (datum.value / maximum) * 100);
+  const [segmentSize, setSegmentSize] = useState(0);
+
+  useEffect(() => {
+    if (segmented) {
+      updateSegmentSize(el, maximum, setSegmentSize);
+    }
+  }, [maximum, segmented]);
```
anthonydillon, line 70: Is it likely segmented with change dynamically?

Caleb-Ellis: Almost 0% chance probably, but the linter complains unless you include all the things that could possibly change.

lyubomir-popov:
@Caleb-Ellis I did some experimenting, this works best as far as I can tell:
https://codepen.io/lyubomir-popov/pen/WNrMPMV

In short, I define the maximum number of segments as a fraction (`*.5` below) of the available width in px. Beyond that, the segments are clamped to that max value.

```
	--width-unitless: 120;
	--w: calc(var(--width-unitless) * 1px);
	
	// this controls the maximum density - the point after which we assume the eye won't count
	--clamp-factor: .25;
	--count-clamp: var(--width-unitless) * var(--clamp-factor);
	--gap: 1.01px; /* a bit over 1 px to antialias properly */
	--b: calc(100%/var(--count) - var(--gap)); /* subtract gap from the width of the blue segment, to ensure last segment doesn't become smaller */
	--e: calc(100%/var(--count));
```

width 160, clamp factor .25:
image

width 120, clamp factor .5:
image

anthonydillon:
@lyubomir-popov I still see some anti-aliasing issues with this version. Different browsers also handle subpixels differently so unlike to get it perfect on all of them. Are you proposing we add 400 classes with styles to the codebase to show segments?

lyubomir-popov:
@anthonydillon 

> I still see some anti-aliasing issues with this version. Different browsers also handle subpixels differently so unlike to get it perfect on all of them.
Screenshot and browser please? I can fine tune with a bit of configuraiton of the variables. At the default width of the table we can ensure gaps falling precisely on pixels, as we can simply clamp to half the width - then regardles of the browser all pixels will be either fully on or fully off. 

> Are you proposing we add 400 classes with styles to the codebase to show segments?
Of course not - this is just to test I can see @Caleb-Ellis is currently doing it with inline-styles in ui/src/app/base/components/Meter/Meter.js → ui/src/app/base/components/Meter/Meter.tsx, I'd expect for this to work the same way.

anthonydillon:
@lyubomir-popov I am referring to the image above:
image

lyubomir-popov:
I see what you mean. I'd say that's a lesser compromise than varying the width of the overall bar. Out of 8 arbitrary segment counts, it happened just once, at 37 cores. 

Also the image above is enlarged. At the actual width in the codepen it looks acceptable to me.

codepen io_lyubomir-popov_pen_WNrMPMV

anthonydillon:
If it only happens on odd numbers we could round to even? It's not like people are counting them.

lyubomir-popov:
> If it only happens on odd numbers we could round to even? It's not like people are counting them.

Good idea. I think rounding them is safe above 32. Another possible optimisation - make the width of the overall bar a power of 2 - this way any power of 2 will fall niceley on precise pixels. For the screenshot underneath I've made the bars 128px wide, and it improves a bit: 

codepen io_lyubomir-popov_pen_WNrMPMV (1)

(the last 3 have been clamped, so appear the same)

Caleb-Ellis:
@anthonydillon @lyubomir-popov So I've played around with this a bit more and I think the "power of 2" bar width is the best option for relatively consistent segment *and* bar widths, but I still think my original version looks the best when viewed in isolation.

It looks strange when viewed at one of the base 2 breakpoints (e.g 127px) because it creates a bit of empty space on the left. I think this is okay because at the default viewport width it looks better, but still interested to hear how you think this could be better handled (but I don't think handling it at the table level is ideal - it would be nice for this Meter component to "just work" wherever we put it as opposed to special-casing the widths in the containers).

## Full viewport
Screenshot_2020-07-07 KVM bolla MAAS


## Just under breakpoint
Screenshot_2020-07-07 KVM bolla MAAS(1)

lyubomir-popov:
@Caleb-Ellis sorry forgot to ask yesterday, can we use the new blue for links and bar graphs (#06c)? Or is it still difficult due to the multiple vanilla versions?

Caleb-Ellis:
> @Caleb-Ellis sorry forgot to ask yesterday, can we use the new blue for links and bar graphs (#06c)? Or is it still difficult due to the multiple vanilla versions?

Updated it now. Some of the styling (color, widths) is determined in the javascript and I just forgot to update it when upgrading Vanilla.

ui/src/app/base/selectors/pod/pod.ts
```diff
@@ -137,7 +137,7 @@ const getHost = createSelector(
   [machine.all, controller.all, (_: RootState, pod: Pod) => pod],
   (machines: Machine[], controllers: Controller[], pod: Pod) => {
     if (!(pod && pod.host)) {
-      return null;
+      return undefined;
```
squidsoup, line 140: Why this change, I would have thought `null` (absence of value) made more sense?

ui/src/app/base/selectors/pod/pod.ts
```diff
@@ -137,7 +137,7 @@ const getHost = createSelector(
   [machine.all, controller.all, (_: RootState, pod: Pod) => pod],
   (machines: Machine[], controllers: Controller[], pod: Pod) => {
     if (!(pod && pod.host)) {
```
squidsoup, line 139: You can use optional chaining here `!(pod?.host)`

https://www.typescriptlang.org/docs/handbook/release-notes/typescript-3-7.html#optional-chaining

ui/src/app/kvm/views/KVMList/KVMListTable/OSColumn/OSColumn.tsx
```diff
@@ -13,37 +13,44 @@ import DoubleRow from "app/base/components/DoubleRow";
...
 
-  let osText = "Unknown";
-  if (host) {
-    osText = getStatusText(host, osReleases);
-  } else if (!host && loading) {
-    osText = "";
-  }
+  if (pod) {
+    const loading =
+      Boolean(pod.host) &&
```
squidsoup, line 34: why the `Boolean` cast here? That seems a bit unidiomatic.

ui/src/app/utils/formatBytes.ts
```diff
@@ -1,27 +1,32 @@
...
+  precision?: number;
+};
 
 /**
- * Format bytes to the appropriate unit at supplied precision.
+ * Format bytes to the appropriate/supplied unit at supplied precision.
  *
  * @param {number} value - the value in the supplied byte unit
  * @param {string} unit - the byte unit, e.g. "KB", "TB"
  * @param {object} config - config object
```
squidsoup, line 14: `FormatBytesConfig` rather than `object` in jsdoc here

ui/src/app/base/components/Meter/Meter.tsx
```diff
@@ -0,0 +1,171 @@
...
+  overColor?: string;
+  segmented?: boolean;
+  separatorColor?: string;
+  small?: boolean;
+};
+
+const Meter = ({
+  className,
+  data,
+  emptyColor = "#D3E4ED",
```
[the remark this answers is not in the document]
Caleb-Ellis: Yep I can add some consts! As for the naming I think keeping it as "color" makes sense because it reflects how inline styles (`background-color`, `border-color`)

ui/src/app/base/selectors/pod/pod.ts
Caleb-Ellis: This was a "strict mode" change. The selector returns a `Host`, which can be either a `Machine` or a `Controller`, or it doesn't exist (i.e. `undefined`), but it can't be `null`. It's a bit pedantic but it makes sense, or at least more sense than `type Host =  Machine | Controller | null`.

ui/src/app/base/selectors/pod/pod.ts
Caleb-Ellis: Adding that now

ui/src/app/kvm/views/KVMList/KVMListTable/OSColumn/OSColumn.tsx
Caleb-Ellis: Another strict mode change. `loading` is meant to be a boolean, but if it short circuits here it actually becomes a string (`pod.host`) which isn't right. I prefer this method over `!!pod.host` which is a bit cryptic if you're not familiar with javascript.

ui/src/app/utils/formatBytes.ts
Caleb-Ellis: Adding that now

ui/src/app/base/selectors/pod/pod.ts
Caleb-Ellis: Changed it so that the selector returns `Host | null` instead

lyubomir-popov:
@Caleb-Ellis I get an endless loading when I try to run it as a snup with docker on my mac (not inside multipass):

image

is 8400 still the correct port? tried 8402 as  Isaw that in the terminal , but no luck either

ui/src/app/kvm/views/KVMList/KVMListTable/OSColumn/OSColumn.tsx
squidsoup: I think maybe we should consider turning on https://github.com/typescript-eslint/typescript-eslint/blob/master/packages/eslint-plugin/docs/rules/strict-boolean-expressions.md which prevents any values in boolean expressions being anything other than `boolean` or `never`. What do you think?

Caleb-Ellis:
> @Caleb-Ellis I get an endless loading when I try to run it as a snup with docker on my mac (not inside multipass):
> 
> image
> 
> is 8400 still the correct port? tried 8402 as Isaw that in the terminal , but no luck either

Do you mean you're running the entire MAAS snap? That won't have these changes, even on edge, because it hasn't been merged yet. You should just be able to pull this code, update `proxy/.env.local` to point to bolla and then `yarn` and `yarn start`. After it's started up you should just need to connect to the VPN to be able to view the UI.

I forgot to mention that redirects from `/` to `/MAAS/r` havent been set up yet so you have to manually go to `/MAAS/r` or whatever path you want to go to.

Anyway I'm going to merge this and you can review it as part of the epic review (canonical-web-and-design/MAAS-squad#1955) - it'll be up on bolla so won't need any local set up.

lyubomir-popov:
Thanks @Caleb-Ellis, are these instructions available anywhere? 
The readme used to contain them but now doesn't.

PR #1344, opened 2020-07-06
Move types to model dirs

huwshimi:
## Done

- Created basic store structure.
- Split up types into store dirs.
- Updated imports to point to the store.

## Fixes

Fixes: https://github.com/canonical-web-and-design/maas-ui/issues/1343.

### Reviewer notes

This is all code copy and paste and updating imports. See if you agree with how I've divided the types. Particularly the machine/controller/node types as I had to make some extra splits to avoid circular imports.


ui/src/app/store/node/types.ts
```diff
@@ -0,0 +1,82 @@
...
+  | "Failed to enter rescue mode"
+  | "Exiting rescue mode"
+  | "Failed to exit rescue mode"
+  | "Testing"
+  | "Failed testing";
+
+/**
+ * Node represents the intersection of Machines and Controllers
+ */
+export type BaseNode = SimpleNode & {
```
huwshimi, line 63: This was previously named `Node` but that conflicts with the DOM Node feature.

squidsoup: Ah, out of interest, how did you realise that? I wasn't seeing any warnings or eslint errors. The docstring does need updating here to reflect the new name.

ui/src/app/store/host/types.ts
```diff
@@ -0,0 +1,4 @@
+import type { Controller } from "app/store/controller/types";
```
squidsoup, line 1: If our goal with this refactor is to have every subdirectory under `store` a representation of a slice of state, then I don't think `store/host` makes sense. I guess this is the problem with having more granular type modules. Perhaps just co-locate this with Machine?

ui/src/app/store/node/types.ts
```diff
@@ -0,0 +1,82 @@
+import type { Model } from "app/store/types";
```
squidsoup, line 1: Same again, `node` isn't a slice, so I think these types should probably live somewhere else. Perhaps these should live in the `store/types.ts` module?

ui/src/app/store/node/types.ts
huwshimi: When I moved the types around it didn't complain about missing variables and the TypeScript hints were for the DOM Node type.

ui/src/app/store/node/types.ts
huwshimi: I'm thinking maybe store/types should be a dir with store/types/node.ts etc. to make it a bit more obvious where those types live. I'm sure we'll have a lot of shared types that will need to live externally to avoid circular imports.

ui/src/app/store/host/types.ts
huwshimi: Moved to store/types/host

ui/src/app/store/node/types.ts
huwshimi: store/types/node.ts

ui/src/app/store/types/node.ts
```diff
@@ -0,0 +1,82 @@
+import type { Model } from "app/store/types/model";
+
+/**
+ * A named foreign model reference, e.g. machine.domain
+ */
+export type ModelRef = Model & {
```
squidsoup, line 6: This could be any model, so I'm not sure it belong in `node`

PR #1346, opened 2020-07-06
Add husky and pre-push betterer hook.

squidsoup:
In order to make the best use of betterer, and ensure that new typescript issues we introduce are immediately obvious, I'd like to try the [approach recommended by betterer maintainer](https://github.com/phenomnomnominal/betterer/issues/144) for a while. I know there's been some contention around git hooks in the past, but they do provide an immediacy which means we're more likely to try to address issues as we create them. That's the theory anyway :)

This means we would now have the following workflow:

1) Push some commits
2) Betterer runs and alerts us to any regressions moving us further from strict mode (or whatever other betterer tests we've defined). You will _only_ see issues that your commits have introduced as we're ensuring that results are updated with every PR (CI will now fail if the results file has not been updated).
3) Take the opportunity to fix the typescript errors, or decide that it's too hard for now.
4) If necessary, force update the results file for the issues you've decided to defer with `yarn betterer --update`.

## Done
- Add husky and a pre-push hook which runs betterer.
- Remove `continue-on-error` in CI to ensure that each PR has updated betterer results correctly.

### MAAS deployment

- [Bolla](/HACKING.md#bolla)

### Running the branch

You can run this branch by:

### QA steps

- Commit some junk, betterer should run. Unless you've introduce a new typescript issue, it should exit 0.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -41,7 +41,6 @@ jobs:
           node-version: ${{ matrix.node-version }}
       - run: CYPRESS_INSTALL_BINARY=0 yarn install
       - run: yarn betterer > results.txt
-        continue-on-error: true
         env:
           CI: true
       - name: Upload betterer results
--- a/package.json
+++ b/package.json
@@ -44,7 +44,7 @@
   },
   "husky": {
     "hooks": {
-      "post-commit": "yarn run betterer"
+      "pre-push": "yarn run betterer"
     }
   },
   "workspaces": {
--- a/package.json
+++ b/package.json
@@ -42,6 +42,11 @@
     "test": "yarn build-shared && yarn test-legacy && yarn test-ui && yarn test-shared",
     "ui": "cd proxy && yarn start-ui"
   },
+  "husky": {
+    "hooks": {
+      "pre-commit": "yarn run betterer"
+    }
+  },
   "workspaces": {
     "packages": [
       "legacy",
@@ -57,5 +62,8 @@
       "**/jest",
       "**/webpack**"
     ]
+  },
+  "devDependencies": {
+    "husky": "4.2.5"
   }
 }
--- a/package.json
+++ b/package.json
@@ -44,7 +44,7 @@
   },
   "husky": {
     "hooks": {
-      "pre-commit": "yarn run betterer"
+      "post-commit": "yarn run betterer"
     }
   },
   "workspaces": {
--- a/ui/.betterer.results
+++ b/ui/.betterer.results
@@ -1,47 +1,103 @@
-// BETTERER RESULTS V1.
+// BETTERER RESULTS V2.
 exports[`stricter compilation`] = {
-  timestamp: 1592774737565,
   value: `{
-    "src/app/base/reducers/pod/pod.test.ts:1930988215": [
-      [144, 20, 2, "Type \'number\' is not assignable to type \'never\'.", "5861160"],
-      [144, 27, 4, "Type \'string\' is not assignable to type \'never\'.", "2087876002"]
+    "src/app/base/actions/pod/pod.test.ts:2464677720": [
+      [61, 15, 7, "Property \'refresh\' does not exist on type \'{ fetch: ActionCreatorWithPreparedPayload<[any?], { params: any; }, string, never, { model: any; method: string; }>; create: ActionCreatorWithPreparedPayload<[any?], { params: any; }, string, never, { ...; }>; update: ActionCreatorWithPreparedPayload<...>; delete: ActionCreatorWithPreparedPayload<...>; cleanup: Acti...\'.", "1380666520"],
+      [76, 15, 11, "Property \'setSelected\' does not exist on type \'{ fetch: ActionCreatorWithPreparedPayload<[any?], { params: any; }, string, never, { model: any; method: string; }>; create: ActionCreatorWithPreparedPayload<[any?], { params: any; }, string, never, { ...; }>; update: ActionCreatorWithPreparedPayload<...>; delete: ActionCreatorWithPreparedPayload<...>; cleanup: Acti...\'.", "1023496814"]
     ],
-    "src/app/base/selectors/pod/pod.ts:361290740": [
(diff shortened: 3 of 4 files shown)


nottrobin:
The less intrusive version of this would be to add a `setup-git-hooks` script to `package.json`, and then mention in the README that you should run `yarn setup-git-hooks` (or `dotrun setup-git-hooks`) the first time you clone the project.

I never work on MAAS-UI, so I have less of a strong opinion about this project, but I can confirm that many other people have also been frustrated (and confused) by Husky automatically setting up git hooks. So I really do believe it adds a significant amount of unnecessary friction for those trying to run your project.

Would it be possible for you to try doing this as an optional `package.json` script instead?

squidsoup:
> I never work on MAAS-UI, so I have less of a strong opinion about this project, but I can confirm that many other people have also been frustrated (and confused) by Husky automatically setting up git hooks. So I really do believe it adds a significant amount of unnecessary friction for those trying to run your project.

Can you clarify what is frustrating or confusing about having a git hook setup automatically? Perhaps it is unexpected the first time you see it, but surely isn’t thereafter. Husky had 3M weekly downloads, it’s not an infrequently used tool.

I’m not sure I understand the source of friction.

> Would it be possible for you to try doing this as an optional `package.json` script instead?

It’s deliberately not optional, to ensure regressions we introduce are obvious and visible.

nottrobin:
The one in ubuntu.com breaks quite often, takes a long time etc. It is confusing to many people even after they've seen it every time they commit, because they never understood what was happening, they just accepted it. It slows down their commits quite considerably, and they don't know why, they just persist with a degraded experience, allowing it to slow down their workflow and eroding their sense of control, until it literally breaks sometime and they reach out to me or someone else for help.

I know this has specifically frustrated at least @pmahnke, @anthonydillon and @sowasred2012. Personally I have no problem simply doing `rm .git/hooks/*`, which is what I advise others to do when they encounter hooks they didn't personally set up and don't understand.

I do get that what you're proposing here is a pre-push hook, which is less annoying than a pre-commit hook, but still if it slows people down it might lead to them syncing their work with their feature branch less often, which is generally something that should be encouraged. And it's more of a point of principle.

I just personally find it very obnoxious for anything to automatically set up background tasks that jump into my local development workflow and slow me down, and as I say, if someone tries to add such steps, I will simply delete them. My local git repos are my domain, and if my personal workflow involves commiting or pushing several times a minute, Husky has no business adding expensive operations into that local workspace. It breaks the powerful philosophy of distributed version control. It's when someone wants to integrate with the upstream repo that checks and blocks should happen.

squidsoup:
> The one in ubuntu.com breaks quite often, takes a long time etc.
What breaks or takes a long time? Surely that has nothing to do with the git hook, and what the hook is running.

> It is confusing to many people even after they've seen it every time they commit
The hook is `pre-push` in this branch, so it won't be seen on every commit.

> because they never understood what was happening, they just accepted it.
This is part of our workflow, so anyone working on maas-ui is expected to understand the betterer output.

> eroding their sense of control
Fixing betterer issues is a requirement to land your branch, so it is not optional anyway.

> It's when someone wants to integrate with the upstream repo that checks and blocks should happen.
We don't have a means of displaying betterer output on a PR, so this is our solution for now.

Addressing project quality issues is our priority. As I stated in the PR, the plan is to try it for a while. If our team finds it onerous, we can remove it.

nottrobin:
> This is part of our workflow, so anyone working on maas-ui is expected to understand the betterer output.

It's not simply understanding betterer, it's understanding the concept of a pre-push hook, which is quite a surprising concept to anyone not used to them (and they don't exist in the vast majority of git projects). It's not betterer that erodes their sense of control, it's git hooks, and specifically git hooks that they didn't deliberately and knowingly set up.

For me the whole thing can be solved by simply making it an explicit step in starting work on MAAS-UI. If I have to explicitly run `setup-betterer-hook` myself, then I am at least going to be mildly curious what that might have done, and hopefully I will remember having run that when I see surprising output I'm not used to when running `git push`. If you consider running that step to be a requirement for all devs who want to work on the project, simply say so in the README.

Ultimately, Husky or no, these hooks can be simply deleted by devs, and probably will be. So it's not actually a good mechanism for you to enforce anything that is a solid requirement. The only place that it's actually possible to enforce these checks is upon integration into upstream. That's just a fact. The actual utility of git hooks is to help the developers themselves, not to enforce anything. And so it is absolutely right and proper to recommend and help them set up git hooks that are going to ultimately save them time by catching things that would have failed in CI, but it's pointless if they aren't aware of it.

And it's all very well to say "understanding this is simply a requirement for people to work on this repo" but the whole reason I'm pushing back here is because we need to be very careful about adding requirements for people to work on the repo. It is strongly desirable for our repositories to be as simple to work on as possible. The costs associated with adding 1 extra requirement to what people have to understand to work on your repo, when considered in terms of the people who are able to review PRs, whether we can rotate devs into the MAAS squad, the cost of onboarding new joiners, are large.

I understand that MAAS is already a complex thing to work on, but that's not in and of itself an excuse to make it moreso.

squidsoup:
While I appreciate the feedback @nottrobin, and I understand you have a personal opinion about this, we need to do what's best for the project. As I mentioned, if the team finds it onerous, we'll remove it. I'll continue to try to find a way to get the betterer output into PRs, after which this becomes mostly redundant.

nottrobin:
> While I appreciate the feedback @nottrobin, and I understand you have a personal opinion about this, we need to do what's best for the project. As I mentioned, if the team finds it onerous, we'll remove it. I'll continue to try to find a way to get the betterer output into PRs, after which this becomes mostly redundant.

I'm not quite sure what to say here. We have discussed it twice in dev catch-up and twice we came to a agreement there to use the pattern I describe instead of Husky - which also seems like it would work here. So it does seem somewhat right to at least bring it back up again in that meeting.

However, if you really feel this is a blocker for your project, then go ahead and disregard me.

anthonydillon:
This is quite the rabbit hole, Typescript > Betterer > Husky. 

We have discussed the use of Husky in our projects a few times and there is generally a sense of annoyance over time when used in projects. But, it seems this is a pain that has not been experienced by the React apps. 

I believe it is ok to introduce this here with the proviso that we continue to look for a way to output the result as a comment in the PR and do not subdue any frustration introduced by Husky.

@nottrobin is right that we need to be careful about adding requirements for people to work on the repo and it can drive maas-ui into a silo'ed position where only the devs that work on the project can work on it. We have seen many of these projects and should avoid it at all costs.

PR #1347, opened 2020-07-07
Add vscode betterer config

squidsoup:
## Done

- Add vs code workspace config for betterer extension
- Remove js file count test, as it [appears to be failing](https://github.com/phenomnomnominal/betterer/issues/149) when run in the extension context. This test isn't as valuable as having immediate feedback in the IDE for "stricter compilation", so I think dropping it isn't a huge loss.

image

image

### MAAS deployment

- [Bolla](/HACKING.md#bolla)

### Running the branch

You can run this branch by:

### QA steps

- Install the betterer vs code extension.
- Rename a js file to ts, wait a bit (can take a while for problems to appear), check the problems tab and you should hopefully see some betterer output. You should also get some red squiggles in the IDE (just as we do for ts compiler errors and eslint).

## Backports

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -8,5 +8,8 @@
     "cSpell.words": [
         "lifecycles",
         "unmount"
-    ]
+    ],
+    "betterer.configPath": "./ui/.betterer.ts",
+    "betterer.resultsPath": "./ui/.betterer.results",
+    "betterer.tsconfigPath": "./ui/tsconfig.json"
 }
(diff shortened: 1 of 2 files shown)


ui/.betterer.ts
```diff
@@ -1,24 +1,9 @@
...
-)
-  .replace("\n", "")
-  .trim());
 
 export default {
   "stricter compilation": typescriptBetterer("./tsconfig.json", {
     strict: true,
   }),
   "no TSFixMe types": regexpBetterer("**/*.ts", /(\s*TSFixMe)/i),
-  "migrate js files to ts": {
```
huwshimi, line 19: Awww.. i thought this was a good reminder if you absentmindedly added a js file instead of a ts. Can we rewrite this somehow?

squidsoup: I can try a different approach to counting files perhaps.

squidsoup: I think this is actually a bug in the betterer extension, not anything to do with the implementation of the file count. I think it does probably make sense to remove the test until that's resolved so that can get some IDE feedback.

PR #1352, opened 2020-07-08
Move selectors to store

huwshimi:
## Done

- Move all selectors to store dirs.
- Update selectors to TS.

## Reviewer notes

These are mostly file moves and renames to .ts with minimal changes to make TS happy.

## Fixes


huwshimi:
Will be replaced with smaller PRs.

PR #1363, opened 2020-07-08
Create factories for user and auth models

huwshimi:
## Done

- Added factories for user and auth types.
- Updated selector tests to use the user state factory

## Fixes

Fixes: https://github.com/canonical-web-and-design/maas-ui/issues/1357.

### DIFF
--- a/ui/src/app/base/selectors/auth/auth.test.js
+++ b/ui/src/app/base/selectors/auth/auth.test.js
@@ -1,10 +1,10 @@
 import auth from "./auth";
-import { userState } from "testing/factories";
+import { userState as userStateFactory } from "testing/factories";
 
 describe("auth", () => {
   it("can get the current user details", () => {
     const state = {
-      user: userState({
+      user: userStateFactory({
         auth: {
           user: { username: "admin" },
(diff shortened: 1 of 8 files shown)


ui/src/app/base/selectors/auth/auth.test.js
```diff
@@ -1,72 +1,69 @@
 import auth from "./auth";
+import { userState } from "testing/factories";
```
squidsoup, line 2: I think the convention `import { userState as userStateFactory }` might be worth adopting (I'm doing that in my branch), to make it clear what these functions are.

ui/src/app/base/selectors/auth/auth.test.js
```diff
@@ -1,72 +1,69 @@
...
 
 describe("auth", () => {
   it("can get the current user details", () => {
     const state = {
-      user: {
+      user: userState({
```
huwshimi, line 7: I'm not totally sure if these should all get updated as this state object will need to get replaced with the root state factory, but we can't do that until we have factories for all the state objects.

PR #1376, opened 2020-07-09
Change cypress MAAS snap from latest/edge to 2.8/edge

Caleb-Ellis:
## Done

- Changed cypress MAAS snap from latest/edge to 2.8/edge for the 2.8 branch,

### QA steps

- `test_cypress` should pass

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -46,7 +46,7 @@ jobs:
           name: Install MAAS snap and set up postgres database
           command: |
             sudo systemctl enable snapd
-            sudo snap install maas --channel=latest/edge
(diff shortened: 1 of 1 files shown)


huwshimi:
Hmm, not sure about this. It's possible we'll add integration tests against new features that include new API endpoints etc. so in those cases we'd need the edge snap.

squidsoup:
> Hmm, not sure about this. It's possible we'll add integration tests against new features that include new API endpoints etc. so in those cases we'd need the edge snap.

This is only proposed against 2.8, master will remain on edge. I think that's okay @huwshimi?

huwshimi:
Oh hah, sorry, ignore me :)

PR #1378, opened 2020-07-10
Build KVM configuration form.

Caleb-Ellis:
## Done

- Added KVM configuration form.
- Changed FormikForm buttons so that "Cancel" button is only displayed if `onCancel` prop is provided.
- Drive by fix for refresh form in KVM details so that form closes when successfully completed.
- Drive by fix for TagSelector spacing

### MAAS deployment

- [Bolla](/HACKING.md#bolla)

### Running the branch

You can run this branch by:

### QA steps

- Go to /MAAS/r/kvm and click on a KVM
- Click on the "Configuration" tab and check that the config form shows up
- Check that you can change and save the values, and the saved values persist

## Fixes

Fixes canonical-web-and-design/MAAS-squad#1960
Fixes canonical-web-and-design/MAAS-squad#2039

## Screenshot
0 0 0 0_8400_MAAS_r_kvm_180_edit (1)

### DIFF
--- a/ui/src/app/base/components/Slider/Slider.test.tsx
+++ b/ui/src/app/base/components/Slider/Slider.test.tsx
@@ -4,20 +4,26 @@ import React from "react";
 import Slider from "./Slider";
 
 describe("Slider", () => {
-  it("renders without number input", () => {
+  it("renders correctly", () => {
     const wrapper = shallow(
-      <Slider max={10} min={0} onChange={jest.fn()} value={5} />
+      <Slider max={10} min={0} onChange={jest.fn()} showInput value={5} />
     );
     // Snapshot should not include extra number input beside the slider.
     expect(wrapper).toMatchSnapshot();
   });
 
-  it("renders with number input", () => {
+  it("can render without an additional number input", () => {
     const wrapper = shallow(
-      <Slider max={10} min={0} onChange={jest.fn()} showInput value={5} />
+      <Slider
+        max={10}
+        min={0}
+        onChange={jest.fn()}
+        showInput={false}
+        value={5}
+      />
     );
     // Snapshot should include extra number input beside the slider.
-    expect(wrapper).toMatchSnapshot();
+    expect(wrapper.find("input[type='number']").exists()).toBe(false);
   });
 
   it("correctly displays filled portion", () => {
--- a/ui/src/app/base/components/Slider/Slider.test.tsx
+++ b/ui/src/app/base/components/Slider/Slider.test.tsx
@@ -24,7 +24,7 @@ describe("Slider", () => {
     );
     const backgroundStyle = wrapper.find(".p-slider").props().style
       ?.background as string;
-    const trimmed = backgroundStyle.replace(/\s\s+/g, " ");
+    const trimmed = backgroundStyle?.replace(/\s\s+/g, " ");
     const expectedPercentage = "25%"; // ((4 - 2) / (10 - 2)) * 100 = (2 / 8) * 100 = 25
     expect(trimmed.includes(expectedPercentage)).toBe(true);
   });
--- a/ui/src/app/base/components/Slider/Slider.test.tsx
+++ b/ui/src/app/base/components/Slider/Slider.test.tsx
@@ -8,7 +8,6 @@ describe("Slider", () => {
     const wrapper = shallow(
       <Slider max={10} min={0} onChange={jest.fn()} showInput value={5} />
     );
-    // Snapshot should not include extra number input beside the slider.
     expect(wrapper).toMatchSnapshot();
   });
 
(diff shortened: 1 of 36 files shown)


ui/src/app/base/components/Slider/Slider.test.tsx
```diff
@@ -0,0 +1,48 @@
...
+      <Slider max={10} min={0} onChange={jest.fn()} value={5} />
+    );
+    expect(wrapper).toMatchSnapshot();
+  });
+
+  it("renders with number input", () => {
+    const wrapper = shallow(
+      <Slider max={10} min={0} onChange={jest.fn()} showInput value={5} />
+    );
+    expect(wrapper).toMatchSnapshot();
```
huwshimi, line 18: I might be good to be explicit in this or the above test so it's clear what's being tested.

ui/src/app/kvm/views/KVMDetails/KVMConfiguration/KVMConfiguration.tsx
```diff
@@ -1,28 +1,132 @@
...
     dispatch(podActions.fetch());
+    dispatch(resourcePoolActions.fetch());
+    dispatch(tagActions.fetch());
+    dispatch(zoneActions.fetch());
   }, [dispatch]);
 
-  return <></>;
+  const loaded = resourcePoolsLoaded && tagsLoaded && zonesLoaded;
+
+  if (typeof pod === "object" && loaded) {
```
huwshimi, line 80: I couldn't figure out why `typeof` is used here. How come a truthy check isn't enough (also `typeof null === "object"` returns true)? I'm probably missing something obvious...

ui/src/app/kvm/views/KVMDetails/KVMConfiguration/KVMConfiguration.test.tsx
```diff
@@ -0,0 +1,138 @@
...
+          initialEntries={[{ pathname: "/kvm/1/edit", key: "testKey" }]}
+        >
+          <KVMConfiguration />
+        </MemoryRouter>
+      </Provider>
+    );
+    expect(wrapper.find("Spinner").length).toBe(1);
+  });
+
+  it("can handle updating a KVM", () => {
```
huwshimi, line 83: I think it'd be good to add a test for a non-kvm as well as it send slightly different data.

ui/src/app/kvm/views/KVMDetails/KVMConfiguration/KVMConfiguration.tsx
huwshimi: Whatever this state is I think it'd be good to add a test for it.

huwshimi:
QA:
When I typed in the password field I got an error: "A component is changing an uncontrolled input of type password to be controlled".

Should there be a cancel button? Seems a bit strange that you can save and then cancel...

ui/src/app/kvm/views/KVMDetails/KVMConfiguration/KVMConfiguration.tsx
Caleb-Ellis: Yeah this definitely looks strange, but it's a strict mode fix. Firstly I had to fix having a non-boolean in a boolean check (since `pod` is not a boolean). But also I needed to fix betterer warnings in the conditional code block about `pod.attribute` not existing if `pod` is not an object. And I think adding `pod?.type`, `pod?.password`, etc everywhere is a bit messy.

There's has to be a better way of satisfying strict mode because I agree it's not really clear here.

ui/src/app/base/components/Slider/Slider.test.tsx
Caleb-Ellis: Good call

ui/src/app/kvm/views/KVMDetails/KVMConfiguration/KVMConfiguration.test.tsx
Caleb-Ellis: Good idea, one for lxd and one for virsh

Caleb-Ellis:
> Should there be a cancel button? Seems a bit strange that you can save and then cancel...

Yeah probably not. I just kept it the same as the other forms but I guess this is the first case in which it's not necessary

ui/src/app/kvm/views/KVMDetails/KVMConfiguration/KVMConfiguration.tsx
Caleb-Ellis: Opting instead for `if (!!pod && loaded) {`

Caleb-Ellis:
> QA:
> When I typed in the password field I got an error: "A component is changing an uncontrolled input of type password to be controlled".

Ahh looks like the password can be `undefined` which I think doesn't initialise the form properly. I'll make the default `""` and make sure not to send the param if the value is an empty string when submitting.

Issue #1362, opened 2020-07-08
Create a RootState factory


huwshimi:
Create a factory to mimic the root state type.

PR #1385, opened 2020-07-13
Add root state factory

huwshimi:
## Done

- Add root state factory and missing model state factories.
- Updated selector tests to use the factory.

## Fixes

Fixes: https://github.com/canonical-web-and-design/maas-ui/issues/1362.


ui/src/app/store/machine/types.ts
```diff
@@ -40,6 +40,31 @@ export type Machine = BaseNode & {
   zone: ModelRef;
 };
 
+export type MachineStatuses = {
+  [x: number]: {
```
Caleb-Ellis, line 44: The machine system_id is a string isn't it? I wonder if in this case `[system_id: string]` might be more explanatory

huwshimi: Ah yes, you're right. I wanted to do `Machine["system_id"]` there, but it wouldn't allow it.

PR #1389, opened 2020-07-14
Move general selectors

huwshimi:
## Done

- Move general selectors to the store dir and update to typescript.

## Fixes


ui/src/app/kvm/views/KVMList/AddKVMForm/AddKVMForm.tsx
```diff
@@ -55,7 +55,7 @@ export const AddKVMForm = (): JSX.Element => {
   const zones = useSelector(zoneSelectors.all);
   const zonesLoaded = useSelector(zoneSelectors.loaded);
 
-  const [hostType, setHostType] = useState("");
+  const [hostType, setHostType] = useState<TSFixMe>();
```
Caleb-Ellis, line 58: Is it possible to be `useState<Pod["type"]>()`?

huwshimi: I think that's what it will become, but for that to work I think I'd have to update a bunch of files to typescript so I opted to make this branch a bit smaller. Think that's OK?

Caleb-Ellis: Yeah that's fine :+1:

PR #1391, opened 2020-07-15
Move base selectors

huwshimi:
## Done

- Move base selectors to store. Update to TypeScript.

## Fixes


ui/src/app/store/machine/selectors.test.ts
```diff
@@ -0,0 +1,195 @@
+import {
```
huwshimi, line 1: These are not new files as the diff suggests. They have the same changes as the other selector tests.

PR #1395, opened 2020-07-16
Build KVM compose machine form with default interface/storage constraints.

Caleb-Ellis:
## Done

- Built KVM compose form without the interface or storage constraint sections (so just the normal inputs)

### MAAS deployment

- [Bolla](/HACKING.md#bolla)

### Running the branch

You can run this branch by:

### QA steps

- Go to a KVM details page and choose "Compose" from the action dropdown
- Check that defaults are loaded in correctly and the help text for cores and memory is correct
- Change some values and click "Compose machine"
- Check that a notification appears confirming the machine is created, and the new machine appears in the KVM's machine list.

## Fixes

Fixes canonical-web-and-design/MAAS-squad#1961

## Screenshot
0 0 0 0_8400_MAAS_r_kvm_185


squidsoup:
@caleb when composing, I would have expected the default to be the memory remaining, but it appears to be a higher value:

image

squidsoup:
Looks like there is potentially a rounding issue?

image

Caleb-Ellis:
> @caleb when composing, I would have expected the default for memory to be the amount remaining, but it appears to be a higher value:

@squidsoup Hmm... I think a warning would make sense here. I'll change it so that the value isn't auto-filled, but is instead in the placeholder. And if the available RAM is less than the default, a warning will show.

> Looks like there is potentially a rounding issue?

Yep, the default precision in `formatBytes` is 3 (so precise up to 999). I'll bump it up to 6 - can't imagine there are many machines out there with more than 999999MiB of RAM.

ui/src/app/kvm/components/KVMActionFormWrapper/ComposeForm/ComposeForm.tsx
```diff
@@ -0,0 +1,157 @@
...
+  pool: string;
+  storage: string;
+  zone: string;
+};
+
+type Props = {
+  setSelectedAction: (action: string | null) => void;
+};
+
+const ComposeForm = ({ setSelectedAction }: Props): JSX.Element | null => {
```
squidsoup, line 40: This never returns `null`, return type needs updating

PR #1413, opened 2020-07-20
Add error message to deploy form if no images have been downloaded.

### LINKED ISSUE #1393: Deploy form crashes if no images
**Describe the bug**
If you don't have any images added and you try to open the deploy form from the machine list it has an error about no release options and displays an error instead of the form.

Caleb-Ellis:
## Done

- Disable and add an error message to the deploy form if no OSes/releases are detected in state.

### MAAS deployment

- [Bolla](/HACKING.md#bolla)

### Running the branch

You can run this branch by:

### QA steps

- Spin up a new MAAS and do not download any images when prompted to. I don't think this will work if you try to delete all the images from an existing MAAS, as MAAS prevents users from deleting the default commissioning/deployment image.
- Select one or more machines and choose the "Deploy" action.
- The deploy form should be disabled with an error message that asks you to visit the images page.

## Fixes

Fixes #1393 

## Screenshot
0 0 0 0_8400_MAAS_r_machines (2)

### DIFF
--- a/ui/src/app/machines/views/MachineList/MachineListHeader/ActionFormWrapper/DeployForm/DeployForm.tsx
+++ b/ui/src/app/machines/views/MachineList/MachineListHeader/ActionFormWrapper/DeployForm/DeployForm.tsx
@@ -55,7 +55,7 @@ export const DeployForm = ({ setSelectedAction }: Props): JSX.Element => {
   }, [dispatch]);
 
   // Default OS+release is set in the backend even if the image has not yet been
-  // downloaded. The following condiionals check whether the OS+release actually
+  // downloaded. The following conditionals check whether the OS+release actually
   // exist in state before setting initial values in the form.
   let initialOS = "";
   let initialRelease = "";
--- a/ui/src/app/machines/views/MachineList/MachineListHeader/ActionFormWrapper/DeployForm/DeployForm.tsx
+++ b/ui/src/app/machines/views/MachineList/MachineListHeader/ActionFormWrapper/DeployForm/DeployForm.tsx
@@ -43,7 +43,9 @@ export const DeployForm = ({ setSelectedAction }: Props): JSX.Element => {
   const defaultMinHweKernel = useSelector(
     generalSelectors.defaultMinHweKernel.get
   );
-  const osInfo = useSelector(generalSelectors.osInfo.get);
+  const { default_osystem, default_release, osystems, releases } = useSelector(
+    generalSelectors.osInfo.get
+  );
   const deployingSelected = useSelector(machineSelectors.deployingSelected);
 
   useEffect(() => {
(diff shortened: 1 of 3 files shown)


huwshimi:
@Caleb-Ellis do you think the error message should replace the form? The form can't be filled out and the message is a bit subtle being under the form.

Caleb-Ellis:
> @Caleb-Ellis do you think the error message should replace the form? The form can't be filled out and the message is a bit subtle being under the form.

Hmm, I want to try to avoid hiding markup behind conditionals as much as possible... maybe disabling the form fields and having an error notification at the top would make it more obvious?

Caleb-Ellis:
@huwshimi do you think this is a bit more obvious?

0 0 0 0_8400_MAAS_r_machines (3)

huwshimi:
@Caleb-Ellis that's definitely better. Curious to know a bit more about:
> I want to try to avoid hiding markup behind conditionals as much as possible

Caleb-Ellis:
> @Caleb-Ellis that's definitely better. Curious to know a bit more about:
> 
> > I want to try to avoid hiding markup behind conditionals as much as possible

@huwshimi there's a lot of places in the angular app where content is hidden depending on the current state, e.g:
- The RSD tab in the main nav only shows if the user has added at least one RSD.
- The "Take action" button only shows if you select something first.
- Some tabs in machine details are hidden if the machine does not have a particular networking configuration, or it's not a container/VM instance.

There's many more instances that don't come to mind, but it just feels like odd UX to hide things without any info why, or how to make it appear. It's not really the case here, because there *would* be an explanation (just the notification without the form fields) but in general I feel like "disable with message" gives a fuller picture for what's possible to do, as opposed to removing things.

huwshimi:
> There's many more instances that don't come to mind, but it just feels like odd UX to hide things without any info why, or how to make it appear. It's not really the case here, because there _would_ be an explanation (just the notification without the form fields) but in general I feel like "disable with message" gives a fuller picture for what's possible to do, as opposed to removing things.

Yeah, good point. I'll do another review shortly.

ui/src/app/machines/views/MachineList/MachineListHeader/ActionFormWrapper/DeployForm/DeployForm.tsx
```diff
@@ -52,16 +54,33 @@ export const DeployForm = ({ setSelectedAction }: Props): JSX.Element => {
     dispatch(machineActions.fetch());
   }, [dispatch]);
 
+  // Default OS+release is set in the backend even if the image has not yet been
+  // downloaded. The following condiionals check whether the OS+release actually
```
huwshimi, line 58: `condiionals`

ui/src/app/machines/views/MachineList/MachineListHeader/ActionFormWrapper/DeployForm/DeployFormFields/DeployFormFields.test.tsx
```diff
@@ -259,17 +239,30 @@ describe("DeployFormFields", () => {
...
+    const wrapper = mount(
+      <Provider store={store}>
+        <MemoryRouter
+          initialEntries={[{ pathname: "/machines/add", key: "testKey" }]}
+        >
+          <DeployForm setSelectedAction={jest.fn()} />
+        </MemoryRouter>
+      </Provider>
+    );
+    expect(wrapper.find("[data-test='images-error']").exists()).toBe(true);
```
huwshimi, line 263: Might also be good to check the fields are disabled.

ui/src/app/machines/views/MachineList/MachineListHeader/ActionFormWrapper/DeployForm/DeployForm.tsx
Caleb-Ellis: Thanks :+1:

ui/src/app/machines/views/MachineList/MachineListHeader/ActionFormWrapper/DeployForm/DeployFormFields/DeployFormFields.test.tsx
Caleb-Ellis: Done

PR #1428, opened 2020-07-29
Dismiss notifications with x icon

huwshimi:
## Done

- Update the notification list to dismiss with the x icon.
- Convert component to TypeScript.
- Update to use Notification component.

### MAAS deployment

- [Bolla](/HACKING.md#bolla)

### Running the branch

You can run this branch by:

### QA steps

- Add some notifications (I use `make harness` and then something like `from maasserver.testing.factory import factory` then `for i in range(0, 10): factory.make_Notification(admins=True)`).
- Load a React page.
- The notifications should each have an x icon.
- Clicking the icon should dismiss the notification.

## Fixes

Fixes: https://github.com/canonical-web-and-design/maas-squad/issues/2060.


huwshimi:
CI is passing now.

ui/src/app/base/components/NotificationGroup/NotificationGroup.test.tsx
```diff
@@ -143,55 +118,39 @@ describe("NotificationGroup", () => {
...
+    const notifications = [notificationFactory({ dismissable: true })];
 
     const wrapper = mount(
       <Provider store={store}>
         <NotificationGroup notifications={notifications} type="negative" />
       </Provider>
     );
 
-    wrapper.find("button[data-test='action-link']").simulate("click");
+    wrapper.find("button.p-icon--close").simulate("click");
```
squidsoup, line 129: Seems a shame to revert using `data-test` here, was that unavoidable? Best to avoid using classes for selectors if possible.

ui/src/app/base/components/NotificationGroup/NotificationGroup.tsx
```diff
@@ -0,0 +1,107 @@
...
+import { useDispatch } from "react-redux";
+import { Button, Notification } from "@canonical/react-components";
+import type { Dispatch } from "redux";
+
+import { capitaliseFirst } from "app/utils";
+import { useVisible } from "app/base/hooks";
+import { notification as notificationActions } from "app/base/actions";
+import type { Notification as NotificationType } from "app/store/notification/types";
+
+export const notificationTypes = {
```
squidsoup, line 14: Can this be an `enum`?

I think the convention would be to have something like:

```typescript
enum NotificationTypes = {
Caution = "caution",
...
}
```

ui/src/app/base/components/NotificationGroup/NotificationGroup.tsx
```diff
@@ -0,0 +1,107 @@
...
+  });
+};
+
+const dismiss = (id: NotificationType["id"], dispatch: Dispatch) => {
+  dispatch(notificationActions.delete(id));
+};
+
+type Props = {
+  notifications: NotificationType[];
+  type: "caution" | "negative" | "positive" | "information";
```
squidsoup, line 35: We have a type for this in `types`, `MessageType`.

ui/src/app/base/components/NotificationGroup/NotificationGroup.test.tsx
huwshimi: Yeah that's a part of react-components. I think using classes is going to be unavoidable in a lot of cases.

ui/src/app/base/components/NotificationGroup/NotificationGroup.tsx
huwshimi: Now that I look at this again I've noticed it isn't even used anywhere!

ui/src/app/base/components/NotificationGroup/NotificationGroup.tsx
huwshimi: So we do... thanks!

PR #1463, opened 2020-08-10
Add LP to GH issues script and cron

anthonydillon:
## Done
- Created a scripts directory to hold the bug mitigator script
- Wrote a script that gets all LP bugs tagged "ui" which does not have a linked bug tracker with maas-ui
- Add an action to run the script overnight

## QA
Check the code. The action is likely to fail due to a lack of an LP bot account which I will request.

_If the bot takes a few days, I will disable the action but I want to see it run once to check the action itself_

## Fixes

Fixes: https://github.com/canonical-web-and-design/MAAS-squad/issues/2047.


scripts/lp-bugs-to-gh-issues.py
```diff
@@ -0,0 +1,77 @@
...
+from github import Github
+
+
+class GithubApi:
+    def __init__(self):
+        self.repo_name = None
+        self.g = None
+
+    def set_env(self, token, repo_name):
+        self.set_token(token)
```
nottrobin, line 13: Why isn't the token passed on `__init__`?

scripts/lp-bugs-to-gh-issues.py
```diff
@@ -0,0 +1,77 @@
+#!/usr/bin/python3
+
+from launchpadlib.launchpad import Launchpad
+from github import Github
+
+
+class GithubApi:
```
nottrobin, line 7: This whole class looks like a whole load of overengineering - it looks to me like Github gives you everything you need:

``` python3
class GithubApi():
   # ... 30ish more lines

github_token = os.getenv("GITHUB_TOKEN")
github_repo = "canonical-web-and-design/maas-ui"
github_session = GithubApi()
github_session.set_env(github_token, github_repo)

# ...

github_session.create_issue(bug.title, issue_body)
```

Should be simply:

``` python3
github = Github(os.getenv("GITHUB_TOKEN"))
repo = github.get_repo("canonical-web-and-design/maas-ui")
repo.create_issue(bug.title, issue_body)
```

Removing about 35 lines of code.

scripts/lp-bugs-to-gh-issues.py
```diff
@@ -0,0 +1,77 @@
...
+
+# The repository to add this issue to
+status = ["New"]
+launchpad = Launchpad.login_with("Canonical web team stats", "production")
+project = launchpad.projects["maas"]
+ui_project = launchpad.projects["maas-ui"]
+
+
+def generate_open_bugs():
+    for task in project.searchTasks(status=status, tags=["ui"]):
```
nottrobin, line 47: You define this variable outside this function, just for use once within this function. It seems less clear. You should just write:

```suggestion
    for task in project.searchTasks(status=["New"], tags=["ui"]):
```

scripts/lp-bugs-to-gh-issues.py
```diff
@@ -0,0 +1,77 @@
...
+    watch = bug.addWatch(
+        bug_tracker=ui_project.bug_tracker, remote_bug=issue.number
+    )
+    ui_task.bug_watch = watch
+    ui_task.lp_save()
+
+
+def get_username(owner_link):
+    # Get the username and return the display name
+    return owner_link.replace("https://api.launchpad.net/1.0/~", "")
```
nottrobin, line 65: I get that you're trying to provide friendly names for blocks of code with your functions here, but I don't think single-line functions actually help readability, because for every function you have to hunt down where the function is to find out what's actually going on.

I think this function should be replaced by a simple variable assignment, which should provide the same clarify of naming:

```suggestion
    username = owner_link.replace("https://api.launchpad.net/1.0/~", "")
```

scripts/lp-bugs-to-gh-issues.py
```diff
@@ -0,0 +1,77 @@
...
+
+print("Getting ui bugs")
+count = 0
+for bug in generate_open_bugs():
+    issue_body = f"Bug originally filed by {get_username(bug.owner_link)} at {bug.web_link}\n\n{bug.description}"
+    issue = github_session.create_issue(bug.title, issue_body)
+    create_bug_task(github_repo, issue, bug)
+    print(issue_body)
+    count += 1
+    break
```
nottrobin, line 76: Why `break`? Doesn't this mean you'll just handle the first bug and then stop? In which case, why the `for` loop?

scripts/lp-bugs-to-gh-issues.py
anthonydillon: Awesome, much smaller :+1:

scripts/lp-bugs-to-gh-issues.py
anthonydillon: Good point, done

scripts/lp-bugs-to-gh-issues.py
anthonydillon: Yep, I created the function because I though I needed another call to the User object but didn't in the end.

scripts/lp-bugs-to-gh-issues.py
anthonydillon: Exactly, but only for the first run as a test so there are more to test after but I will remove and delete issues if I need more tests.

scripts/lp-bugs-to-gh-issues.py
```diff
@@ -0,0 +1,38 @@
+#!/usr/bin/python3
+
+from launchpadlib.launchpad import Launchpad
+from github import Github
+
+github_token = Github(os.getenv("GITHUB_TOKEN"))
+github = Github(github_token)
```
nottrobin, line 7: Think you messed this up a bit.

```suggestion
github = Github(os.getenv("GITHUB_TOKEN"))
```

.github/workflows/lp-to-gh-issues.yml
```diff
@@ -0,0 +1,19 @@
...
+  schedule:
+    - cron:  '0 22 * * *'
+
+jobs:
+  lp-to-gh-issues:
+    runs-on: ubuntu-latest
+
+    steps:
+      - uses: actions/checkout@v2
+      - name: Set up Python
```
nottrobin, line 12: Surprised that you need this = doesn't `ubuntu-latest` come with python 3.8 already? We've never had to set it up manually in our other actions - e.g. [here](https://github.com/canonical-web-and-design/ubuntu.com/blob/master/.github/workflows/pr.yaml#L92)

anthonydillon: I think this is to ping the version of python used. Got it from here: https://docs.github.com/en/actions/language-and-framework-guides/using-python-with-github-actions

nottrobin: In that case they do it to run with all sorts of python versions. I really don't think it's necessary. I know this is merged, but I'd appreciate it if you would remove this if you get a chance.

PR #1476, opened 2020-08-17
Add Python action to set up with version 3.8

anthonydillon:
## Done
- Add "Python set up" step from LP to GH issue mitigator action

## QA
Check the code.


squidsoup:
@tbille @anthonydillon just fyi, we're trying to use [conventional commits](https://www.conventionalcommits.org/en/v1.0.0/) for this project now (see Contributing in the README).

PR #1481, opened 2020-08-18
fix(ui): adjust spacing around deploy options

huwshimi:
## Done

- Fix the spacing around the deploy options.

### MAAS deployment

- [Bolla](/HACKING.md#bolla)

### Running the branch

You can run this branch by:

### QA steps

- Visit the machine list, check a deployable machine and open the deploy form via the take action menu.
- Check that there is a nice amount of space between the two checkbox options.
- Reduce the size of your window until the labels wrap onto a new line. The wrapping text should be indented to match the label (and clear the checkbox).
- Now visit a deployable machine's details page and open the deploy form via the take action menu.
- Check that there is a nice amount of space between the two checkbox options.
- Reduce the size of your window until the labels wrap onto a new line. The wrapping text should be indented to match the label (and clear the checkbox).

## Fixes

Fixes: canonical-web-and-design/MAAS-squad#2040.

### DIFF
--- a/legacy/src/app/partials/nodedetails/deploy-options.html
+++ b/legacy/src/app/partials/nodedetails/deploy-options.html
@@ -4,9 +4,7 @@
     <p>Customise options</p>
   </div>
   <div class="col-9">
-    <div
-      class="p-form-group p-form-validation"
-    >
+    <div class="p-form-group p-form-validation">
       <div class="p-form__control u-clearfix">
         <input
           id="installKVM"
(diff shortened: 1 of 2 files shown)


Caleb-Ellis (changes requested):
This doesn't seem right:




Do the checkboxes need to be inline blocks?

huwshimi:
@Caleb-Ellis Whoops! Fixed.

legacy/src/app/partials/nodedetails/deploy-options.html
```diff
@@ -4,78 +4,65 @@
...
-            </p>
-          </div>
-        </div>
-      </li>
-      <li class="p-inline-list__item">
-        <a
-          class="p-link--external"
-          href="https://maas.io/docs/kvm-introduction"
-          target="_blank"
+        <p
```
Caleb-Ellis, line 32: Do you think we can remove this `p` block? It's not in the react version and it seems redundant to show a warning when the checkbox warns the same thing.

Caleb-Ellis (approved):
LGTM! Just one comment for consideration.

legacy/src/app/partials/nodedetails/deploy-options.html
huwshimi: I think you're right, pretty sure it was removed from the react version at some point, but not removed here.

PR #1486, opened 2020-08-19
fix(scripts): Authenticate with Launchpad via credentials file

anthonydillon:
## Done
- Create a temp file with the contents of the environment variable: LAUNCHPAD_CREDENTIALS
- Pass that file as the credentials when initialising a Launchpad session

## QA
- Check the code


jkfran (approved):
👍🏻 Nice!

PR #1512, opened 2020-08-25
fix(scripts): Find all "ui" bugs in all statuses

anthonydillon:
## Done
- Updated the status filter from the bug search to include all bugs with the status New, Confirmed and Triaged.

## QA
- Check the code.

### DIFF
--- a/scripts/lp-bugs-to-gh-issues.py
+++ b/scripts/lp-bugs-to-gh-issues.py
@@ -22,7 +22,7 @@ ui_project = launchpad.projects["maas-ui"]
 
 
 def generate_open_bugs():
-    for task in project.searchTasks(status=["New"], tags=["ui"]):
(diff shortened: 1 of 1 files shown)


sparkiegeek:
The code is fine, and can land but not sure it'll do what you want.

Which bugs should be duplicated to GitHub?

```
In [35]: all_tasks = lp.projects['maas'].searchTasks(tags=['ui'])              

In [36]: Counter(task.status for task in all_tasks)                            
Out[36]: 
Counter({'Fix Committed': 8,
         'Triaged': 75,
         'Incomplete': 1,
         'In Progress': 1,
         'Confirmed': 2,
         'New': 11})
```

anthonydillon:
Good point, we probably only care about New, Confirmed and Triaged. Everything else has ben handled.

sparkiegeek (approved):

PR #1579, opened 2020-08-27
feat(ui): build a doughnut chart component

huwshimi:
## Done

- Create a doughnut chart for the KVM pages.

### MAAS deployment

- [Bolla](/HACKING.md#bolla)

### Running the branch

You can run this branch by:

### QA steps

- Visit a kvm details page.
- You should see a doughnut chart.
- Hovering the doughnut should raise a section and show a tooltip.

## Fixes

Fixes: canonical-web-and-design/maas-squad#2073

### DIFF
--- a/ui/src/app/base/components/DoughnutChart/DoughnutChart.tsx
+++ b/ui/src/app/base/components/DoughnutChart/DoughnutChart.tsx
@@ -48,7 +48,7 @@ export const DoughnutChart = ({
 }: Props): JSX.Element => {
   const [showTooltip, setShowTooltip] = useState(null);
   const [tooltipStyle, setTooltipStyle] = useState(null);
-  const removeListener = useRef(() => {});
+  const removeListener = useRef<() => void>(null);
 
   useEffect(
     () => () => {
--- a/ui/src/app/base/components/DoughnutChart/DoughnutChart.tsx
+++ b/ui/src/app/base/components/DoughnutChart/DoughnutChart.tsx
@@ -1,0 +1,195 @@
+import { nanoid } from "@reduxjs/toolkit";
+import React, { useRef, useEffect, useState } from "react";
+
+type Segment = {
+  /**
+   * The colour of the segment.
+   */
+  color: string;
+  /**
+   * The segment tooltip.
+   */
+  tooltip?: string;
+  /**
+   * The segment length.
+   */
+  value: number;
+};
+
+type Props = {
+  /**
+   * The label in the centre of the doughnut.
+   */
+  label?: string;
+  /**
+   * The width of the segments when hovered.
+   */
+  segmentHoverWidth: number;
+  /**
+   * The width of the segments.
+   */
+  segmentWidth: number;
+  /**
+   * The doughnut segments.
+   */
+  segments: Segment[];
+  /**
+   * The size of the doughnut.
+   */
+  size: number;
+};
+
+export const DoughnutChart = ({
+  label,
+  segmentHoverWidth,
+  segmentWidth,
+  segments,
+  size,
+}: Props): JSX.Element => {
+  const [showTooltip, setShowTooltip] = useState(null);
+  const [tooltipStyle, setTooltipStyle] = useState(null);
+  const removeListener = useRef(() => {});
+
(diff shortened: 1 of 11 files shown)


ui/src/app/base/components/DoughnutChart/DoughnutChart.tsx
```diff
@@ -0,0 +1,195 @@
...
+          )} ${segmentLength.toFixed(2)} ${remainingSpace.toFixed(2)}`,
+        }}
+        // Rotate the segment so that the segments start at the top of
+        // the chart.
+        transform={`rotate(-90 ${radius},${radius})`}
+      />
+    );
+  });
+  return (
+    <div className="p-tooltip--right">
```
Caleb-Ellis, line 157: I think a `className` prop would be good - then you could give it `"u-align--center"` which would make the NUMA view look like the design

ui/src/app/kvm/components/KVMResourcesCard/KVMResourcesCard.tsx
```diff
@@ -56,10 +57,40 @@ const KVMResourcesCard = ({
...
+          <DoughnutChart
+            label={`${ram.general.total + (ram.hugepage?.total || 0)}${
+              ram.general.unit
+            }`}
+            segmentHoverWidth={20}
+            segmentWidth={15}
+            segments={[
+              {
+                color: "#06C",
+                tooltip: `General allocated ${ram.general.allocated}`,
```
Caleb-Ellis, line 69: The tooltip values are missing units.

Now that I'm looking at it, it probably makes sense to just have one unit for RAM instead of one for general and one for hugepage, but that's neither here nor there because it'll be replaced by real data anyway. :man_shrugging:

ui/src/app/kvm/components/KVMResourcesCard/KVMResourcesCard.tsx
```diff
@@ -56,10 +57,40 @@ const KVMResourcesCard = ({
...
           <h4 className="p-heading--small">RAM</h4>
-          <div className="kvm-resources-card__ram-chart">
-            {ram.general.total}
-            {ram.general.unit}
-          </div>
+          <DoughnutChart
+            label={`${ram.general.total + (ram.hugepage?.total || 0)}${
+              ram.general.unit
+            }`}
+            segmentHoverWidth={20}
```
Caleb-Ellis, line 64: Think it's worth defining all these numbers as `const`s?

Caleb-Ellis (approved):
This looks awesome! I just have some minor suggestions. There are also a couple of minor design things (chart positioning at different breakpoints; inner radius stays constant when hovering over a segment, i.e. it only expands outward) but they can just be addressed in follow ups.

ui/src/app/kvm/components/KVMResourcesCard/KVMResourcesCard.tsx
```diff
@@ -56,10 +57,40 @@ const KVMResourcesCard = ({
...
+              },
+              ...(ram.hugepage
+                ? [
+                    {
+                      color: "#0E8420",
+                      tooltip: `Hugepage allocated ${ram.hugepage.allocated}`,
+                      value: ram.hugepage.allocated,
+                    },
+                    {
+                      color: "#56a863",
```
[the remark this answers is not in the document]
huwshimi: Thanks, how about `POSITIVE_MID` for the new one?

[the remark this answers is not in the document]
Caleb-Ellis: That works!

ui/src/app/kvm/components/KVMResourcesCard/KVMResourcesCard.tsx
huwshimi: Yeah probably, but might be worth waiting until this chart is used a second time and see if they stay the same.

squidsoup:
This is awesome Huw!

huwshimi:
I've updated this to only grow outwards on hover and be centred on the card.

PR #1599, opened 2020-09-03
fix(legacy): replace old version-specific docs link with version agnostic link

sparkiegeek:


squidsoup (approved):
Nice, thanks @sparkiegeek!

sparkiegeek:
image
Oops, think i shouldn't have copied that small-text class

PR #1643, opened 2020-09-16
Yarn cache/parallel jobs

huwshimi:
## Done

- Cache and restore node_modules for ci runs.
- Split lint, test and build into separate jobs so that they run in parallel.

## QA

- The jobs should pass in CI and you should see each job listed separately.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -9,10 +9,9 @@ on:
 jobs:
   lint:
     name: Lint
-    runs-on: ${{ matrix.os }}
+    runs-on: ubuntu-latest
     strategy:
       matrix:
-        os: [ubuntu-18.04]
         node-version: [12.x]
     steps:
       - uses: actions/checkout@v2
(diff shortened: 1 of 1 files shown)


.github/workflows/test.yml
```diff
@@ -7,25 +7,80 @@ on:
     branches:
       - master
 jobs:
-  build:
-    name: Test, build and lint
+  lint:
+    name: Lint
     runs-on: ${{ matrix.os }}
```
squidsoup, line 12: While you're here, we're not really making the best us of the matrix strategy, I think we could just set `runs-on: ubuntu-latest` and just preserve `node-version` in the matrix, dropping `os`. What do you think?

squidsoup (approved):
Looks great, a big improvement! Just one thought on dropping `os` from `matrix`.

.github/workflows/test.yml
huwshimi: Sure, I'll update it.

Issue #1703, opened 2020-09-28
TypeError: event is undefined


huwshimi:
https://sentry.is.canonical.com/canonical/maas-ui/issues/10652/?referrer=github_plugin

```
TypeError: event is undefined
  at NodeDetailsController/$scope.getEventText (/MAAS/r/1.6f85774a2928f988cdb6.js:34694:46)
  at fn (/MAAS/r/1.6f85774a2928f988cdb6.js%20line%2011536%20%3E%20Function:4:348)
  at expressionInputWatch (/MAAS/r/1.6f85774a2928f988cdb6.js:11629:129)
  at $digest (/MAAS/r/1.6f85774a2928f988cdb6.js:12439:35)
  at $apply (/MAAS/r/1.6f85774a2928f988cdb6.js:12663:163)
...
(7 additional frame(s) were not displayed)
```

PR #1707, opened 2020-09-28
fix(legacy): handle events without data

huwshimi:
## Done

- Gracefully handle websocket events without data.

## QA

No QA. I wasn't able to reproduce this but it seems to be connected to dropped websocket connections. This should at least handle the bad messages.

## Fixes

Fixes: canonical-web-and-design/maas-ui#1703


Caleb-Ellis (approved):
Seems good to me!

PR #1716, opened 2020-09-29
fix(legacy): prevent removal of default commissioning scripts from machine details form

### LINKED ISSUE #1668: Builtin commissioning scripts should be unable to be deselected on the node-details page
Bug originally filed by ltrager at https://bugs.launchpad.net/bugs/1896174

LP:1892238 solved this for the nodes listing page but when I use the commissioning form on the node details page I can still deselect builtin commissioning scripts.

Caleb-Ellis:
## Done

- The third-party ngTagsInput library doesn't allow disabling a subset of tags, so I've used a terrible css hack to prevent default commissioning scripts from being removed in the legacy form.

### MAAS deployment

- [Bolla](/HACKING.md#bolla)

### Running the branch

You can run this branch by:

### QA steps

- Go to the machine details page of a machine that can be commissioned and open the Commission form
- Check that the default commissioning scripts show up by default, they cannot be removed and they are sorted alphanumerically
- Add a non-default script and check that you can then remove it
- Open the Test action form and check that it has remained unchanged

## Fixes

Fixes #1668


legacy/src/app/directives/script_select.js
```diff
@@ -17,6 +17,12 @@ function filterScriptsByParam(scripts, param) {
   });
 }
 
+const sortScripts = (a, b) => {
```
squidsoup, line 20: This is fine (given we're killing this soon), but just so you know there's a nicer way to do this today with [localCompare](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/localeCompare#Browser_compatibility):

```
['qux', 'bar', 'foo'].sort((a, b) => a.localeCompare(b))
> ["bar", "foo", "qux"]
```

squidsoup (approved):
+1 for filthy css hack

legacy/src/app/directives/script_select.js
Caleb-Ellis: I'll remember for next time!

Issue #987, opened 2020-04-16
Please don't use "Ipmi" (capitalized this way) in machines overview


lilyvidenova:
[LP bug 1872664](https://bugs.launchpad.net/maas/+bug/1872664)

Hi,

Running MAAS snap 2.7.0-8235-g.fea3a1678

In the machines overwiew, I see that the power type is written as "Ipmi". Please use either "IPMI" or "ipmi" (first form preferred) - in the same way that you use "PXE" and not "Pxe".

Thanks

PR #1742, opened 2020-10-07
fix(legacy): show the full power type name

huwshimi:
## Done

- Show the full power type name on controller and machine details

### MAAS deployment

- [Bolla](/HACKING.md#bolla)

### Running the branch

You can run this branch by:

### QA steps

- Visit a machine and controller summary. Both should display the full power type name (specifically IPMI should be capitalised).

## Fixes

Fixes: canonical-web-and-design/maas-ui#987

### DIFF
--- a/legacy/src/app/controllers/node_details.js
+++ b/legacy/src/app/controllers/node_details.js
@@ -5,7 +5,11 @@
  */
 import angular from "angular";
 
-import { generateLegacyURL, generateNewURL } from "@maas-ui/maas-ui-shared";
+import {
+  extractPowerType,
+  generateLegacyURL,
+  generateNewURL,
+} from "@maas-ui/maas-ui-shared";
 import { HardwareType, NodeTypes } from "../enum";
 
 // Convert MiB to GiB value if over 1024 and round to 4 significant figures.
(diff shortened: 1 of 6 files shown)


Caleb-Ellis (commented):
We had it like this a while ago, but a lot of the power type descriptions are really long so would look awkward in the cell, like "American Power Conversion (APC) PDU". Maybe we should have a name map that translates the name into something nice for the UI (e.g. "ipmi" => "IPMI") with the description as a fallback?

huwshimi:
> We had it like this a while ago, but a lot of the power type descriptions are really long so would look awkward in the cell, like "American Power Conversion (APC) PDU". Maybe we should have a name map that translates the name into something nice for the UI (e.g. "ipmi" => "IPMI") with the description as a fallback?

What about if we extract matching strings from the descriptions? Then we'd end up with cases like this:

openbmc in "OpenBMC Power Driver" -> OpenBMC
ipmi in "IPMI" -> IPMI

And for any case where there is no match we will fall back to the power type key.

Caleb-Ellis:
Sounds good!

huwshimi:
@Caleb-Ellis updated.

Caleb-Ellis (approved):
LGTM!

PR #1750, opened 2020-10-09
perf(ui): load machines in larger batches after the first request

huwshimi:
## Done
Thinking about this issue: https://discourse.maas.io/t/landing-page-takes-multiple-minutes-to-load-2000-nodes/3302 it seems that the number of websocket round trips is a large part of the issue, especially on high latency networks. This branch reduces the number of round trips by fetching 25 machines then batches of 100. For 2000 machines this drops the round trips from 40 to 11 (on my local setup this reduced the total time to load 2000 machines by nearly half).

### MAAS deployment

- [Bolla](/HACKING.md#bolla)

### Running the branch

You can run this branch by:

### QA steps

- Add a heap of machines, using the dev maas you can do:
```console
make harness
from maasserver.testing.factory import factory
for i in range(0, 2000):  factory.make_Node()
```
- Load the machine list. Notice that the first batch of machines will be 25 then should load in batches of 100.
- As it's loading check that the ui rendering isn't slowed down too much as it loads in each batch.


ui/src/app/base/sagas/websockets.js
```diff
@@ -173,6 +173,10 @@ export function* handleBatch({ request_id, result }) {
       deleteBatchRequest(request_id);
       // Set the next batch to start at the last id we received.
       let nextBatch = { ...batchRequest };
+      if (nextBatch?.meta?.subsequentLimit) {
```
squidsoup, line 176: I think this could use a comment to explain the fetching strategy.

squidsoup (commented):
@huwshimi this certainly seems like a sensible strategy, but perhaps we should just run it by core first as increasing the batch size will increase memory usage on the server. Can we profile this?

huwshimi:
@squidsoup thanks, I've sent the email and will wait for a reply.

sparkiegeek:
@albertodonato please share your findings with @huwshimi here

squidsoup (approved):
Spoke to ack last night, and he feels confident that this strategy is an improvement. Thanks Huw

huwshimi:
> Spoke to ack last night, and he feels confident that this strategy is an improvement. Thanks Huw

Great, thanks!

PR #1788, opened 2020-10-20
fix(cypress): fix cypress tests rendering incorrect navigation on new MAAS installs

### LINKED ISSUE #1795: Fix broken Cypress navigation tests

Caleb-Ellis:
## Done

- In the Cypress tests, the intro is skipped by setting a cookie. This allowed access to the rest of the app but did not update the navigation. This was only apparent on fresh MAAS installs that had not yet completed the setup or user intros.
- Fixed screenshot upload and removed video upload

### QA steps

- Checkout master, run the UI (`yarn start`) and run Cypress in separate terminals (`yarn cypress-open`).
- Click "Run all specs" in the Cypress interactive runner and confirm that the tests regarding navigation all fail, except for user preferences.
- Checkout this branch
- Create a new user and update the `integration/cypress.json` env variables with their credentials.
- Run all Cypress specs and check that all of the tests regarding navigation pass. There will be some failed tests due to MAAS having gone through the setup intro already and the UI running in development. A full, "proper" QA would involve building a new MAAS snap and running Cypress on that, but I think checking that the navigation tests pass should be enough.

## Fixes

Fixes #1795

### DIFF
--- a/.github/workflows/cypress.yml
+++ b/.github/workflows/cypress.yml
@@ -28,7 +28,7 @@ jobs:
         with:
           config: baseUrl=${{env.MAAS_URL}},pageLoadTimeout=100000
           install: false
-          spec: cypress/integration/no-users/login.spec.ts
+          spec: "**/no-users/*.spec.ts"
           wait-on: "${{env.MAAS_URL}}/MAAS/r/machines"
           working-directory: integration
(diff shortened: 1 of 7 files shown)


huwshimi (approved):
This seems to work so let's land it and see how it goes.

Issue #1794, opened 2020-10-21
Logging out while using external auth displays normal login form


squidsoup:
**Describe the bug**
Logging out while using external auth displays the normal login form, instead of the external auth form with button. Refreshing the page **does** however correctly display the external auth form.

**maas-ui version**
tip

**MAAS version**
tip

**LP issue**
Is this an [issue on launchpad](https://bugs.launchpad.net/maas)? If so, please provide a bug is in the form lp#number.

**To Reproduce**
1. Configure external auth with RBAC
2. Login to MAAS
3. Logout
4. See wrong login form

**Expected behavior**
The external auth form should display.

**Screenshots**

**Desktop (please complete the following information):**

**Additional context**

PR #1796, opened 2020-10-21
fix(ui): return to the correct login screen when logging out with external auth

huwshimi:
## Done

- When using external auth return to the external auth login screen after logging out.
- Update App to typescript.

### MAAS deployment

- [Bolla](/HACKING.md#bolla)

### Running the branch

You can run this branch by:

### QA steps

- Setup your maas to use external auth (ask me).
- Log in and then log out, you should be returned to the external auth login screen.

## Fixes

Fixes: canonical-web-and-design/maas-ui#1794.


squidsoup (approved):
Thanks Huw, did manage to QA this without the error, which does appear to occur intermittently on master, so I don't think this branch is at fault.

PR #1798, opened 2020-10-21
fix(legacy): fix the legacy navigation css

huwshimi:
## Done

- Restore the legacy navigation css so that it looks correct when the hardware menu is shown.

### MAAS deployment

- [Bolla](/HACKING.md#bolla)

### Running the branch

You can run this branch by:

### QA steps

- Navigate to a legacy page.
- Resize your window until the hardware menu is shown.
- Check that the items in the menu don't also appear in the main nav.

Fixes https://github.com/canonical-web-and-design/maas-ui/issues/1799

### DIFF
--- a/legacy/src/scss/_patterns_navigation.scss
+++ b/legacy/src/scss/_patterns_navigation.scss
@@ -1,0 +1,27 @@
+@mixin maas-navigation {
+  $breakpoint-hardware-menu: 1200px;
+
+  .hardware-menu {
+    display: none;
+
+    @media (min-width: $breakpoint-navigation-threshold) and (max-width: $breakpoint-hardware-menu) {
+      display: inherit;
+    }
+  }
+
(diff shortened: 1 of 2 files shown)


anthonydillon (approved):
LGTM :+1: for a quick fix. I think these styles should reside in the shared part of this application but is out of scope for this fix.

I wondered if instead of having duplicated styling/functionality which needs to be manually kept in sync. You might want to just drop:
```scss
@import "../../../ui/src/scss/_patterns_navigation";
@include maas-navigation;
```
in `legacy/src/scss/build.scss`?

huwshimi:
Thanks @anthonydillon, these were styles I removed earlier in the week (but shouldn't have). For some reason they're a bit different to the ui nav styles, I'll add a card to follow up and dedupe if we can.

PR #1827, opened 2020-10-29
feat(ui): add action/selector/reducer for setting active machines.

### LINKED ISSUE #1820: UPDATE_MODEL_NOTIFY removes details parameters from the model in state
We need to change the `UPDATE_MODEL_NOTIFY` reducers so that they don't completely overwrite the model in state with the non-detail version.

The `set_active` websocket handler for many of the models seems to address this. i.e. it lets the region know that the item needs all parameters when updated.

Reminder to update `useMachineActionForm` to use `active` state

Caleb-Ellis:
## Done

- Added action/selectors/reducers for setting active machines. Setting a machine as active lets the server know to send all the data of that machine whenever it is updated, otherwise it only sends the data required for the machine list.
- Updated machine reducer tests to use the test factories

### MAAS deployment

- [Bolla](/HACKING.md#bolla)

### Running the branch

You can run this branch by:

### QA steps

- Navigate to a machine details page and check that `machine.active` is set to the machine's system_id.
- Perform an action on the machine (any should do) and check that the machine is updated with all of the details data intact.
- Navigate back to the machine list and check that `machine.active` is set back to `null`.

## Fixes

Fixes #1820

### DIFF
--- a/ui/src/app/store/machine/selectors.test.ts
+++ b/ui/src/app/store/machine/selectors.test.ts
@@ -57,6 +57,26 @@ describe("machine selectors", () => {
     expect(machine.saved(state)).toEqual(true);
   });
 
+  it("can get the active machine's system ID", () => {
+    const state = rootStateFactory({
+      machine: machineStateFactory({
+        active: "abc123",
+      }),
+    });
+    expect(machine.activeID(state)).toEqual("abc123");
+  });
+
+  it("can get the active machine", () => {
+    const activeMachine = machineFactory();
+    const state = rootStateFactory({
+      machine: machineStateFactory({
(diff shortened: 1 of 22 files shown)


ui/src/app/store/machine/selectors.test.ts
```diff
@@ -79,17 +99,6 @@ describe("machine selectors", () => {
     expect(machine.getById(state, "909")).toStrictEqual(items[1]);
   });
 
-  it("can get the error state", () => {
```
Caleb-Ellis, line 82: Duplicate test

huwshimi (approved):
Nice, thanks @Caleb-Ellis

PR #1828, opened 2020-10-29
feat(ui): add actions/reducers/selectors for setting an active pod

### LINKED ISSUE #1826: Set KVM as active when navigating to KVM details page
If you edit a KVM's config and THEN try and compose a machine the KVM object is missing `attached_vlans` which is used for some of the interface table logic. We need to set the KVM as active so that whenever it's updated the region sends all the data.

Caleb-Ellis:
## Done

- Added action/selectors/reducers for setting active pods.
- Updated pod reducer tests to use the test factories.

### MAAS deployment

- [Bolla](/HACKING.md#bolla)

### Running the branch

You can run this branch by:

### QA steps

- Navigate to a KVM details page and check that `pod.active` is set to the pod's id.
- Edit the KVM configuration and check that the pod does not lose the `attached_vlans` parameter in state.
- Navigate back to the KVM list and check that `pod.active` is set back to `null`.

## Fixes

Fixes #1826


huwshimi (approved):
Looks good!

Issue #1842, opened 2020-11-04
Handle per-machine action errors


huwshimi:
When performing actions from machine details the errors are stored at the top level errors key, but we need a way to store/associate the error with the machine that performed the action.

These errors should then be displayed on the machine details page(s) and cleaned up when navigating away.

PR #1904, opened 2020-11-19
refactor(ui): use generic state type for model states

huwshimi:
## Done

- Update the model state types to use the generic state.

## QA

n/a

## Fixes

First branch for: canonical-web-and-design/maas-ui#1842.
Fixes: canonical-web-and-design/maas-ui#1905.


Caleb-Ellis (approved):
Nice!

Issue #1878, opened 2020-11-13
Add lint for type imports


huwshimi:
https://github.com/typescript-eslint/typescript-eslint/blob/master/packages/eslint-plugin/docs/rules/consistent-type-imports.md

PR #1923, opened 2020-11-25
refactor(ui): add lint rule for type imports

huwshimi:
##  Done

- Add a lint rule to enforce type imports.
- Fix type import lint errors (using `lint --fix`).

## QA

n/a

## Fixes

Fixes: canonical-web-and-design/maas-ui#1878.

### DIFF
--- a/ui/.eslintrc.js
+++ b/ui/.eslintrc.js
@@ -53,6 +53,7 @@ module.exports = {
             ignoreRestSiblings: true,
           },
         ],
+        "@typescript-eslint/consistent-type-imports": 2,
         "import/namespace": "off",
         "import/no-named-as-default": 0,
         "import/order": [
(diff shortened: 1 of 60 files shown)


ui/.eslintrc.js
```diff
@@ -69,7 +70,7 @@ module.exports = {
                 group: "internal",
               },
             ],
-            "pathGroupsExcludedImportTypes": ["react"],
+            pathGroupsExcludedImportTypes: ["react"],
             "newlines-between": "always",
```
squidsoup, line 74: well, quite right, this isn't json, but should the next line be fixed too? `newlines-between`?

squidsoup (approved):
Nice! One small question about the driveby.

ui/.eslintrc.js
huwshimi: That was just prettier doing its thing. For `newlines-between` the key contains a hyphen so has to have quotes to be valid.

PR #1928, opened 2020-11-26
refactor(ui): store errors with id and action

huwshimi:
## Done

- Associate machine errors with the machine and the action that performed the action.

### MAAS deployment

- [Bolla](/HACKING.md#bolla)

### Running the branch

You can run this branch by:

### QA steps

- Go to the machine list and select a few machines and perform an action that returns an error (e.g. open the tag form and create a tag with a symbol in the name), the error should be displayed.
- Go to a machine details page and again perform an action that returns an error, the error should be displayed.

## Fixes

Fixes: canonical-web-and-design/maas-ui#1842.


ui/src/app/machines/components/ActionFormWrapper/MarkBrokenForm/MarkBrokenForm.tsx
```diff
@@ -28,11 +28,13 @@ type Props = {
...
-  const { machinesToAction, processingCount } = useMachineActionForm(
-    NodeActions.MARK_BROKEN
-  );
+  const {
+    errors: machineErrors,
+    machinesToAction,
+    processingCount,
+  } = useMachineActionForm(NodeActions.MARK_BROKEN);
+  const errors =
+    Object.keys(machineErrors || {}).length > 0 ? machineErrors : null;
```
Caleb-Ellis, line 37: Do you think this step is necessary? `formatErrors` should already get run in `ActionForm.tsx`. Anyway it wasn't introduced in this PR but if you think it's an easy driveby fix go ahead

ui/src/app/machines/hooks.tsx
```diff
@@ -35,6 +36,7 @@ export const useToggleMenu = (
 export const useMachineActionForm = (
   actionName: MachineActionName | "check-power"
 ): {
+  errors: MachineState["eventErrors"][0]["error"][];
```
Caleb-Ellis, line 39: `errors` seems like a misnomer here if it can only ever be one error

ui/src/app/machines/hooks.tsx
```diff
@@ -44,9 +46,24 @@ export const useMachineActionForm = (
   // If in the machine details view, the machine is not in selected state so
   // instead we use the regular selector.
   const processingMachines = useSelector(
-    machineSelectors[activeMachine ? action.status : `${action.status}Selected`]
+    machineSelectors[
+      selectedMachines ? action.status : `${action.status}Selected`
```
Caleb-Ellis, line 50: Looks like the logic has flipped here and should be flipped back.

ui/src/app/store/machine/selectors.ts
```diff
@@ -146,6 +146,56 @@ const selected = createSelector(
...
+ * Select the event errors for all machines.
+ * @param state - The redux state.
+ * @returns The event errors.
+ */
+const eventErrors = (state: RootState): MachineState["eventErrors"] =>
+  state.machine.eventErrors;
+
+/**
+ * Select the event errors for a machine or machines.
+ * @param id - A system ID for a machine.
```
Caleb-Ellis, line 159: Should be `ids`

Caleb-Ellis (approved):
LGTM! Just some minor suggestions, but otherwise seems to work really well!

ui/src/app/machines/components/ActionFormWrapper/MarkBrokenForm/MarkBrokenForm.tsx
huwshimi: Yes, quite right, I'll remove it.

ui/src/app/machines/hooks.tsx
huwshimi: You're right that we only return a single "error" (for one machine), but the server can return errors as a string, array or object of errors. So the "errors" could be multiple errors or just a single string...

At any rate the type here was wrong which I'll fix.

ui/src/app/machines/hooks.tsx
huwshimi: Strange, done.

PR #1996, opened 2020-12-10
refactor(ui): update selected state handling in available storage table

### LINKED ISSUE #1997: Update selected state handling in available storage table

Caleb-Ellis:
## Done

- Updated selected state handling in available storage table.
  - Now the entire storage device object is set in selected state. This is for the actions to come later which will need to know about more than just the id and type.
  - Added a `useEffect` which keeps the selected state from becoming stale, e.g. for when a disk is updated or deleted

### MAAS deployment

- [Bolla](/HACKING.md#bolla)

### Running the branch

You can run this branch by:

### QA steps

- Go to the storage tab of a machine in Ready or Allocated state
- Change the storage layout to blank
- Add a partition to one available disk
- Check that the header checkbox and individual row checkboxes work as expected
- Check the checkbox for the partition, then open the action dropdown and select "Remove partition..." and confirm. Note that the checkbox doesn't actually serve a purpose for this action, it's just so we can test selected state.
- The partition should be removed from the table, and should also be removed from selected state. It will have worked correctly if the header checkbox is unchecked.
- Create a mounted partition so it goes straight into the filesystems table
- Check the header checkbox to select everything, then unmount the filesystem
- The partition should move to the available table and the header checkbox should change to mixed selection state (`-` instead of a tick)

## Fixes

Fixes #1997

### DIFF
--- a/ui/src/app/machines/views/MachineDetails/MachineStorage/AvailableStorageTable/AvailableStorageTable.tsx
+++ b/ui/src/app/machines/views/MachineDetails/MachineStorage/AvailableStorageTable/AvailableStorageTable.tsx
@@ -1,4 +1,4 @@
-import { useState } from "react";
+import { useEffect, useState } from "react";
 
 import { Input, MainTable } from "@canonical/react-components";
 import classNames from "classnames";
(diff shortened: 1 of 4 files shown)


huwshimi (approved):
Works well!

PR #2007, opened 2020-12-15
refactor(ui): extract common FilesystemFields component from storage forms

### LINKED ISSUE #1970: Fix swap filesystem handling in storage
Swap filesystems are a special case that don't use a mount point.

Caleb-Ellis:
## Done

- Created `FilesystemFields` component which houses some common fields found in a few of the storage forms
- Made formik field values consistent between forms (so they're all `fstype` instead of `filesystemType`, which also matches the api)
- Fixed some incorrect behaviour with the filesystem fields:
  - Removed requirement for a mount point if a fstype is selected. This is only meant to be the case for special filesystems, which was the first of the filesystem forms to be built.
  - Changed the fields to disabled instead of hidden if a fstype is not selected. We're trying to move away from hiding elements.
  - Fixed handling of "swap" filesystems

### MAAS deployment

- [Bolla](/HACKING.md#bolla)

### Running the branch

You can run this branch by:

### QA steps

- Go to the storage tab of a Ready or Allocated machine
- Open the "Add partition" form for a disk - you should see the disabled form fields for mount point and mount options
- Check that selecting a filesystem enables the mount point and mount options fields
- Check that you can add a partition with a filesystem even if it doesn't have a mount point
- Unmount the filesystem and edit it
- Change the filesystem to "swap"
- Check that the mount point field is disabled with "none" as the placeholder

## Fixes

Fixes #1970


huwshimi:
I know it wasn't changed in this branch, but one thing I noticed was that when you're editing a partition the green button is "Edit partition" rather than "Save" in the legacy version. I think "Save" feels better there (I paused when I saw "Edit partition").

huwshimi (approved):

PR #2009, opened 2020-12-15
fix(legacy): stop setting gateway on gateway-less subnets (#1860)

vtapia:
## Done

Unify the behavior for both standard and relayed networks, populating the placeholder with the suggested gateway without setting the IP when the subnet does not have one configured.

## QA

- Configure a subnet without gateway
- Go to related VLAN and configure DHCP
- Verify that the gateway field has remained unchanged

## Fixes

Fixes: #1860 .

## Launchpad issue

lp#1903586

## Backports

2.8 and 2.7 are affected too


Caleb-Ellis (commented):
Looks good! Just need to update the `prepares provideDHCPAction on actionOptionChanged and populates suggested gateway` test in `src/app/controllers/tests/test_vlan_details.js`

Caleb-Ellis (approved):
Looks good to me! Thanks @vtapia

Issue #1974, opened 2020-12-04
Network tab - add link/alias rows


huwshimi:

huwshimi:
Also, confirm how alias rows should check if they are connected. The legacy code checks the member status:

https://github.com/canonical-web-and-design/maas-ui/blob/master/legacy/src/app/controllers/node_details_networking.js#L2468

However, it appears that only bonds and bridges can have members:

https://github.com/canonical-web-and-design/maas-ui/blob/master/legacy/src/app/controllers/node_details_networking.js#L408

PR #2037, opened 2021-01-07
feat(ui): add alias rows to the network table

huwshimi:
## Done

- Add the rows that display the links/aliases in the network tab of machine details.

### MAAS deployment

- [Bolla](/HACKING.md#bolla)

### Running the branch

You can run this branch by:

### QA steps

- View the network tab in legacy for a machine that can edit network (New, Ready, Allocated or Broken.)
- Create a Bond (if one doesn't exist already).
- use the action menu to add an alias.
- open the network tab for the same machine in the react client
- you should see the alias row and it should have the same data as the legacy page.

## Fixes

Fixes: #1974.

### DIFF
--- a/ui/src/app/machines/views/MachineDetails/MachineNetwork/NetworkTable/NetworkTable.tsx
+++ b/ui/src/app/machines/views/MachineDetails/MachineNetwork/NetworkTable/NetworkTable.tsx
@@ -19,20 +19,18 @@ import machineSelectors from "app/store/machine/selectors";
 import type {
   NetworkInterface,
   NetworkLink,
-  NetworkLinkInterface,
   Machine,
 } from "app/store/machine/types";
+import { NetworkInterfaceTypes } from "app/store/machine/types";
 import {
-  NetworkLinkMode,
-  NetworkInterfaceTypes,
-} from "app/store/machine/types";
-import {
-  getBondOrBridgeChild,
+  getInterfaceName,
   getInterfaceNumaNodes,
   getInterfaceTypeText,
+  getLinkInterface,
+  hasInterfaceType,
+  isBondOrBridgeParent,
   isBootInterface,
   isInterfaceConnected,
-  isBondOrBridgeParent,
 } from "app/store/machine/utils";
 import type { RootState } from "app/store/root/types";
 import { actions as subnetActions } from "app/store/subnet";
@@ -67,26 +65,33 @@ const getSortValue = (sortKey: SortKey, row: NetworkRow) =>
   row.sortData[sortKey];
 
 const generateRow = (
-  nic: NetworkLinkInterface,
+  nic: NetworkInterface | null,
+  link: NetworkLink | null,
   machine: Machine,
   fabrics: Fabric[],
   vlans: VLAN[],
   fabricsLoaded: boolean,
   vlansLoaded: boolean
-): NetworkRow => {
-  const isABondOrBridgeParent = isBondOrBridgeParent(machine, nic);
-  const connectingInterface = getBondOrBridgeChild(machine, nic);
-  const isBoot = isBootInterface(machine, nic);
-  const numaNodes = getInterfaceNumaNodes(machine, nic);
-  const vlan = vlans.find(({ id }) => id === nic.vlan_id);
-  const fabric = vlan ? fabrics.find(({ id }) => id === vlan.fabric) : null;
+): NetworkRow | null => {
+  if (link && !nic) {
+    [nic] = getLinkInterface(machine, link);
+  }
+  if (!nic) {
+    return null;
+  }
+  const isABondOrBridgeParent = isBondOrBridgeParent(machine, nic, link);
+  const isBoot = isBootInterface(machine, nic, link);
+  const numaNodes = getInterfaceNumaNodes(machine, nic, link);
+  const vlan = vlans.find(({ id }) => id === nic?.vlan_id);
+  const fabric = vlan ? fabrics.find(({ id }) => id === vlan?.fabric) : null;
+  const name = getInterfaceName(machine, nic, link);
   return {
     columns: [
       {
         content: (
           <DoubleRow
             data-test="name"
-            primary={nic.name}
+            primary={name}
             secondary={nic.mac_address}
           />
         ),
@@ -100,16 +105,21 @@ const generateRow = (
           ) : null,
       },
       {
-        content: [
-          NetworkInterfaceTypes.BOND,
-          NetworkInterfaceTypes.BRIDGE,
-          NetworkInterfaceTypes.VLAN,
-        ].includes(nic.type) ? null : (
+        content: hasInterfaceType(
+          [
+            NetworkInterfaceTypes.BOND,
+            NetworkInterfaceTypes.BRIDGE,
+            NetworkInterfaceTypes.VLAN,
+          ],
+          machine,
+          nic,
+          link
+        ) ? null : (
           <DoubleRow
             data-test="speed"
             icon={
               <>
-                {isInterfaceConnected(nic) ? null : (
+                {isInterfaceConnected(machine, nic, link) ? null : (
                   <Tooltip
                     position="top-left"
                     message="This interface is disconnected."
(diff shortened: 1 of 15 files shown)


ui/src/app/machines/views/MachineDetails/MachineNetwork/NetworkTable/NetworkTable.tsx
```diff
@@ -58,184 +66,240 @@ type SortKey = keyof NetworkRowSortData;
...
 
-const generateRows = (
+const generateRow = (
+  nic: NetworkLinkInterface,
   machine: Machine,
   fabrics: Fabric[],
   vlans: VLAN[],
   fabricsLoaded: boolean,
   vlansLoaded: boolean
-): NetworkRow[] => {
```
huwshimi, line 67: Despite what this diff shows, all these changes in the `generateRow` function are indentation changes (content moved from generateRow**s** to generateRow).

ui/src/app/machines/views/MachineDetails/MachineNetwork/NetworkTable/NetworkTable.tsx
```diff
@@ -58,184 +66,240 @@ type SortKey = keyof NetworkRowSortData;
...
+  vlans: VLAN[],
+  fabricsLoaded: boolean,
+  vlansLoaded: boolean
+): NetworkRow[] => {
+  if (!machine || !("interfaces" in machine)) {
+    return [];
+  }
+  const interfaces: NetworkLinkInterface[] = [];
+  // Create a list of interfaces and aliases to use to generate the table rows.
+  machine.interfaces.forEach((nic: NetworkInterface) => {
```
Caleb-Ellis, line 257: It seems like there's a bit too much object manipulation happening in this function, so it's hard to know exactly what is going where. Introducing a new type `NetworkLinkInterface` mitigates this a bit, however it introduces its own weirdness because it feels like the type is masquerading as something defined in the API but it really only exists as a convenience for generating table rows in the UI.

I ran into a similar issue in the storage tab where disks (interfaces) can have partitions (links) but the hierarchy isn't really visualised in the table. Originally I wrote it similar to this where a common "row" object is defined (like `NetworkLinkInterface`, but more bespoke), but when I started implementing the actions it became apparent that they're two separate entities and should be treated as such. My solution was to keep the data from the API intact as long as possible by rewriting most of the row generation logic so that the functions accepted both a `Disk` and `Partition` type, and only at the render level make the differences apparent.

I think it works better and is simpler. With the intermediary type, the data flow went like:
- Get disk or partition data from API
- Convert disk or partition into intermediary type
- Determine whether the intermediary type was originally a disk or partition
- Render the disk or partition data

And without the intermediary type it goes like:
- Get disk or partition data from API
- Determine whether it's a disk or partition
- Render the disk or partition data

This is all to say I'm not actually sure if it's the best way to address this, because now the conditionals have just been moved from the row object generation to the render function so there's a bunch of conditionals like `{isDisk(storageDevice) ? diskValue : partitionValue}` which is its own version of weirdness. But I think the most important thing is that the data from the API is changed as little as possible.

I'm interested to hear what you think, and sorry for the wall of text!

huwshimi: Thanks @Caleb-Ellis I think it's good to try and get this right so that it's not as hard to untangle as the legacy version of this code (where you never really know where the row data has come from).

I'll have another try at using the original types and see how I go.

huwshimi:
@Caleb-Ellis I've updated this now to remove the intermediate type and everything accepts either an interface or a link and it then grabs the data from either the link or the interface as needed. I think it's better, what do you think?

ui/src/app/machines/views/MachineDetails/MachineNetwork/NetworkTable/NetworkTableActions/NetworkTableActions.tsx
```diff
@@ -24,33 +40,36 @@ const NetworkTableActions = ({ nic, systemId }: Props): JSX.Element | null => {
...
+        {
+          children: "Mark as disconnected",
+        },
+      ]),
+      ...(canAddAliasOrVLAN && [
+        {
+          children: "Add alias or VLAN",
+        },
+      ]),
+      ...(!cannotEditInterface && [
```
Caleb-Ellis, line 61: Not that it matters for this PR but it seems a bit odd that the rest of the actions are `can` and this one is `!cannot`

ui/src/app/store/machine/utils/networking.ts
```diff
@@ -18,22 +18,146 @@ const INTERFACE_TYPE_DISPLAY = {
...
+ * @param machine - The nic's machine.
+ * @param link - A link to an interface.
+ * @return Whether this is a link.
+ */
+export const isAlias = (
+  machine: Machine,
+  link: NetworkLink | null
+): boolean => {
+  const [, linkIndex] = getLinkInterface(machine, link);
+  // The first link provides supplementary data for the non-alias interface.
```
Caleb-Ellis, line 56: Such a weird api! :man_shrugging:

ui/src/app/store/machine/utils/networking.ts
```diff
@@ -18,22 +18,146 @@ const INTERFACE_TYPE_DISPLAY = {
...
+  if (link && !nic) {
+    [nic] = getLinkInterface(machine, link);
+  }
+  return link && isAlias(machine, link)
+    ? NetworkInterfaceTypes.ALIAS
+    : nic?.type || null;
+};
+
+/**
+ * Get the type of an interface.
```
Caleb-Ellis, line 105: doc block needs updating

ui/src/app/store/machine/utils/networking.ts
```diff
@@ -18,22 +18,146 @@ const INTERFACE_TYPE_DISPLAY = {
...
+  machine: Machine,
+  link: NetworkLink | null
+): boolean => {
+  const [, linkIndex] = getLinkInterface(machine, link);
+  // The first link provides supplementary data for the non-alias interface.
+  return !!link && typeof linkIndex === "number" && linkIndex > 0;
+};
+
+/**
+ * Get the type of an interface.
```
Caleb-Ellis, line 61: doc block needs updating

ui/src/app/store/machine/utils/networking.ts
```diff
@@ -18,22 +18,146 @@ const INTERFACE_TYPE_DISPLAY = {
...
+      if (links[j].id === link.id) {
+        return [machine.interfaces[i], j];
+      }
+    }
+  }
+  return [null, null];
+};
+
+/**
+ * Whether an interface is a link.
```
Caleb-Ellis, line 46: doc block needs updating

Caleb-Ellis (approved):
Nice! I feel like it's much easier to follow along with what's happening now. Just need to update some doc blocks.

Have to say though, the way to determine if a link is an alias is *really* weird.

ui/src/app/machines/views/MachineDetails/MachineNetwork/NetworkTable/NetworkTableActions/NetworkTableActions.tsx
huwshimi: Hmm yeah good point, that was probably me following the legacy utils. I'll see if the logic can be reversed when I implement the util.

ui/src/app/store/machine/utils/networking.ts
huwshimi: Done.

ui/src/app/store/machine/utils/networking.ts
huwshimi: Yeah, not the most helpful!

ui/src/app/store/machine/utils/networking.ts
huwshimi: Done.

ui/src/app/store/machine/utils/networking.ts
huwshimi: Done.

Issue #2011, opened 2020-12-16
Network tab - make rows sortable


huwshimi:
Make sure all columns have sort data.

PR #2051, opened 2021-01-14
feat(ui): make the network table sortable

huwshimi:
## Done

- Move the data manipulation to utils and use it in the table to set the sortable data.

### MAAS deployment

- [Bolla](/HACKING.md#bolla)

### Running the branch

You can run this branch by:

### QA steps

- View the react network tab for a few machines.
- You should be able to sort the table by all the headers (the node/bridge interfaces will be made to keep their order in #1946.

## Fixes

Fixes: #2011.

### DIFF
--- a/ui/src/app/machines/views/MachineDetails/MachineNetwork/NetworkTable/IPColumn/IPColumn.test.tsx
+++ b/ui/src/app/machines/views/MachineDetails/MachineNetwork/NetworkTable/IPColumn/IPColumn.test.tsx
@@ -8,41 +8,48 @@ import { HardwareType, ResultType } from "app/base/enum";
 import { NetworkLinkMode } from "app/store/machine/types";
 import type { RootState } from "app/store/root/types";
 import { ResultStatus } from "app/store/scriptresult/types";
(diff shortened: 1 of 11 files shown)


Caleb-Ellis (approved):
LGTM!

I wonder if maybe in the future it would be good to have hooks that get an interface's fabric/subnet/vlan, where they're responsible for getting all the relevant bits of state to search through. So in a component you would just do something like:
``` typescript
const vlan = useInterfaceVlan(machine, nic);
```
Then you wouldn't need to use selectors to get all the fabrics/vlans/subnets in the component itself. Anyway just an idea, seems like some of the utils are getting a bit  "arg-heavy" and just wondering if there would be a good way to simplify it.

huwshimi:
@Caleb-Ellis thank, I think that's a good idea, but I think those hooks would just end up wrapping the current utils, as we need to be able to loop through the interfaces in the interface table and use the utils to get the correct data and hooks have to be called at the top level of the component. But we could add those hooks and use them in the sub components where we have the individual interface/link ids.

PR #2078, opened 2021-01-22
refactor(ui): move group checkbox to shared component

huwshimi:
## Done

- Dedupe the checkbox for selecting everything in a table.

### MAAS deployment

- [Bolla](/HACKING.md#bolla)

### Running the branch

You can run this branch by:

### QA steps

- Go to the machine list.
- Make sure you can use the group checkbox and top/all checkbox to select/deselect the rows.
- Make sure selecting some rows makes the top checkboxe turn into a dash.
- Do the same for the RSD and KVM lists.
- Navigate to the storage tab for a machine and make sure the checkboxes work in the available storage table.

## Fixes

First step for: canonical-web-and-design/maas-squad#2284.

### DIFF
--- a/ui/src/app/base/components/GroupCheckbox/GroupCheckbox.test.tsx
+++ b/ui/src/app/base/components/GroupCheckbox/GroupCheckbox.test.tsx
@@ -1,0 +1,40 @@
+import { shallow } from "enzyme";
+
+import GroupCheckbox from "./GroupCheckbox";
+
+describe("GroupCheckbox", () => {
+  it("renders", () => {
+    const wrapper = shallow(
+      <GroupCheckbox
+        items={[]}
+        selectedItems={[]}
+        handleGroupCheckbox={jest.fn()}
+      />
+    );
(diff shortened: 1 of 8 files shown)


Caleb-Ellis (commented):
Nice! Always feels great when a new generic component is created. Anyway it mostly looks good, but just a couple of adjustments to be made:
- The `u-align-header-checkbox` class is only meant to be used on checkboxes in table headers, because the header font is smaller than normal. For the machine list group headers it makes it look misaligned. I think we either need an extra prop for `wrapperClassName` or `header` to only apply the class on table header group checkboxes. Either that or we could amend `u-nudge--checkbox` to be context-aware and get rid of `u-align-header-checkbox`.
- `u-nudge--checkbox` is being applied to the machine list group checkboxes twice. I think we just need to get rid of `primaryClassName` on the group `DoubleRow`

huwshimi:
@Caleb-Ellis those issues should be fixed now.

Caleb-Ellis (approved):
LGTM!

PR #2093, opened 2021-01-27
feat(ui): add network table checkboxes

huwshimi:
## Done

- Add checkboxes to the network table in machine details.

### MAAS deployment

- [Bolla](/HACKING.md#bolla)

### Running the branch

You can run this branch by:

### QA steps

- Visit the network tab in machine details for a new or ready machine.
- Check that there are checkboxes and that they work as expected.
- Visit a deployed machine and check that the checkboxes are disabled.

## Fixes

Fixes: canonical-web-and-design/maas-squad#2284.

### DIFF
--- a/ui/src/app/machines/views/MachineDetails/MachineNetwork/NetworkTable/NetworkTable.tsx
+++ b/ui/src/app/machines/views/MachineDetails/MachineNetwork/NetworkTable/NetworkTable.tsx
@@ -523,6 +523,15 @@ const NetworkTable = ({
     vlansLoaded
   );
   const sortedRows = sortRows(rows);
+  const nicIds = machine.interfaces.reduce<NetworkInterface["id"][]>(
+    (nics = [], nic) => {
+      if (!isBondOrBridgeParent(machine, nic)) {
+        nics.push(nic.id);
+      }
+      return nics;
+    },
+    []
+  );
   return (
     <MainTable
       className="p-table-expanding--light"
@@ -534,7 +543,8 @@ const NetworkTable = ({
           content: (
             <>
               <GroupCheckbox
-                items={machine.interfaces.map(({ id }) => id)}
+                disabled={isAllNetworkingDisabled}
+                items={nicIds}
                 selectedItems={selected}
                 handleGroupCheckbox={handleGroupCheckbox}
               />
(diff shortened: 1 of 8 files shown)


Caleb-Ellis (commented):
The checkboxes are correctly not showing up for the parent interfaces, but they're still being factored into the group checkbox behaviour, i.e. manually selecting all child interfaces will have the group checkbox showing a `-` even though everything that *can* be selected, is selected. Similarly, checking the group checkbox makes it a tick, which probably means the parent interface id is getting included in selected state.

As an aside it looks like the table headers have messed up at some point. This happened to me too in the storage tab when making one of the tables expanding, which turns the headers into flex elements. Wrapping the double row into a single container div should fix it.

huwshimi:
> The checkboxes are correctly not showing up for the parent interfaces, but they're still being factored into the group checkbox behaviour, i.e. manually selecting all child interfaces will have the group checkbox showing a `-` even though everything that _can_ be selected, is selected. Similarly, checking the group checkbox makes it a tick, which probably means the parent interface id is getting included in selected state.

Thanks, that should be fixed now.

> As an aside it looks like the table headers have messed up at some point. This happened to me too in the storage tab when making one of the tables expanding, which turns the headers into flex elements. Wrapping the double row into a single container div should fix it.

I have a fix for this in another branch (next PR). It was making it an expanding table that broke it, just like your situation.

ui/src/app/machines/views/MachineDetails/MachineNetwork/NetworkTable/NetworkTable.tsx
```diff
@@ -500,6 +542,11 @@ const NetworkTable = ({ systemId }: Props): JSX.Element => {
         {
           content: (
             <>
+              <GroupCheckbox
+                items={nicIds}
```
Caleb-Ellis, line 546: Maybe this should include the disabled prop if networking is disabled?

Caleb-Ellis (approved):
LGTM! Just one comment for consideration

ui/src/app/machines/views/MachineDetails/MachineNetwork/NetworkTable/NetworkTable.tsx
huwshimi: Oops yes it needs that!

PR #2123, opened 2021-02-03
refactor(ui): create generic TableActionsDropdown component and use in storage

### LINKED ISSUE #2072: Refactor storage table actions
Disks and partitions generate actions in a similar way - we should just combine them into a single component. Plus it will help reduce the size of the `AvailableDisksAndPartitions` component

Caleb-Ellis:
## Done

- Created `TableActionsDropdown` component, which is an extension of `TableMenu` but used specifically for the actions cell.
- Moved storage action code out of `AvailableStorageTable` and into new `StorageDeviceActions` component.
- Converted storage action strings into enums. 

### MAAS deployment

- [Bolla](/HACKING.md#bolla)

### Running the branch

You can run this branch by:

### QA steps

- Click around the storage tab of a Ready or Allocated machine and check that the actions still work.

## Fixes

Fixes #2072

### DIFF
--- a/ui/src/app/base/components/TableActionsDropdown/TableActionsDropdown.tsx
+++ b/ui/src/app/base/components/TableActionsDropdown/TableActionsDropdown.tsx
@@ -1,0 +1,48 @@
+import type { Props as ButtonProps } from "@canonical/react-components/dist/components/Button";
+
+import TableMenu from "app/base/components/TableMenu";
+
+export type TableAction<A> = {
+  label: string;
+  show?: boolean;
+  type: A;
+};
+
+// This allows the "data-test" attribute to be used for the action links, which
+// is technically not a valid HTML prop, but we use it throughout maas-ui.
+type TableActionsLink = ButtonProps & { "data-test": string };
+
+type Props<A> = {
+  actions: TableAction<A>[];
+  disabled?: boolean;
+  onActionClick: (action: A) => void;
+};
+
+const TableActionsDropdown = <A extends string>({
+  actions,
+  disabled = false,
+  onActionClick,
+}: Props<A>): JSX.Element => {
+  const actionLinks = actions.reduce<TableActionsLink[]>((links, action) => {
+    if (!(action.show === false)) {
+      // Show actions that do not explicitly set show to false.
+      links.push({
+        children: action.label,
+        "data-test": action.type,
+        onClick: () => onActionClick(action.type),
+      });
+    }
+    return links;
+  }, []);
+
+  return (
+    <TableMenu
+      disabled={disabled || actions.length === 0}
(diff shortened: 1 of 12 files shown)


ui/src/app/base/components/TableActionsDropdown/TableActionsDropdown.tsx
```diff
@@ -0,0 +1,48 @@
...
+
+export type TableAction<A> = {
+  label: string;
+  show?: boolean;
+  type: A;
+};
+
+// This allows the "data-test" attribute to be used for the action links, which
+// is technically not a valid HTML prop, but we use it throughout maas-ui.
+type TableActionsLink = ButtonProps & { "data-test": string };
```
huwshimi, line 13: This seems strange as it seems to work for other types that we've defined. Have we done something wrong with the Button type? Maybe `HTMLProps<HTMLElement>` doesn't include `data-*`.

huwshimi: Hmm it doesn't complain if you add use it directly on a button `<Button data-test="works" />`.

ui/src/app/base/components/TableActionsDropdown/TableActionsDropdown.tsx
```diff
@@ -0,0 +1,48 @@
...
+};
+
+const TableActionsDropdown = <A extends string>({
+  actions,
+  disabled = false,
+  onActionClick,
+}: Props<A>): JSX.Element => {
+  const actionLinks = actions.reduce<TableActionsLink[]>((links, action) => {
+    if (!(action.show === false)) {
+      // Show actions that do not explicitly set show to false.
```
huwshimi, line 28: Why does it need to be an explicit `false`?

huwshimi (approved):
Lots of nice improvements here! Just a couple of questions.

ui/src/app/base/components/TableActionsDropdown/TableActionsDropdown.tsx
Caleb-Ellis: Just so if the action object doesn't have a `show` key, the assumption is to show the action anyway. I could have rewritten the  object to have a `hide` key, then `if (!action.hide)` would do the same thing but I think `show` is clearer in the list of actions.

ui/src/app/base/components/TableActionsDropdown/TableActionsDropdown.tsx
Caleb-Ellis: I think the issue might lie with `ContextualMenu`'s `MenuLink` type, but I can't quite figure it out...

PR #2127, opened 2021-02-03
refactor(ui): refactor some common form Select components

### LINKED ISSUE #2126: Create common form select components

Caleb-Ellis:
## Done

- Created common form Select components that are used in a few places. Each of them do the same thing - load the relevant items into state and turn them into a select. These will also be used in the machine configuration tab (and likely devices and controllers)
  - `ArchitectureSelect`
  - `DomainSelect`
  - `MinimumKernelSelect`
  - `ResourcePoolSelect`
  - `ZoneSelect` 

### MAAS deployment

- [Bolla](/HACKING.md#bolla)

### Running the branch

You can run this branch by:

### QA steps

- Go to /MAAS/r/machines/add and check that the selects still work correctly (i.e. is populated and can select option), and the form submits correctly

## Fixes

Fixes #2126

### DIFF
--- a/ui/src/app/base/components/FormikField/FormikField.tsx
+++ b/ui/src/app/base/components/FormikField/FormikField.tsx
@@ -1,3 +1,4 @@
+import type { HTMLProps } from "react";
 import { useRef } from "react";
 
 import { Input } from "@canonical/react-components";
@@ -10,7 +11,7 @@ import type { TSFixMe } from "app/base/types";
 type Props = {
   Component?: JSX.Element;
   name: string;
-  value?: number | string;
+  value?: HTMLProps<HTMLElement>["value"];
   [x: string]: TSFixMe;
 };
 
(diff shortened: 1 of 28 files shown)


ui/src/app/base/components/FormikField/FormikField.tsx
```diff
@@ -10,7 +10,7 @@ import type { TSFixMe } from "app/base/types";
 type Props = {
   Component?: JSX.Element;
   name: string;
-  value?: number | string;
+  value?: number | string | readonly string[];
```
huwshimi, line 13: Could be worth a comment as to why this might be readonly.

huwshimi (approved):
Looks good!

ui/src/app/base/components/FormikField/FormikField.tsx
Caleb-Ellis: I've opted to use `HTMLProps<HTMLElement>["value"]` instead which I think is clearer what `value` is supposed to/can be. `FormikField` would probably be better off using a generic to determine what props are allowed but I think that might need a bit more thought.

Issue #2132, opened 2021-02-04
DomainDetailsController: Cannot read property 'length' of undefined


anthonydillon:
https://sentry.is.canonical.com/canonical/maas-ui/issues/15981/?referrer=github_plugin

```
TypeError: Cannot read property 'length' of undefined
  at ChildScope.DomainDetailsController.$scope.canBeDeleted (/MAAS/r/1.6f85774a2928f988cdb6.js:34355:106)
  at fn (/MAAS/r/1.6f85774a2928f988cdb6.js:1:1)
  at Scope.$digest (/MAAS/r/1.6f85774a2928f988cdb6.js:12439:35)
  at Scope.$apply (/MAAS/r/1.6f85774a2928f988cdb6.js:12663:163)
  at None (/MAAS/r/1.6f85774a2928f988cdb6.js:13882:454)
...
(2 additional frame(s) were not displayed)
```

PR #2133, opened 2021-02-04
fix(legacy) null check rrsets in domain

anthonydillon:
## Done
Add extra null checks for the existence of `rrsets` when asserting if a domain can be deleted.

### MAAS deployment
- [Bolla](/HACKING.md#bolla)

### Running the branch
You can run this branch by:

### QA steps
- Go to /MAAS/l/domain/2
- Try and delete to the domain and see there are no errors in the console

## Fixes
Fixes: https://github.com/canonical-web-and-design/maas-ui/issues/2132

## Backports
N/A


hatched (approved):

legacy/src/app/controllers/domain_details.js
```diff
@@ -129,7 +129,7 @@ function DomainDetailsController(
 
   // Called to check if the space can be deleted.
   $scope.canBeDeleted = function () {
-    if (angular.isObject($scope.domain)) {
+    if (angular.isObject($scope.domain) && $scope.domain.rrsets) {
```
huwshimi, line 132: This could just be `if ($scope.domain?.rrsets) {`

huwshimi: Also, is it possible this should remain unchanged and the line below should be: `return !$scope.domain.rrsets || $scope.domain.rrsets.length === 0;`?

huwshimi:
Hey @anthonydillon just curious why you pushed this to upstream instead of your fork?

huwshimi (changes requested):
@Caleb-Ellis and I have been unable to get the error on master so that we can confirm that this fixes it. Did you manage to reproduce it?

At any rate, I think `return !$scope.domain.rrsets || $scope.domain.rrsets.length === 0;` (or `return !$scope.domain?.rrsets?.length`) is probably a better fix as it looks like the intention here is to check that there are no `rrsets`.

I think it might also be sensible to add a test without the `rrsets` array: https://github.com/canonical-web-and-design/maas-ui/blob/master/legacy/src/app/controllers/tests/test_domain_details.js#L162

legacy/src/app/controllers/domain_details.js
anthonydillon: Sure, I would have used optional chaining in modern code but I didn't want to introduce a different way of coding in the middle of a file. This video is to be replaced in the near future. 

The second suggestion seems very complex to understand for the next developer that comes a long. I wanted to make it simple to read.

anthonydillon:
I can add a test, just wasn't sure it was worth it for a null check in a code base that is being replaced

huwshimi:
> I can add a test, just wasn't sure it was worth it for a null check in a code base that is being replaced

We tend to use the tests as a reference when rebuilding (or in the case of this utility probably porting it ). Also, in lieu of being able to QA the issue this would at least confirm that it fixes it.

huwshimi (approved):

anthonydillon:
@huwshimi thanks, test added

huwshimi:
Thanks @anthonydillon!

PR #2157, opened 2021-02-10
refactor(ui): convert power utils to TS and move to general directory

Caleb-Ellis:
## Done

- Converted power utils to TS and moved to `app/store/general/utils`:
  - `formatPowerParameters` has changed args so that chassis and scope are separate (necessary for machine config)
  - `usePowerParametersSchema` is now `generatePowerParametersSchema` (no longer a hook). It generates a Yup object shape based on the given power type and its fields
  - `useAllPowerParameters` is now `useInitialPowerParameters` - I'll need to be able to initialise the form with predefined values for the machine configuration form so it's been updated to allow that

### MAAS deployment

- [Bolla](/HACKING.md#bolla)

### Running the branch

You can run this branch by:

### QA steps

- Check that the forms for Add Machine, Add Chassis and Add KVM all still validate properly
- Try to add a machine (doesn't have to be valid values) and check the websocket request - make sure it only includes the values relevant to the selected power type

## Fixes

Fixes canonical-web-and-design/MAAS-squad#2333


ui/src/app/machines/views/AddMachine/AddMachineForm/AddMachineForm.js
```diff
@@ -107,13 +82,31 @@ export const AddMachineForm = () => {
...
-
-  const allPowerParameters = useAllPowerParameters(powerTypes);
-
+  const initialPowerParameters = useInitialPowerParameters(powerTypes);
+  const AddMachineSchema = Yup.object().shape({
+    architecture: Yup.string().required("Architecture required"),
+    domain: Yup.string().required("Domain required"),
+    extra_macs: Yup.array().of(
+      Yup.string().matches(
+        /^([0-9A-Fa-f]{2}:){5}([0-9A-Fa-f]{2})$/,
```
huwshimi, line 91: We should make this a constant somewhere.

ui/src/app/store/machine/types.ts
```diff
@@ -238,39 +238,7 @@ export type MachineNumaNode = Model & {
...
-  power_port?: string;
-  power_protocol?: string;
-  power_user?: string;
-  power_uuid?: string;
-  power_vm_name?: string;
-  privilege_level?: string;
-  server_name?: string;
-  uuid?: string;
-};
+export type PowerParameters = { [x: string]: string | number };
```
huwshimi, line 241: Are there other types that weren't captured by the types above? Could we define the types for the node and/or power type instead (e.g. a type for the power types for virsh)? Would there be any value in doing that?

huwshimi (approved):
This is looking great, one suggestion and one question.

ui/src/app/machines/views/AddMachine/AddMachineForm/AddMachineForm.js
Caleb-Ellis: Done - added `app/base/validation/ts` with this regex

ui/src/app/store/machine/types.ts
Caleb-Ellis: Yeah when I first wrote this type I just dumped all the possible power parameters from all the possible power types at that point. But since then (not long ago), some power types have been added (with their own parameters) and some parameters have changed, so I'm thinking it's a bit too dynamic to be so strongly typed.

Issue #2161, opened 2021-02-11
Fix loading legacy machine details


huwshimi:
From https://bugs.launchpad.net/maas/+bug/1915166:

> Version 2.9 has a bug when we try to load node information, if you have more than 2k nodes, you will receive a message 'no item with pk: machine_id', it is an intermittent issue, sometimes it works.
> the API shows the machine information without a problem.
> Looks like a limitation of the Maas and the number of nodes, no problem with the database as well.

PR #2162, opened 2021-02-11
fix(legacy): fix race condition when loading machine details

huwshimi:
## Done

- Always add the loaded machine to the store when fetching a single machine for machine details.

### MAAS deployment

- [Bolla](/HACKING.md#bolla)

### Running the branch

You can run this branch by:

### QA steps

- Visit a machine details page.
- Click on DNS in the header and navigate to one of the domains.
- From the list of resources click a machine (if there is no machine you might need to add one or look in another domain).
- The machine details should load.
- Do this a few more times and you should see the machine every time.

## Fixes

Fixes: #2161.


Caleb-Ellis (approved):
Nice! Was able to reproduce the bug (semi) reliably and looks like this fixes it! :+1:

PR #2164, opened 2021-02-12
feat(ui): build machine power configuration form

Caleb-Ellis:
## Done

- Built "Power configuration" section of the config tab
- Changed links to point to the React version of the config tab
- Updated `FormikForm`'s `onCancel` prop so you can give it the entire formik context as an argument and access things like `initialValues` and `resetForm`. I've used it in the MachineForm and PowerForm to reset the form when "Cancel" is pressed.

### MAAS deployment

- [Bolla](/HACKING.md#bolla)

### Running the branch

You can run this branch by:

### QA steps

- Create a machine manually (i.e. not in a pod) and go to its details page
- Click "Configuration" and check that it takes you to the React configuration screen
- Check that the power details are initialised correctly
- Click the "Edit" button next to "Power configuration"
- Check that the select is enabled and you can change the power type and power parameters
- Change a few values then click "Cancel" and check that the values are reverted to the originals
- Enter an invalid value (e.g "1" for an ip address) and submit and check that en error is displayed
- Check that you can successfully update the power config with valid values
- Go to the config page for a machine that's in a pod
- Check that the select is always disabled, and you only see fields that are scoped to individual nodes

## Fixes

Fixes canonical-web-and-design/MAAS-squad#2223

### DIFF
--- a/ui/src/app/base/components/FormikForm/FormikForm.tsx
+++ b/ui/src/app/base/components/FormikForm/FormikForm.tsx
@@ -5,7 +5,7 @@ import type { FormikHelpers } from "formik";
 import PropTypes from "prop-types";
 import { useDispatch } from "react-redux";
 import { Redirect } from "react-router";
-import type { AnySchema } from "yup";
+import type { SchemaOf } from "yup";
 
 import FormikFormContent from "app/base/components/FormikFormContent";
 import type {
@@ -23,7 +23,7 @@ export type Props<V, E = FormErrors> = {
   };
   onSubmit: (values?: V, formikHelpers?: FormikHelpers<V>) => void;
   savedRedirect?: string;
-  validationSchema?: AnySchema;
+  validationSchema?: SchemaOf<V>;
 } & ContentProps<V, E>;
 
 const FormikForm = <V, E = FormErrors>({
--- a/ui/src/app/base/components/FormikForm/FormikForm.tsx
+++ b/ui/src/app/base/components/FormikForm/FormikForm.tsx
@@ -5,7 +5,7 @@ import type { FormikHelpers } from "formik";
 import PropTypes from "prop-types";
 import { useDispatch } from "react-redux";
 import { Redirect } from "react-router";
-import type { SchemaOf } from "yup";
+import type { AnySchema } from "yup";
 
 import FormikFormContent from "app/base/components/FormikFormContent";
 import type {
@@ -23,7 +23,7 @@ export type Props<V, E = FormErrors> = {
   };
   onSubmit: (values?: V, formikHelpers?: FormikHelpers<V>) => void;
   savedRedirect?: string;
-  validationSchema?: SchemaOf<V>;
+  validationSchema?: AnySchema;
 } & ContentProps<V, E>;
 
 const FormikForm = <V, E = FormErrors>({
@@ -44,7 +44,6 @@ const FormikForm = <V, E = FormErrors>({
   onSaveAnalytics = {},
   onSubmit,
   onValuesChanged,
-  resetOnCancel,
   resetOnSave,
   saving,
   savingLabel,
@@ -96,7 +95,6 @@ const FormikForm = <V, E = FormErrors>({
         initialValues={initialValues}
         onCancel={onCancel}
         onValuesChanged={onValuesChanged}
-        resetOnCancel={resetOnCancel}
         resetOnSave={resetOnSave}
         loading={loading}
         saving={saving}
@@ -133,7 +131,6 @@ FormikForm.propTypes = {
   }),
   onSubmit: PropTypes.func.isRequired,
   onValuesChanged: PropTypes.func,
-  resetOnCancel: PropTypes.bool,
   resetOnSave: PropTypes.bool,
   loading: PropTypes.bool,
   saving: PropTypes.bool,
--- a/ui/src/app/base/components/FormikForm/FormikForm.tsx
+++ b/ui/src/app/base/components/FormikForm/FormikForm.tsx
@@ -44,6 +44,7 @@ const FormikForm = <V, E = FormErrors>({
   onSaveAnalytics = {},
   onSubmit,
   onValuesChanged,
+  resetOnCancel,
   resetOnSave,
   saving,
   savingLabel,
@@ -95,6 +96,7 @@ const FormikForm = <V, E = FormErrors>({
         initialValues={initialValues}
         onCancel={onCancel}
         onValuesChanged={onValuesChanged}
+        resetOnCancel={resetOnCancel}
         resetOnSave={resetOnSave}
         loading={loading}
         saving={saving}
@@ -131,6 +133,7 @@ FormikForm.propTypes = {
   }),
   onSubmit: PropTypes.func.isRequired,
   onValuesChanged: PropTypes.func,
+  resetOnCancel: PropTypes.bool,
   resetOnSave: PropTypes.bool,
   loading: PropTypes.bool,
   saving: PropTypes.bool,
--- a/ui/src/app/base/components/PowerTypeFields/PowerTypeFields.tsx
+++ b/ui/src/app/base/components/PowerTypeFields/PowerTypeFields.tsx
@@ -11,6 +11,8 @@ import type { PowerType } from "app/store/general/types";
 import { PowerFieldScope } from "app/store/general/types";
 
 type Props = {
+  disableFields?: boolean;
+  disableSelect?: boolean;
   forChassis?: boolean;
   powerParametersValueName?: string;
   powerTypeValueName?: string;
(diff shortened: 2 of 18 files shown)


huwshimi:
@Caleb-Ellis a few QA things.

Non-pod machine config:
- If you change the power type then press cancel it doesn't revert the form to the old power type.
- If you enter a value for e.g. ip address and then change the power type then if there is another field for ip address in the new form it has the value you previously entered, is that intentional? At first I wasn't sure if it had prefilled the field some other way.
- If you have a form error it shows in the power form and the machine config form above it.
- If you get an error and then cancel it doesn't clear the error.
- I got an error for a bad ip address format that was really confusing. Could we do IP validation on the client or is it not possible because the fields are generated?
Screen Shot 2021-02-15 at 11 16 56 am

Pod machine config:
- There's a warning about changing config for fields that aren't shown. Do we have any control over that warning?
Screen Shot 2021-02-15 at 11 19 27 am

Caleb-Ellis:
@huwshimi thanks for the review! I thought breaking this form down into several smaller issues would've ironed out the kinks, but clearly not!

>If you change the power type then press cancel it doesn't revert the form to the old power type.

Fixed. I've updated `FormikForm`'s `onCancel` prop so you can give it the entire formik context as an argument and access things like `initialValues` and `resetForm`.

>If you enter a value for e.g. ip address and then change the power type then if there is another field for ip address in the new form it has the value you previously entered, is that intentional? At first I wasn't sure if it had prefilled the field some other way.

Fixed. In `PowerTypeFields` I've updated the select to set the power type defaults on change and to reset touched/errors so they don't carry over.

>If you have a form error it shows in the power form and the machine config form above it.

I'm not sure how to fix this one. I've changed it so that errors only show if the machine is in editing state, but they're both `machine.update` forms with different fields so they both have the same errors associated with them in state. Any ideas? Maybe something to address in a follow up?

>If you get an error and then cancel it doesn't clear the error.

Fixed. Added `cleanup` to the `onCancel` function.

>I got an error for a bad ip address format that was really confusing. Could we do IP validation on the client or is it not possible because the fields are generated?

I think if we want to do IP validation on the client the power field from the api should specify as such, otherwise there's the potential for the validation to be wrong. Right now we only know whether a field is a `string`, `password` or `choice`.

>There's a warning about changing config for fields that aren't shown. Do we have any control over that warning?

I actually just copied over the logic from the angular version, but I think the warning might be unnecessary. The fields that are displayed are limited to `PowerFieldScope.NODE` if the machine is in a pod, so it shouldn't be possible to change anything that'll affect other nodes. Removing for now.

ui/src/app/base/components/FormikForm/FormikForm.tsx
```diff
@@ -23,7 +23,7 @@ export type Props<V, E = FormErrors> = {
   };
   onSubmit: (values?: V, formikHelpers?: FormikHelpers<V>) => void;
   savedRedirect?: string;
-  validationSchema?: SchemaOf<V>;
+  validationSchema?: AnySchema;
```
Caleb-Ellis, line 26: This was causing a typescript error for any `FormikForm` that was given a generic

huwshimi:
@Caleb-Ellis this looks mostly good now except if you make some changes and save then make some more changes and press cancel the changes go back to the original version before the first save.

ui/src/app/base/components/FormikForm/FormikForm.tsx
huwshimi: I think the previous type was correct, but looks like you have to do a little bit more setup to define the type of the Schema:

```
  const PowerFormSchema: SchemaOf<PowerFormValues> = Yup.object()
    .shape({
      powerParameters: Yup.object().shape(powerParametersSchema),
      powerType: Yup.string().required("Power type is required"),
    })
    .defined();
```
From: https://github.com/jquense/yup/blob/master/docs/typescript.md

ui/src/app/base/components/PowerTypeFields/PowerTypeFields.tsx
```diff
@@ -101,12 +117,30 @@ export const PowerTypeFields = <F extends Record<string, unknown>>({
               value: powerType.name,
             })),
           ]}
+          onChange={(e: React.ChangeEvent<HTMLSelectElement>) => {
```
huwshimi, line 120: I think what this does is loop through and reset any fields that are for the selected power type. Should it also reset the values for the old power type (or maybe just everything) as couldn't it leave values in the formik state for the old fields? Not sure exactly how that works.

Either way, it would probably also good to leave a comment here explaining why this is necessary.

ui/src/app/machines/views/MachineDetails/MachineConfiguration/MachineForm/MachineForm.tsx
```diff
@@ -84,7 +84,7 @@ const MachineForm = ({ systemId }: Props): JSX.Element | null => {
           buttons={FormCardButtons}
           cleanup={cleanup}
           editable={editing}
-          errors={errors}
+          errors={editing ? errors : undefined}
```
huwshimi, line 87: Does this need to be `undefined` here? If so maybe leave a comment (or maybe see if it can be fixed so it doesn't have to be `undefined`) otherwise this should probably use `null`.

ui/src/app/machines/views/MachineDetails/MachineConfiguration/PowerForm/PowerForm.test.tsx
```diff
@@ -0,0 +1,142 @@
...
+      <Provider store={store}>
+        <PowerForm systemId="abc123" />
+      </Provider>
+    );
+
+    expect(wrapper.find("FormikForm").prop("editable")).toBe(false);
+
+    wrapper.find("button[data-test='edit-power-config']").simulate("click");
+
+    expect(wrapper.find("FormikForm").prop("editable")).toBe(true);
```
huwshimi, line 69: Looks

like

this

could

probably

have

less

empty

lines.

ui/src/app/machines/views/MachineDetails/MachineConfiguration/PowerForm/PowerForm.tsx
```diff
@@ -0,0 +1,151 @@
...
+            )}
+          </Col>
+        </Row>
+        <FormikForm<PowerFormValues>
+          allowAllEmpty
+          allowUnchanged
+          buttons={FormCardButtons}
+          cleanup={cleanup}
+          editable={editing}
+          errors={editing ? errors : undefined}
```
huwshimi, line 103: `null` here as well?

ui/src/app/machines/views/MachineDetails/MachineConfiguration/PowerForm/PowerForm.test.tsx
```diff
@@ -0,0 +1,142 @@
...
+          power_parameters: {},
+          power_type: "manual",
+          pxe_mac: machine.pxe_mac,
+          system_id: machine.system_id,
+        },
+      },
+      type: "machine/update",
+    });
+  });
+});
```
huwshimi, line 142: Might also be good to add a couple of tests covering the QA issues I found (resetting fields etc.) so that we catch any regressions.

huwshimi (changes requested):
A few changes required here sorry.

ui/src/app/base/components/PowerTypeFields/PowerTypeFields.tsx
Caleb-Ellis: It shouldn't be necessary to reset any other fields than from the selected power type because the power parameters are trimmed before the websocket call is made. Resetting everything won't work (at least using `resetForm()`) because it'll end up resetting non-power-related form fields as well.

I'll leave a comment explaining what this is for (and add a test)

ui/src/app/base/components/FormikForm/FormikForm.tsx
Caleb-Ellis: Aaaaaaahhhhh I missed the `defined()` bit when I was looking at that earlier. Nice, thanks, I'll revert the change

ui/src/app/machines/views/MachineDetails/MachineConfiguration/PowerForm/PowerForm.test.tsx
Caleb-Ellis: But





why?

ui/src/app/machines/views/MachineDetails/MachineConfiguration/MachineForm/MachineForm.tsx
Caleb-Ellis: Changed to `null`

ui/src/app/machines/views/MachineDetails/MachineConfiguration/PowerForm/PowerForm.test.tsx
Caleb-Ellis: Good call. I've added tests for:
- Power fields being set to defaults when the power type is changed
- MachineForm/PowerForm only showing errors in editing state
- MachineForm/PowerForm being reset to initial values when for is cancelled
- MachineForm/PowerForm dispatching a machine cleanup action on cancel

Caleb-Ellis:
@huwshimi Alright I think this is the one!

huwshimi (approved):
Looks good.

PR #2167, opened 2021-02-12
feat(ui) machine details commissioning details view

anthonydillon:
## Done
- Point commissioning details to a commissioning details view, reusing the test details components
- Made the back from the details view dynamic based on the section you are in
- Updated the path to the React version of commissioning
- ~~**Drive by** Add `is-dense` to TableMenu to stop it jumping so much while loading~~ ***Reverted and [issue filed](https://github.com/canonical-web-and-design/maas-ui/issues/2172)***
- **Drive-by** Right align the back to testing view link on test details view
- **Drive-by** Manage more icon states in the commissioning table
-  **Drive-by** Fixes links to testing from tests

### MAAS deployment

- [Bolla](/HACKING.md#bolla)

### Running the branch
You can run this branch by:

- Run with `dotrun`

### QA steps
- Go to a machine details
- Switch to commissioning
- Find and click a test that has "View details"
- See that the details match bolla
- Click back link and check it returns you to the commissioning view
- Do the same for tests and check the return link works

## Fixes
Fixes: https://github.com/canonical-web-and-design/MAAS-squad/issues/2213

### DIFF
--- a/ui/src/app/base/components/TableMenu/TableMenu.tsx
+++ b/ui/src/app/base/components/TableMenu/TableMenu.tsx
@@ -38,7 +38,7 @@ const TableMenu = <L extends null>({
       // See: https://github.com/canonical-web-and-design/react-components/issues/377
       positionNode={positionNode || undefined}
       toggleAppearance="base"
-      toggleClassName="u-no-margin--bottom p-table-menu__toggle is-dense"
+      toggleClassName="u-no-margin--bottom p-table-menu__toggle"
       toggleDisabled={disabled || false}
     />
   );
--- a/ui/src/app/base/components/TableMenu/TableMenu.tsx
+++ b/ui/src/app/base/components/TableMenu/TableMenu.tsx
@@ -38,7 +38,7 @@ const TableMenu = <L extends null>({
       // See: https://github.com/canonical-web-and-design/react-components/issues/377
       positionNode={positionNode || undefined}
       toggleAppearance="base"
-      toggleClassName="u-no-margin--bottom p-table-menu__toggle"
+      toggleClassName="u-no-margin--bottom p-table-menu__toggle is-dense"
       toggleDisabled={disabled || false}
     />
   );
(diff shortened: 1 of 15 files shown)


ui/src/app/base/components/TableMenu/TableMenu.tsx
```diff
@@ -38,7 +38,7 @@ const TableMenu = <L extends null>({
       // See: https://github.com/canonical-web-and-design/react-components/issues/377
       positionNode={positionNode || undefined}
       toggleAppearance="base"
-      toggleClassName="u-no-margin--bottom p-table-menu__toggle"
+      toggleClassName="u-no-margin--bottom p-table-menu__toggle is-dense"
```
Caleb-Ellis, line 41: I'm not sure about adding this as a drive-by fix. We use `TableMenu` in quite a few places so there's potential for this to have an unintended side effect *somewhere*. Plus it kinda makes the button oblong shaped instead of square.

In other places we just make the dropdown disabled if no actions are available, which should address any jumping around.

ui/src/app/machines/views/MachineDetails/MachineTests/MachineTestsDetails/MachineTestsDetails.tsx
```diff
@@ -61,8 +63,8 @@ const MachineTestsDetails = (): JSX.Element | null => {
           <Col size="8">
             <h2 className="p-heading--four">{result.name} details</h2>
           </Col>
-          <Col size="4">
-            <Link to={`/machine/${id}/testing`}>
+          <Col size="4" className="u-align--right">
```
Caleb-Ellis, line 66: Nice. That was bothering me a little bit hahaha

Caleb-Ellis (changes requested):
Just a couple of QA issues:
- Some of the icons are different for the list view and the details view. I think `getIcon()` needs to be used in both?
- Logs are only loaded for the first details page you visit. It looks like one of the conditionals in a `useEffect()` needs to change to  `if (!(logs && logs[id]) && result) {` otherwise it's only checking if there are *any* logs in state.

ui/src/app/base/components/TableMenu/TableMenu.tsx
anthonydillon: Good point, reverted.

anthonydillon:
@Caleb-Ellis thanks for the points. I think I have addressed them all.

Caleb-Ellis (approved):
LGTM! There's a conflict in the betterer results but you can just merge master and run `yarn betterer --update` to fix it

PR #2177, opened 2021-02-16
feat(ui): allow LXD pods to compose machines with multiple disks

### LINKED ISSUE #2135: Support composing LXD VMs with multiple disks in the UI
Bug originally filed by ack at https://bugs.launchpad.net/bugs/1914590

MAAS now supports composing LXD VMs with multiple disks.

We should enable the feature in the UI (currently the button to add extra disks is disabled with a message).

Caleb-Ellis:
## Done

- Removed the conditional that prevented LXD pods from composing machines with multiple disks

### MAAS deployment

- [Bolla](/HACKING.md#bolla)

### Running the branch

You can run this branch by:

### QA steps

- Go to the KVM details page of a LXD KVM.
- Choose "Compose" from the "Take action" menu and check that the "Add disk" button is not disabled
- Add one or more disks then click compose
- Check that the machine is created with the correct number of disks. Note that there is currently a bug in the backend where the storage sizes are wrong while the machine is commissioning, and the boot disk + tags do not persists properly through commissioning.
- Check that after commissioning is finished the disks all have the correct sizes.

## Fixes

Fixes #2135 

## Launchpad issue

[lp#1915715](https://bugs.launchpad.net/maas/+bug/1915715)

### DIFF
--- a/ui/src/app/kvm/components/KVMActionFormWrapper/ComposeForm/StorageTable/StorageTable.test.tsx
+++ b/ui/src/app/kvm/components/KVMActionFormWrapper/ComposeForm/StorageTable/StorageTable.test.tsx
@@ -88,21 +88,6 @@ describe("StorageTable", () => {
     });
   });
 
-  it("disables add disk button with tooltip if KVM type is LXD", () => {
-    const pod = podDetailsFactory({ id: 1, type: "lxd" });
-    const state = { ...initialState };
-    state.pod.items = [pod];
-    const store = mockStore(state);
-    const wrapper = generateWrapper(store, pod);
-
-    expect(wrapper.find("[data-test='add-disk'] button").prop("disabled")).toBe(
-      true
-    );
-    expect(wrapper.find("[data-test='add-disk']").prop("message")).toBe(
-      "For the Beta version of LXD VM hosts each VM can only be assigned a single block device."
-    );
-  });
-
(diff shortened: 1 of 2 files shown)


huwshimi:
@Caleb-Ellis a couple of issues when QAing. Not totally sure if it was something I was doing wrong or a problem with the PR.

When I tried to create a second disk there was limited space left, but when I made the size smaller than the available space it still had an error:
Screen Shot 2021-02-17 at 8 47 37 am

If I add a disk and then remove it the compose button stays disabled:
Screen Shot 2021-02-17 at 8 48 16 am

Caleb-Ellis:
@huwshimi it's a bit confusing, but the validation is correct. There's only 5.49GB available in the `default` storage pool, but you've requested 8 + 4 = 12GB. The error only shows in the last size field otherwise it's repeated for every disk from that storage pool. I think it's a bit more obvious what's going on if there are more storage pools to choose from, but bolla only has the one.

A similar thing is happening on your second point - it's correct that the submit is disabled, but the error isn't showing. I think it's how the formik forms only seem to run validation on blur, so this scenario would only happen if you had a LXD pod that had less than 8GB free (because you could add and remove a disk without ever having touched the size field, but it would be invalid right from the start)

huwshimi (approved):

PR #2182, opened 2021-02-16
feat(ui): add action for creating a physical interface

huwshimi:
## Done

- Add support for creating physical interfaces via the api.
- Add selector for fetching the creating interface status.

## QA

N/A

## Fixes

Fixes: #2181.

### DIFF
--- a/ui/src/app/store/machine/actions.test.ts
+++ b/ui/src/app/store/machine/actions.test.ts
@@ -1,4 +1,5 @@
 import { actions } from "./slice";
+import { NetworkLinkMode } from "./types";
 
 import { NodeActions } from "app/store/types/node";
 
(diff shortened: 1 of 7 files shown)


ui/src/app/store/machine/slice.ts
```diff
@@ -28,6 +29,24 @@ export type ScriptInput = {
   [x: string]: { url: string };
 };
 
+const generateParams = <P extends { [x: string]: unknown }>(
```
Caleb-Ellis, line 32: Nice. Is the plan to have a type for each action's params? Might be time to break up the types file :grimacing:

Caleb-Ellis (approved):
LGTM!

ui/src/app/store/machine/actions.test.ts
```diff
@@ -616,6 +617,72 @@ describe("machine actions", () => {
     });
   });
 
+  it("can handle creating a physical interface", () => {
+    expect(
+      actions.createPhysical({
+        enabled: true,
+        interface_speed: "10",
```
Caleb-Ellis, line 624: Shouldn't `interface_speed` and `link_speed` be numbers?

ui/src/app/store/machine/slice.ts
huwshimi: That could be helpful, but this at least should remove the need for all of the optional param spreads e.g.
```typescript
...("partitionIds" in params && {
  add_partitions: params.partitionIds,
}),
```

ui/src/app/store/machine/actions.test.ts
huwshimi: Thanks, yes.

ui/src/app/store/machine/slice.ts
Caleb-Ellis: Yep that was getting a bit unwieldy.

PR #2184, opened 2021-02-17
feat(ui): add interface name generator

huwshimi:
## Done

- Add a util to generate the next available interface name.

## QA

N/A

## Fixes

Fixes: #2183.

### DIFF
--- a/ui/src/app/store/machine/utils/networking.ts
+++ b/ui/src/app/store/machine/utils/networking.ts
@@ -407,7 +407,7 @@ export const isInterfaceConnected = (
   return nic.link_connected;
 };
 
-const LINK_MODE_DISPLAY = {
+export const LINK_MODE_DISPLAY = {
   [NetworkLinkMode.AUTO]: "Auto assign",
   [NetworkLinkMode.DHCP]: "DHCP",
   [NetworkLinkMode.LINK_UP]: "Unconfigured",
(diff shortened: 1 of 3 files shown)


ui/src/app/store/machine/utils/networking.ts
```diff
@@ -614,3 +614,34 @@ export const getRemoveTypeText = (
...
   }
 };
+
+/**
+ * Find the next available name for an interface.
+ * @param machine - A machine.
+ * @param interfaceType - A network interface type.
+ * @return An available name.
+ */
+export const getNextNicName = (
```
Caleb-Ellis, line 624: Awesome! I did the same thing but in like 6 different places in the storage tab - maybe if we get some time at the end of the cycle we should align a lot of these things.

Caleb-Ellis (approved):
LGTM!

ui/src/app/store/machine/utils/networking.ts
huwshimi: If you remember where you did it would you mind filing an issue? I just had a quick look and didn't spot them.

Caleb-Ellis: Done

https://github.com/canonical-web-and-design/maas-ui/issues/2188

PR #2186, opened 2021-02-17
feat(ui): add generic networking fields

huwshimi:
## Done

- Add a number of fields for networking forms.
- Update the add machine form to use the mac address
field.

### MAAS deployment

- [Bolla](/HACKING.md#bolla)

### Running the branch

You can run this branch by:

### QA steps

- Go to the machine list and open the add machine form.
- Check that the mac address fields works as before.

## Fixes

Fixes: #2185.


ui/src/app/base/components/FabricField/FabricField.tsx
```diff
@@ -0,0 +1,40 @@
...
+  if (!fabricsLoaded) {
+    return <Spinner />;
+  }
+
+  return (
+    <FormikField
+      component={Select}
+      label="Fabric"
+      name="fabric"
+      options={fabrics.map((fabric) => ({
```
Caleb-Ellis, line 31: In most selects I've included a disabled default option that just says "Select foo" - it was a carry over from the angular version but we should at least keep them consistent

ui/src/app/base/components/SubnetField/SubnetField.tsx
```diff
@@ -0,0 +1,59 @@
...
+  filterFunction?: (subnet: Subnet) => boolean;
+} & Partial<FormikFieldProps>;
+
+export const SubnetField = ({
+  defaultOption,
+  filterFunction,
+  ...props
+}: Props): JSX.Element => {
+  const dispatch = useDispatch();
+  let subnets: Subnet[] = useSelector(subnetSelectors.all);
```
Caleb-Ellis, line 24: Does this need to be typed?

Caleb-Ellis (changes requested):
Mostly looks good, my only change request is that `name` be required on all the formik fields. The rest are mainly just to keep things consistent with the other generic fields, but that can come later maybe when we've decided how everything *should* work.

ui/src/app/base/components/SubnetField/SubnetField.tsx
huwshimi: For some reason these are coming back as `unknown` so when the `.map` happens below it says `Property 'map' does not exist on type 'unknown'`.

ui/src/app/base/components/FabricField/FabricField.tsx
huwshimi: OK, I'll add them.

ui/src/app/base/components/FabricField/FabricField.tsx
```diff
@@ -0,0 +1,40 @@
...
+import { useDispatch, useSelector } from "react-redux";
+
+import FormikField from "app/base/components/FormikField";
+import type { Props as FormikFieldProps } from "app/base/components/FormikField/FormikField";
+import { actions as fabricActions } from "app/store/fabric";
+import fabricSelectors from "app/store/fabric/selectors";
+
+type Props = Partial<FormikFieldProps>;
+
+export const FabricField = ({ ...props }: Props): JSX.Element => {
```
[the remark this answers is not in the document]
huwshimi: Sure, I'll make those changes.

ui/src/app/base/components/SubnetField/SubnetField.tsx
Caleb-Ellis: Right, I guess it could be because it's declared as a `let`, which could change to be anything.

huwshimi: Even with `let` you'd need to explicitly cast it to another type if you wanted to redeclare it with something else. I think it's more likely our selector util doesn't maintain the types for some reason.

ui/src/app/machines/components/ActionFormWrapper/TestForm/TestForm.test.tsx
```diff
@@ -157,17 +157,20 @@ describe("TestForm", () => {
   it("prepopulates scripts of a given hardwareType", () => {
     const state = { ...initialState };
     const networkScript = scriptsFactory({
+      name: "test1",
```
huwshimi, line 160: Drive-by fix for duplicate key warnings in the tests.

huwshimi:
@Caleb-Ellis this is ready for another look.

Caleb-Ellis (approved):
LGTM!

PR #2199, opened 2021-02-18
refactor(ui): make adding and updating interface actions consistent

huwshimi:
## Done

- Update the actions for creating and updating interfaces to be consistent.

### MAAS deployment

- [Bolla](/HACKING.md#bolla)

### Running the branch

You can run this branch by:

### QA steps

- Go to a machine with editable network.
- Add an interface. It should be created correctly.
- Edit the interface. It should save your changes.

## Fixes

Fixes: #2197.


anthonydillon (approved):
It's strange to mix syntax in a single code base. Could confused things. What's the reason we started using snake case?

huwshimi:
> It's strange to mix syntax in a single code base. Could confused things. What's the reason we started using snake case?

Data that comes from the API is snake case and we don't change the case before it enters the redux store. By keeping the action params the same as the API it means any errors from the API match the field names in the forms.

PR #2210, opened 2021-02-22
fix(ui): only fetch scripts if they have not been fetched before

### LINKED ISSUE #2209: UI is using API to request scripts with full content
Bug originally filed by ltrager at https://bugs.launchpad.net/bugs/1916317

While looking through the MAAS logs I noticed that the UI is repeatedly requesting

/MAAS/api/2.0/scripts/?include_script=true

This uses the API to return all scripts including their content. The API returns historic data so not only does this include script content but the content for every version of a script. The API also includes other data which the UI is not currently using.

On a fresh MAAS install with no script history the returned size is as follows:

/MAAS/api/2.0/scripts/ - 36K
/MAAS/api/2.0/scripts/?include_script=true - 258K

This causes a lot of unnecessary load on the region and alot of unnecessary data over the network which can effect slow connections. This also seems to happening repeatedly so content is not cached.

1. MAAS has a websocket to return script data. It does not currently have a way to return content but that can be easily added.
2. Script content should be loaded as needed. IMHO its better to wait a second for each script to load when viewing content then to load it all at once. I discovered this while on the nodes-listing page where you can't view script content.

Caleb-Ellis:
## Done

- Remove script fetch from machine list table because it already gets fetched from the components that need them 
- Only fetch scripts if they have not been fetched before

### MAAS deployment

- [Bolla](/HACKING.md#bolla)

### Running the branch

You can run this branch by:

### QA steps

- Open your browser's network tab
- Select some machines and open the test or commissioning action form
- Check that an http request is made to /MAAS/api/2.0/scripts/?include_script=true
- Close the form and reopen it
- Go to Settings > Commissioning scripts
- Go to Settings > Testing scripts
- Check that the http request is not made again

## Fixes

Fixes #2209 

## Launchpad issue

[lp#1916317](https://bugs.launchpad.net/bugs/1916317)

### DIFF
--- a/ui/src/app/machines/components/ActionFormWrapper/CommissionForm/CommissionForm.test.tsx
+++ b/ui/src/app/machines/components/ActionFormWrapper/CommissionForm/CommissionForm.test.tsx
@@ -72,8 +72,9 @@ describe("CommissionForm", () => {
     });
   });
 
-  it("fetches the necessary data on load", () => {
+  it("fetches scripts if they haven't been loaded yet", () => {
     const state = { ...initialState };
+    state.scripts.loaded = false;
     const store = mockStore(state);
     mount(
       <Provider store={store}>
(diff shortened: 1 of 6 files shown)


huwshimi (approved):

huwshimi:
Looks good, thanks @Caleb-Ellis.

PR #2244, opened 2021-02-23
feat(ui): remove RSD components from UI

### LINKED ISSUE #2097: Drop RSD pods UI
Bug originally filed by ack at https://bugs.launchpad.net/bugs/1913464

We want to drop support for RDS pods in the backend, since we only ever supported a now very old version of RSD.

We should first drop the UI part, so the backend can then be dropped.

Caleb-Ellis:
## Done

- Removed RSD components from UI since it will no longer be supported from 2.10

### MAAS deployment

- [Bolla](/HACKING.md#bolla)

### Running the branch

You can run this branch by:

### QA steps

- With a MAAS that has RSD pods (`make sampledata` creates some), check that there is longer a navigation item for RSD in the angular or react apps
- Check that composing machines from a virsh/lxd pod still works properly
- Go to the machine list and click "Add hardware"
- Check that there is no option for RSD

## Fixes

Fixes #2097


huwshimi (approved):
This looks good. I did notice there are a couple of cases where `rsd` was left in the code, were any of these deliberate?

Caleb-Ellis:
@huwshimi oh I wasn't super thorough with the legacy side of things, generally try to avoid changing it as much as possible haha. Looks like I missed a case in `formatHostType` but I think the rest can stay since they're all in tests

PR #2245, opened 2021-02-23
feat(ui): add menu item for adding an alias or VLAN

huwshimi:
## Done

- Add an item to the row menu in the network table to add an alias or VLAN.

### MAAS deployment

- [Bolla](/HACKING.md#bolla)

### Running the branch

You can run this branch by:

### QA steps

- View both the react and legacy network details for a machine.
- Open the menu for different rows and you should see an item for adding an alias or a vlan in the same cases as the legacy page.

## Fixes

Fixes: #2215.

### DIFF
--- a/ui/.betterer.results
+++ b/ui/.betterer.results
@@ -436,8 +436,8 @@ exports[`stricter compilation`] = {
       [357, 35, 9, "Object is possibly \'null\'.", "1794230341"],
       [357, 47, 9, "Object is possibly \'null\'.", "1612642726"]
     ],
-    "src/app/machines/views/MachineDetails/MachineNetwork/NetworkTable/NetworkTableActions/NetworkTableActions.tsx:136171821": [
-      [79, 8, 5, "Object is possibly \'null\'.", "179721187"]
+    "src/app/machines/views/MachineDetails/MachineNetwork/NetworkTable/NetworkTableActions/NetworkTableActions.tsx:2132094581": [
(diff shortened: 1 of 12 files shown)


Caleb-Ellis (commented):
@huwshimi a couple of discrepancies I've noticed on great-tapir on bolla:
- All aliases have the action in react. but none do in legacy
- All VLANs have the action in react, but some don't in legacy
- Interfaces on fabrics with only the default VLAN have the action in react, but not in legacy

More broadly though - do you think it makes sense to lump together adding an alias and VLAN together under the same action? From a quick skim of the legacy code and the actual forms themselves they seem quite different.

huwshimi:
> @huwshimi a few discrepancies I've noticed on great-tapir on bolla:
> 
> * All aliases have the action in react. but none do in legacy
> * All VLANs have the action in react, but some don't in legacy
> * Interfaces on fabrics with only the default VLAN have the action in react, but not in legacy

Thanks, I've fixed these three issues (I was forgetting to include the link when determining the type of the interface).

> More broadly though - do you think it makes sense to lump together adding an alias and VLAN together under the same action? From a quick skim of the legacy code and the actual forms themselves they seem quite different.

Let's have a chat about this later today.

huwshimi:
@Caleb-Ellis I've separated the two actions.

Caleb-Ellis (approved):
Nice! I think it's much easier to follow now

PR #2279, opened 2021-03-01
feat(ui): add disabled alias or vlan action tooltips

huwshimi:
## Done

- Add a tooltip when adding an alias or vlan is disabled.

### MAAS deployment

- [Bolla](/HACKING.md#bolla)

### Running the branch

You can run this branch by:

### QA steps

- Go to the network tab of a machine.
- Create (or edit) a physical interface and choose a subnet then set the IP mode to be unconfigured and submit.
- Now visit the legacy page of the interface and using the row's action menu add VLANs until it won't let you add any more.
- Go back to the react page and open the action menu for the interface and the add vlan/alias items should be disabled and there should be tooltips with the reasons why.

## Fixes

Fixes: #2278.


Caleb-Ellis (approved):
Nice! I think it looks great!

PR #2295, opened 2021-03-02
feat(ui): add api support for creating a bond

huwshimi:
## Done

- Add the action to be able to create a bond via the API.

## QA

N/A

## Fixes

Fixes: #2218.


anthonydillon (approved):
LGTM, nice to see how this part of the app works :+1:

Issue #2213, opened 2021-02-22
State changes in machine details result in websocket requests


huwshimi:
When doing something that results in a state change (e.g. changing between react tabs) then you get a heap of requests over the websocket.

<img width="422" alt="Screen Shot 2021-02-22 at 4 44 19 pm" src="https://user-images.githubusercontent.com/361637/108667551-95139a00-752d-11eb-8ff8-3172914e338a.png">

PR #2303, opened 2021-03-03
feat(ui): edit aliases and vlans form

huwshimi:
## Done

- Add a form to edit aliases and vlans on the network tab.

### MAAS deployment

- [Bolla](/HACKING.md#bolla)

### Running the branch

You can run this branch by:

### QA steps

- View the machine details network tab for a machine with editable network.
- Create a VLAN and an alias if it doesn't already have them.
- Using the action menu click edit vlan/alias.
- Change the details and save.
- The row should update for the vlan/alias.

## Fixes

Fixes: #2225.


anthonydillon:
## QA
I assume this will fail as its auto-selected the same name as an existing alias:
image

It didn't but I assume the second one should be named `enp5s0:2`
image

ui/src/app/machines/views/MachineDetails/MachineNetwork/EditInterface/EditInterface.test.tsx
```diff
@@ -71,6 +72,61 @@ describe("EditInterface", () => {
...
+      <Provider store={store}>
+        <MemoryRouter
+          initialEntries={[{ pathname: "/machines", key: "testKey" }]}
+        >
+          <EditInterface systemId="abc123" close={jest.fn()} nicId={nic.id} />
+        </MemoryRouter>
+      </Provider>
+    );
+    expect(wrapper.find("EditAliasOrVlanForm").exists()).toBe(true);
+    expect(wrapper.find("FormCard").prop("title")).toBe("Edit VLAN");
```
anthonydillon, line 130: Nice way to detect which one is being displayed but slightly coupled to copy changes :man_shrugging:

anthonydillon (approved):
Code and QA ok with a few questions in the QA

huwshimi:
> ## QA
> ### Naming issue?
> I assume this will fail as its auto-selected the same name as an existing alias:

Are these the two names you're referring to?
unnamed
If so, they have different types, one is an alias and the other is a VLAN, and the names are slightly different, the VLAN has a period before the number and the alias has a colon but they both increment independently of the other type.


> ### Closing edit panel slow
> If you open an edit panel and click "Cancel" the UI takes a while to respond.
> Peek 2021-03-03 08-30

For some reason there are a bunch of requests being made when doing some state changes, this is being tracked here: https://github.com/canonical-web-and-design/maas-ui/issues/2213.

> ### Legacy not loading
> Using bolla the legacy Networking tab is not working: http://localhost:8400/MAAS/l/machine/67ds3b?area=network.

Were you running the full client or `dotrun ui`? If just `ui` then the legacy client isn't being served. If you want to compare you can run the full client or visit bolla directly.

ui/src/app/machines/views/MachineDetails/MachineNetwork/EditInterface/EditInterface.test.tsx
huwshimi: I think this might be OK as we do want to check that we're updating the title with the type of the interface, I guess it could just check for `VLAN` in the string...

anthonydillon:
> If so, they have different types, one is an alias and the other is a VLAN, and the names are slightly different, the VLAN has a period before the number and the alias has a colon but they both increment independently of the other type.

Ah I see the difference now.

> For some reason there are a bunch of requests being made when doing some state changes, this is being tracked here: #2213.

:+1: 

> Were you running the full client or dotrun ui? If just ui then the legacy client isn't being served. If you want to compare you can run the full client or visit bolla directly.

Absolutely, my bad.

ui/src/app/machines/views/MachineDetails/MachineNetwork/EditInterface/EditInterface.test.tsx
anthonydillon: Not sure its worth it in this case but something to keep in mind.

PR #2326, opened 2021-03-05
feat(ui): add the bridge form table

huwshimi:
## Done

- Added a component to display the bridge table.

### MAAS deployment

- [Bolla](/HACKING.md#bolla)

### Running the branch

You can run this branch by:

### QA steps

- Navigate to the react network tab for a machine.
- Select an interface and click the "Create bridge" button.
- A table should appear displaying the selected interface's details.

## Fixes

Fixes: #2307.

### DIFF
--- a/ui/src/app/machines/views/MachineDetails/MachineNetwork/InterfaceFormTable/InterfaceFormTable.tsx
+++ b/ui/src/app/machines/views/MachineDetails/MachineNetwork/InterfaceFormTable/InterfaceFormTable.tsx
@@ -1,6 +1,6 @@
 import type { ReactNode } from "react";
 
-import { MainTable, Spinner } from "@canonical/react-components";
+import { Icon, MainTable, Spinner } from "@canonical/react-components";
 import { useSelector } from "react-redux";
 
 import DHCPColumn from "../NetworkTable/DHCPColumn";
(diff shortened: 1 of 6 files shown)


ui/src/app/machines/views/MachineDetails/MachineNetwork/InterfaceFormTable/InterfaceFormTable.tsx
```diff
@@ -119,7 +126,9 @@ const InterfaceFormTable = ({
         {
           content: (
             <>
-              <TableHeader className="u-align--center">PXE</TableHeader>
+              <TableHeader className="u-align--center">
+                {isPrimary ? "Primary" : "PXE"}
```
Caleb-Ellis, line 130: As I understand it `Primary` only makes sense for bonds, since you're determining which of the bond parents' details you want to assign to the new bond.

Caleb-Ellis (approved):
I've approved because it's the same as legacy, but I think the `Primary` header in the bridge form table is actually a bug.

huwshimi:
> I've approved because it's the same as legacy, but I think the `Primary` header in the bridge form table is actually a bug.

Thanks, I did wonder about that. I'll remove the primary bit when I rebase with my bond branch as this section is rewritten in that branch.

PR #2352, opened 2021-03-10
fix(ui): add proxmox field names to add_chassis parameter map

### LINKED ISSUE #2327: Unable to add Proxmox chassis with token
Bug originally filed by ltrager at https://bugs.launchpad.net/bugs/1917953

I recently added Proxmox support to MAAS for a customer. Support has been backported to MAAS 2.9.2. Proxmox allows you to authenticate with either a password or a token. Adding a Proxmox chassis with a password works in the UI. If you try to add a Proxmox chassis with a token the UI insists a password must be given.

The power driver correctly defines which fields are required. power_pass, power_token_name, and power_token_secret all have required=False. When adding an individual machine the UI does the right thing and allows either to be set. Its only the add chassis form which is insisting a password must be set.

This fix should be backported to 2.9.

Caleb-Ellis:
## Done

- Added proxmox field names to the add_chassis parameter map

### MAAS deployment

- [Bolla](/HACKING.md#bolla)

### Running the branch

You can run this branch by:

### QA steps

- Make sure you're running the UI using `yarn start` not `yarn ui` because `add_chassis` is an http method
- Open the add chassis form from the machine list and choose Proxmox
- Submit the form without a password or token and check that you get an error
- Submit the form with both a password and token and check that you get an error
- Check the network tab and see that the add_chassis request includes all the entered form values

## Fixes

Fixes #2327

### DIFF
--- a/ui/src/app/store/general/utils/powerTypes.test.ts
+++ b/ui/src/app/store/general/utils/powerTypes.test.ts
@@ -36,7 +36,10 @@ describe("powerTypes utils", () => {
           powerFieldFactory({ name: "power_pass" }),
           powerFieldFactory({ name: "power_port" }),
           powerFieldFactory({ name: "power_protocol" }),
+          powerFieldFactory({ name: "power_token_name" }),
+          powerFieldFactory({ name: "power_token_secret" }),
           powerFieldFactory({ name: "power_user" }),
+          powerFieldFactory({ name: "power_verify_ssl" }),
         ],
       });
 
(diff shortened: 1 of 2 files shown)


anthonydillon (approved):
LGTM :+1:

anthonydillon:
Might be good to add a test or two to cover the QA steps for the future.

Caleb-Ellis:
@anthonydillon those would be end-to-end tests though since we'd be testing the api responses, and we're not doing Cypress tests at the moment since we don't have the time. I think it's ok (not ideal) for now to just check that the parameter map works as expected. :man_shrugging:

PR #2367, opened 2021-03-12
feat(ui): edit bond form

huwshimi:
## Done

- Add a form for editing bonds.

### MAAS deployment

- [Bolla](/HACKING.md#bolla)

### Running the branch

You can run this branch by:

### QA steps

- View the react network tab for a machine with a bond (or create a bond by adding two physical interfaces on the same VLAN and then clicking 'Create bond').
- Add another physical interface on the same vlan as the bond.
- Use the row action menu to open the edit bond form.
- The green save button should be disabled.
- Click 'Edit bond members' and select a different combination of interfaces. The green save button should become enabled.
- Change the bond members back to the original combination and the save button should become disabled.
- Change some other fields and click save.
- The bond should update with the new info.

## Fixes

Fixes: #2305.

### DIFF
--- a/ui/.betterer.results
+++ b/ui/.betterer.results
@@ -76,8 +76,8 @@ exports[`stricter compilation`] = {
     "src/app/base/components/FormikForm/FormikForm.tsx:1896345427": [
       [64, 4, 5, "Argument of type \'boolean | undefined\' is not assignable to parameter of type \'boolean\'.\\n  Type \'undefined\' is not assignable to type \'boolean\'.", "195688512"]
     ],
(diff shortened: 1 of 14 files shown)


Caleb-Ellis:
@huwshimi everything seems to save properly except for bond mode - I think it might just need to set a default option in the edit form?

huwshimi:
> @huwshimi everything seems to save properly except for bond mode - I think it might just need to set a default option in the edit form?

Thanks, it was being saved, but I was restoring the value in the select box.

Caleb-Ellis (approved):
LGTM!

Is that it for network?!?

huwshimi:
> LGTM!
> 
> Is that it for network?!?

It sure is!

PR #2368, opened 2021-03-12
feat(ui): add analytics for editing bond members

huwshimi:
## Done

- Add analytics to the bond forms when users edit the bond members.

## QA

N/A

## Fixes

Fixes: #2185.


Caleb-Ellis (approved):
LGTM!

PR #2407, opened 2021-03-18
feat(ui): add event slice

huwshimi:
## Done

- Add the slice and selectors for the events model.

## QA

N/A

## Fixes

Fixes: canonical-web-and-design/maas-squad#2366.

### DIFF
--- a/ui/src/app/store/machine/types.ts
+++ b/ui/src/app/store/machine/types.ts
@@ -176,7 +176,8 @@ export type EventType = Model & {
   name: string;
 };
 
-export type Event = Model & {
+// This is named `MachineEvent` as there is already a DOM `Event` type.
+export type MachineEvent = Model & {
   created: string;
   description: string;
   type: EventType;
@@ -275,7 +276,7 @@ export type MachineDetails = BaseMachine & {
   disks: Disk[];
   error_description: string;
   error: string;
-  events: Event[];
+  events: MachineEvent[];
   grouped_storages: {
     count: number;
     disk_type: string;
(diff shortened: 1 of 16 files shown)


ui/src/testing/factories/event.ts
```diff
@@ -0,0 +1,26 @@
...
+export const eventRecord = extend<Model, EventRecord>(model, {
+  action: "action-h7Jza3",
+  created: "Tue, 16 Mar. 2021 03:04:00",
+  description: "desc-LRv9EV",
+  endpoint: 1,
+  ip_address: "152.26.96.176",
+  node_hostname: "honest-bear",
+  node_id: 11,
+  node_system_id: "tftrtd",
+  type: () => ({
```
Caleb-Ellis, line 17: Should this be it own factory?

ui/src/testing/factories/event.ts
```diff
@@ -0,0 +1,26 @@
...
+    description: "description-2HCo4f",
+    level: "info",
```
Caleb-Ellis, line 19: This could be an enum. From the source code it looks like it can be one of `"audit"`, `"critical"`, `"debug"`, `"error"`, `"info"` or `"warning"`.

ui/src/app/store/machine/types.ts
```diff
@@ -176,7 +176,8 @@ export type EventType = Model & {
   name: string;
 };
 
-export type Event = Model & {
+// This is named `MachineEvent` as there is already a DOM `Event` type.
+export type MachineEvent = Model & {
```
Caleb-Ellis, line 180: I wonder if we should be prefixing all the machine types with `Machine` anyway, and do that with all the models.

Caleb-Ellis (approved):
LGTM! Just a couple of suggestions with the test factories

ui/src/testing/factories/event.ts
huwshimi: Thanks, done.

PR #2431, opened 2021-03-24
build(all): don't install Cypress when building

huwshimi:
## Done

- Make it so that Cypress doesn't install when building the app.

## QA

N/A build should pass in CI.

## Fixes

Fixes: #2430.


Caleb-Ellis (approved):
LGTM!

PR #2460, opened 2021-04-01
feat(ui): display script elapsed time

huwshimi:
## Done

- Add a component to display the elapsed time of scripts.

### MAAS deployment

- [Bolla](/HACKING.md#bolla)

### Running the branch

You can run this branch by:

### QA steps

- Go to an LXD pod and compose a machine.
- Go to the machine details for the new machine and click on comissioning.
- Wait a while and you should see the scripts update and there should be elapsed times for some of them.

## Fixes

Fixes: #2311.

### DIFF
--- a/ui/src/app/machines/views/MachineDetails/MachineTests/MachineTestsTable/ScriptRunTime/ScriptRunTime.tsx
+++ b/ui/src/app/machines/views/MachineDetails/MachineTests/MachineTestsTable/ScriptRunTime/ScriptRunTime.tsx
@@ -24,10 +24,10 @@ const getElapsedTime = (starttime: ScriptResult["starttime"]) => {
   const elapsed = [];
   // Show the elapsed days if more than a day has elapsed.
   if (days) {
-    elapsed.push(`${days} ${pluralize("day", days)}`);
+    elapsed.push(pluralize("day", days, true));
   }
   // Display time in the format hh:mm:ss.
-  elapsed.push(`${zeroPad(hours)}:${zeroPad(minutes)}:${zeroPad(seconds)}`);
+  elapsed.push(`${hours}:${zeroPad(minutes)}:${zeroPad(seconds)}`);
   return elapsed.join(", ");
 };
 
--- a/ui/src/app/machines/views/MachineDetails/MachineTests/MachineTestsTable/ScriptRunTime/ScriptRunTime.tsx
+++ b/ui/src/app/machines/views/MachineDetails/MachineTests/MachineTestsTable/ScriptRunTime/ScriptRunTime.tsx
@@ -18,7 +18,7 @@ const zeroPad = (time?: number) => `0${time || "0"}`.slice(-2);
 
 const getElapsedTime = (starttime: ScriptResult["starttime"]) => {
   const { days, hours, minutes, seconds } = intervalToDuration({
-    start: fromUnixTime(starttime),
+    start: starttime ? fromUnixTime(starttime) : Date.now(),
     end: Date.now(),
   });
   const elapsed = [];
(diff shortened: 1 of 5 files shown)


Caleb-Ellis (commented):
Sometimes the elapsed time initially shows `4 days, 23:59:XX of ~X:XX:XX` before eventually circling back to `00:00:00 of ~X:XX:XX`. Do you think this could be due to an error in this PR or from the api? FWIW I've noticed the api sends back the `starttime` with decimals - think that might affect anything?

huwshimi:
> Sometimes the elapsed time initially shows `4 days, 23:59:XX of ~X:XX:XX` before eventually circling back to `00:00:00 of ~X:XX:XX`. Do you think this could be due to an error in this PR or from the api? FWIW I've noticed the api sends back the `starttime` with decimals - think that might affect anything?

It turns out `starttime` is not always present which was causing the elapsed time to be off.  I've pushed a fix to handle that case.

ui/src/app/machines/views/MachineDetails/MachineTests/MachineTestsTable/ScriptRunTime/ScriptRunTime.tsx
```diff
@@ -0,0 +1,74 @@
...
+
+const getElapsedTime = (starttime: ScriptResult["starttime"]) => {
+  const { days, hours, minutes, seconds } = intervalToDuration({
+    start: starttime ? fromUnixTime(starttime) : Date.now(),
+    end: Date.now(),
+  });
+  const elapsed = [];
+  // Show the elapsed days if more than a day has elapsed.
+  if (days) {
+    elapsed.push(`${days} ${pluralize("day", days)}`);
```
Caleb-Ellis, line 27: I think you can just use `elapsed.push(pluralize("day", days, true));`. The third argument prepends the string with the number.

ui/src/app/machines/views/MachineDetails/MachineTests/MachineTestsTable/ScriptRunTime/ScriptRunTime.tsx
```diff
@@ -0,0 +1,74 @@
...
+  }
+  // Display time in the format hh:mm:ss.
+  elapsed.push(`${zeroPad(hours)}:${zeroPad(minutes)}:${zeroPad(seconds)}`);
```
Caleb-Ellis, line 30: Maybe we can remove the extra `0` for hours? It looks like `estimated_runtime` is formatted as `h:mm:ss` so I think it makes sense to be consistent.

Caleb-Ellis (approved):
Nice, that seems to have done the trick. I wonder where the 4 days came from if the starttime was unknown?

huwshimi:
> Nice, that seems to have done the trick. I wonder where the 4 days came from if the starttime was unknown?

I think it was probably using the start of epoch time by default and what we were seeing was the last part of that (... 4 days and however many minutes etc.)

ui/src/app/machines/views/MachineDetails/MachineTests/MachineTestsTable/ScriptRunTime/ScriptRunTime.tsx
huwshimi: Thanks, I remember that now.

ui/src/app/machines/views/MachineDetails/MachineTests/MachineTestsTable/ScriptRunTime/ScriptRunTime.tsx
huwshimi: Ah yeah good point.

PR #2544, opened 2021-04-19
fix(ui): validate VM compose storage on render

### LINKED ISSUE #2338: "Compose" button disabled but no visible error if storage isn't enough
Bug originally filed by ack at https://bugs.launchpad.net/bugs/1918300

I had a test VIRSH pod deployed, with only 8Gb of disk (because it was itself a VM built with defaults).

When trying to compose a VM on it, by default the form is green and I can click "Compose", then I get an error on the disk size field, correctly saying that there's not enough space to make a 8Gb disk.

But, if instead of accepting a default I change some of the form default, for instance number of cores or memory, the "Compose" button gets disabled, and I don't see any indication of an error, unless I click on the storage "Location" dropdown for the disk, where I see the storage pool is grayed out and has an indication of not enough space.

I think in this case there should be an error visible directly on the form to why composing is not allowed.

### LINKED ISSUE #2361: Composing a LXD VM with multiple interfaces doesn't work
Bug originally filed by ack at https://bugs.launchpad.net/bugs/1918713

Trying to test composing a LXD VM with multiple interfaces from the UI, I get a machine with only one.

Steps:

1) go the LXD vM host page, select compose
2) under interfaces, click "define" and then add multiple interfaces (tested with 4). Don't change any details, just add them
3) commission

Looking at the websocket call made, I see:

{
    "method": "pod.compose",
    "type": 0,
    "params": {
        "architecture": "amd64/generic",
        "domain": 0,
        "hostname": "",
        "hugepages_backed": false,
        "id": 190,
        "interfaces": "",
        "memory": "",
        "pool": 0,
        "storage": "1:8(default)",
        "zone": 1
    },
    "request_id": 17
}

The "interfaces" value is empty, so the VM is created with one interface (the default).

### LINKED ISSUE #2497: Unable to compose LXD VM with multiple NICs
Bug originally filed by ltrager at https://bugs.launchpad.net/bugs/1923246

I used the UI to compose an LXD VM. In the UI I requested that the VM was created with two NICs. The UI even asked which space and subnet I want the NICs on.

1. The machine was composed with only 1 NIC, it silently ignored my request for a second NIC
2. During commissioning MAAS discovered my VM host has access to two networks 10.0.0.0/24 and 192.168.122.0/24. My deployment only had 10.0.0.0/24 configured, there was no configured access to 192.168.122.0/24. During composition the UI indicated I could have multiple NICs on either network even though no bridge was available to 192.168.112.0/24.

Caleb-Ellis:
## Done

- VM compose storage table is now validated on render

### MAAS deployment

- [Bolla](/HACKING.md#bolla)

### Running the branch

You can run this branch by:

### QA steps

- Open the Compose form for a KVM with less than 8GB of space and check that an error shows up by default

## Fixes

Fixes #2338

### DIFF
--- a/ui/src/app/kvm/components/KVMActionFormWrapper/ComposeForm/StorageTable/StorageTable.test.tsx
+++ b/ui/src/app/kvm/components/KVMActionFormWrapper/ComposeForm/StorageTable/StorageTable.test.tsx
@@ -29,6 +29,7 @@ import {
   vlanState as vlanStateFactory,
   zoneState as zoneStateFactory,
 } from "testing/factories";
+import { waitForComponentToPaint } from "testing/utils";
 
 const mockStore = configureStore();
 
@@ -243,5 +244,24 @@ describe("StorageTable", () => {
     expect(wrapper.find(".p-form-validation__message").text()).toBe(
       `Error: Only 25GB available in ${pool.name}.`
     );
+  });
+
+  it("displays an error message on render if not enough space", async () => {
+    const pool = podStoragePoolFactory({ available: 0, name: "pool" });
+    const pod = podDetailsFactory({
+      id: 1,
+      default_storage_pool: pool.id,
+      storage_pools: [pool],
+    });
+    const state = { ...initialState };
+    state.pod.items = [pod];
+    const store = mockStore(state);
+    const wrapper = generateWrapper(store, pod);
+    await waitForComponentToPaint(wrapper);
+    expect(
+      wrapper
+        .find("FormikField[name='disks[0].size'] .p-form-validation__message")
+        .text()
+    ).toBe("Error: Only 0GB available in pool.");
   });
 });
--- a/ui/src/app/kvm/components/KVMActionFormWrapper/ComposeForm/StorageTable/StorageTable.tsx
+++ b/ui/src/app/kvm/components/KVMActionFormWrapper/ComposeForm/StorageTable/StorageTable.tsx
@@ -1,4 +1,4 @@
-import * as React from "react";
+import { useEffect } from "react";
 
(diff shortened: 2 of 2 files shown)


huwshimi:
This works, but maybe it's a bit confusing to have the error appear below the ip address field as it makes it seem like you need to fill in that field (but the help text says "Leave empty to auto-assign").

Screen Shot 2021-04-19 at 1 34 43 pm

Caleb-Ellis:
Yeah I suppose you're right. I wonder if we should get @amylily1011's opinion on this

huwshimi:
@amylily1011 maybe you have some thoughts on this bit of interface. Each interface must choose at least one ip address, space or subnet. I don't think we have a common pattern for multiple dependant fields.

Maybe each field could be in the error state until you choose an ip, space or subnet, but I'm not sure where the error should go so that it doesn't look like it belongs to just one of those fields (like it does now).

Screen Shot 2021-04-19 at 5 41 34 pm

amylily1011:
So according to Alberto it looks like the real error here is that you need to choose a subnet to connect it to when you create multiple interfaces. So I think we can do something like this:
- Replace 'Any' with 'Select' and only show the error message for the subnets
Screen Shot 2021-04-19 at 3 57 19 PM

Caleb-Ellis:
@amylily1011 ah ok, that sounds like a plan. I'll update this PR to only address the storage validation because I think the interface validation will take a little bit of time

Caleb-Ellis:
@huwshimi I've updated this so that it's just address the storage validation now so I think it's ready for review

huwshimi (approved):
Looks good.

PR #2553, opened 2021-04-21
fix(ui): fix KVM list resource usage calculations

### LINKED ISSUE #2512: Fix KVM list resource usage calculations
Bug originally filed by ack at https://bugs.launchpad.net/bugs/1923871

In a LXD VM host detail page, both on the "Project" and "Resources" tab, the graph seems to show usage across the whole host, not just the project.

Overall graphs correctly distinguish between tracked project and others.

In the attached screenshot I have 
 - 2 VMs in the tracked project each with 2 cores and 2Gb memory
 - 1 VM in another project, with 1 core and 1Gb memory

The overall RAM graph is correct, but the cores one only shows the project usage
The project graphs are showing resources for the whole host, not just the project.

Caleb-Ellis:
## Done

- Fixed KVM list resource usage, taking resource over-commit into account
- Changed the colours for the Overall charts so that Project is blue and Others is green. I think this makes more sense because it's consistent with everywhere else that uses blue to mean the resources used by the VM host itself.

### MAAS deployment

- [Bolla](/HACKING.md#bolla)

### Running the branch

You can run this branch by:

### QA steps

- Go to the KVM list and take note of the values that show for Cores and RAM (hover over the chart for actual numbers)
- Check that the numbers match what's found in the KVM details Resources tab in the Overall card.

## Fixes

Fixes #2512

### DIFF
--- a/ui/src/app/kvm/views/KVMList/CPUColumn/CPUPopover/CPUPopover.tsx
+++ b/ui/src/app/kvm/views/KVMList/CPUColumn/CPUPopover/CPUPopover.tsx
@@ -1,22 +1,25 @@
-import PropTypes from "prop-types";
+import type { ReactNode } from "react";
 
 import Popover from "app/base/components/Popover";
+import type { Pod, PodResource } from "app/store/pod/types";
+import { resourceWithOverCommit } from "app/store/pod/utils";
 
 type Props = {
-  allocated: number;
-  children: JSX.Element | string;
-  physical: number;
-  overcommit: number;
+  children: ReactNode;
+  cores: PodResource;
+  overCommit: Pod["cpu_over_commit_ratio"];
 };
 
-const CPUPopover = ({
-  allocated,
-  children,
-  physical,
-  overcommit,
-}: Props): JSX.Element => {
-  const total = physical * overcommit;
-  const free = total - allocated;
+const CPUPopover = ({ children, cores, overCommit }: Props): JSX.Element => {
+  const overCommitted = resourceWithOverCommit(cores, overCommit);
+  const hostCores =
+    cores.allocated_other + cores.allocated_tracked + cores.free;
+  const overCommitCores =
+    overCommitted.allocated_other +
+    overCommitted.allocated_tracked +
+    overCommitted.free;
+  const showOther = cores.allocated_other > 0;
+  const hasOverCommit = overCommit !== 1;
 
   return (
     <Popover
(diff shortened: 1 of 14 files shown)


Caleb-Ellis:
@albertodonato I just wanted to run this by you to check if I'm understanding this and fixing it right.

Firstly this PR should fix the bug where `allocated_other` was being used in the project-specific calculations. In the screenshot below, the LXD host has 8 cores, 2 of which are allocated to the first project, and 3 of which are allocated to the second project. The popover shows the calculation for the second project.


The "Overall" resources section is a bit clearer about the breakdown of resources:


I think this is fairly straightforward - basically the "Free" amount should be consistent between projects because it's reporting from the same host.

The trickier part is when over-commit is taken into consideration. The popover below again shows the calculation for the second project, with an over-commit ratio of 2. The first project stays at an over-commit ratio of 1.


In this case the "Free" amount is different for each project even though they belong to the same host. Am I using the right order of operations? I'm not entirely sure what gets multiplied by the over-commit ratio in this case. 

This also changes the "Overall" resources card, given now the over-commited project has access to more cores:


Also pinging @amylily1011 - do you think the KVM list would benefit from using the same style as the "Overall" resources card?

albertodonato:
@Caleb-Ellis thanks for the walk-through of the logic, it looks sound to me.

I guess the only thing that might look weird if you have multiple projects registered, with different overcommit values is that the overall graph would differ depending on which project you're looking at in terms of free resources, but I think it's the nature of it.

LGTM.

Caleb-Ellis:
@albertodonato awesome, thanks for that.

After talking with @amylily1011 we'll go ahead with changing the KVM list graphs to the "Overall" style, which I think will make it clearer how the resources are divided.

Caleb-Ellis:
Pinging @amylily1011
Just letting you know I've updated the colours for the Overall graph in this PR (see above) because I feel like it makes more sense. Happy to change it back if you otherwise otherwise though

ui/src/app/kvm/views/KVMList/CPUColumn/CPUPopover/CPUPopover.tsx
```diff
@@ -1,22 +1,25 @@
...
-  children,
-  physical,
-  overcommit,
-}: Props): JSX.Element => {
-  const total = physical * overcommit;
-  const free = total - allocated;
+const CPUPopover = ({ children, cores, overCommit }: Props): JSX.Element => {
+  const overCommitted = resourceWithOverCommit(cores, overCommit);
+  const hostCores =
+    cores.allocated_other + cores.allocated_tracked + cores.free;
```
huwshimi, line 16: Am I right in thinking this is the same calculation that is done in 'CPUColumn'? If so is there are way we can do these calculations in a util or hook so we can just pass a pod and know we're getting the same results everywhere?

ui/src/app/kvm/views/KVMList/RAMColumn/RAMPopover/RAMPopover.tsx
```diff
@@ -1,33 +1,31 @@
...
-  const physicalMemory = formatBytes(physical, "MiB", {
-    binary: true,
-  });
-  const allocatedMemory = formatBytes(allocated, "MiB", {
-    binary: true,
-  });
-  const freeMemory = formatBytes(physical * overcommit - allocated, "MiB", {
-    binary: true,
-  });
+const RAMPopover = ({ children, memory, overCommit }: Props): JSX.Element => {
```
huwshimi, line 14: Same thing with the CPU components, can we dedupe any of the calculations to prevent errors in the future?

huwshimi (approved):
This looks good, a couple of comments about consolidating the calculations if we can (probably as followup, this branch is already quite large).

Also, comparing the columns and popover markup/css for the ram and cpu versions of each component there seem to be a lot of similarities. As another followup it might be worth considering creating a column and popover presentation component which is wrapped by the ram/cpu components that pass the correct props.

Caleb-Ellis:
@huwshimi yep, good ideas. I'll just add one issue to reduce the duplications for now, and if it turns out to be large I'll break it out into two (resource calc utils + KVM popover component)

Issue #858, opened 2020-03-03
Error message displaying mark-up


anthonydillon:
**Describe the bug**
When logging in to maas-ui on port 5401, the ui renders a broken login screen. It would be nice to have a message in development suggesting that 5400 is the correct port.

**MAAS version**
Master

**To Reproduce**
1. Point the MAAS_URL to karura
2. Run `dotrun`
3. Try to login

**Expected behavior**
The error should just show the message within the HTML.

**Screenshots**
![image](https://user-images.githubusercontent.com/1413534/75759017-c5ff2b80-5d2c-11ea-97eb-753d9a3bf5c3.png)

squidsoup:
Hey @anthonydillon, I can't reproduce this on master.

anthonydillon:
This was due to using the wrong port in development. Still worth fixing but low priority.

PR #2571, opened 2021-04-26
feat(proxy): dev console messages

huwshimi:
## Done

- Display message when loading clients in the console.

### MAAS deployment

- [Bolla](/HACKING.md#bolla)

### Running the branch

You can run this branch by:

### QA steps

- Run `dotrun` or `yarn start`.
- At the start you should see a red message about waiting for the clients to load.
- After all the clients have started you should see a message with the URL to load.
- Visiting that URL should show you the app.
- Run `dotrun ui` or `yarn ui`.
- Shortly you should see a message about waiting for the React client to load.
- When it loads visit the URL and the app should be ready.

## Fixes

Fixes: #858.

### DIFF
--- a/proxy/package.json
+++ b/proxy/package.json
@@ -17,12 +17,16 @@
     "serve-ui": "cd ../ui && BROWSER=none PORT=8401 yarn run start",
     "serve-ui-proxy": "node ./ui.js",
     "serve-ui-standalone": "yarn build-root-css && cd ../ui && BROWSER=none PORT=8401 yarn run standalone",
-    "start": "yarn build-shared && concurrently \"yarn serve-apps\" \"yarn serve-base-wait\"",
-    "start-ui": "concurrently \"yarn serve-ui-standalone\" \"yarn watch-shared\" \"yarn serve-ui-proxy\"",
+    "show-ready": "wait-on http://0.0.0.0:8404 && node ./scripts/ready.js",
+    "show-ready-ui": "wait-on http-get://0.0.0.0:8401 && node ./scripts/ui.js",
+    "start": "node ./scripts/start.js && yarn build-shared && concurrently \"yarn serve-apps\" \"yarn serve-base-wait\" \"yarn show-ready\"",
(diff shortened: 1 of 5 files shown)


Caleb-Ellis (commented):
This is a bit fancy! I did notice there were a ton of `wait until bundle is finished` messages that didn't show up before? It doesn't make anything slower or change anything really but it's a bit weird. Do you think we should not be auto-opening a window if the message shows the correct address now?

huwshimi:
> This is a bit fancy! I did notice there were a ton of `wait until bundle is finished` messages that didn't show up before? It doesn't make anything slower or change anything really but it's a bit weird.

Yeah I think thats from `wait-on` polling the url to see if it's ready and the server rejecting it (maybe from the proxy). I don't really think we can do anything about the warnings, but we might be able to filter out them out from being displayed.

> Do you think we should not be auto-opening a window if the message shows the correct address now?

We could, but maybe I'm just old and grumpy but I find those things a bit annoying, I usually already have a tab for the UI open and then I have to go to all the trouble of closing the tab that opens, IT RUINS MY DAY.

Caleb-Ellis:
> Yeah I think thats from wait-on polling the url to see if it's ready and the server rejecting it (maybe from the proxy). I don't really think we can do anything about the warnings, but we might be able to filter out them out from being displayed.

That must be it. Could filter them out as a follow-up but it's probably not worth the effort, plus the (minimal) added risk of filtering out something we actually want to show.

> We could, but maybe I'm just old and grumpy but I find those things a bit annoying, I usually already have a tab for the UI open and then I have to go to all the trouble of closing the tab that opens, IT RUINS MY DAY.

Nah I feel the same way, and it's impossible that we're *both* old and grumpy so it must just actually be annoying!

Caleb-Ellis (approved):

huwshimi:
> > Yeah I think thats from wait-on polling the url to see if it's ready and the server rejecting it (maybe from the proxy). I don't really think we can do anything about the warnings, but we might be able to filter out them out from being displayed.
> 
> That must be it. Could filter them out as a follow-up but it's probably not worth the effort, plus the (minimal) added risk of filtering out something we actually want to show.

I had a quick look and it's single-spa that's showing the errors, but I can't see any simple way to silence them.

Issue #2491, opened 2021-04-12
[Self-driven snap] Add available upgrade column


huwshimi:
- [ ] If there is a versions.update then display versions.update.version.

PR #2584, opened 2021-04-28
feat(legacy): add available upgrade column

huwshimi:
## Done

- Add the available upgrade column to the controller list.

### MAAS deployment

- [Bolla](/HACKING.md#bolla)

### Running the branch

You can run this branch by:

### QA steps

- Point your UI to Bolla and go to the controllers list.
- You should see an "Available versions" column and an upgrade should show up (if not let me know and I might need to update the backend again).

## Fixes

Fixes: #2491.

### DIFF
--- a/legacy/src/app/controllers/tests/test_nodes_list.js
+++ b/legacy/src/app/controllers/tests/test_nodes_list.js
@@ -1391,4 +1391,42 @@ describe("NodesListController", function () {
       });
     });
   });
+
+  describe("getUpgradeVersion", () => {
+    it("can get a controller upgrade version", () => {
+      makeController();
+      const controller = {
+        versions: {
+          update: {
+            version: "1.2.3",
+          },
+        },
+      };
+      expect($scope.getUpgradeVersion(controller)).toBe("1.2.3");
+    });
+
+    it("handles a controller without versions", () => {
+      makeController();
+      const controller = {};
+      expect($scope.getUpgradeVersion(controller)).toBe(null);
+    });
+
+    it("handles a controller without an update object", () => {
+      makeController();
+      const controller = {
+        versions: {},
+      };
+      expect($scope.getUpgradeVersion(controller)).toBe(null);
+    });
+
+    it("handles a controller without an update version", () => {
+      makeController();
+      const controller = {
+        versions: {
+          update: {},
+        },
+      };
+      expect($scope.getUpgradeVersion(controller)).toBe(null);
+    });
+  });
 });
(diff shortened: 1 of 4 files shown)


legacy/src/app/partials/nodes-list.html
```diff
@@ -1357,6 +1358,14 @@ <h4 class="page-header__dropdown-title">Add rack controller</h4>
                 </span>
               </div>
             </td>
+            <td class="p-table__cell" aria-label="Available upgrade">
+              <div
+                data-ng-show="controller.versions.update"
+                title="{$ controller.versions.update.version $}"
+              >
+                {$ controller.versions.update.version $}
```
Caleb-Ellis, line 1366: Do these need to be given the `&&` treatment, or do angular templates fail "gracefully"? Either way I think we should play it safe with angular and have a simple function that checks that an upgrade is available.

legacy/src/app/partials/nodes-list.html
```diff
@@ -1357,6 +1358,14 @@ <h4 class="page-header__dropdown-title">Add rack controller</h4>
                 </span>
               </div>
             </td>
+            <td class="p-table__cell" aria-label="Available upgrade">
+              <div
+                data-ng-show="controller.versions.update"
```
Caleb-Ellis, line 1363: `data-ng-if` is almost universally the better option. `data-ng-show` still renders the full markup with an added `ng-hide` class that gives it `display: none`.

Caleb-Ellis (changes requested):

legacy/src/app/partials/nodes-list.html
huwshimi: Thanks, I'll fix that up.

legacy/src/app/partials/nodes-list.html
huwshimi: Good idea.

huwshimi: Done.

legacy/src/app/partials/nodes-list.html
huwshimi: Done.

legacy/src/app/controllers/tests/test_nodes_list.js
```diff
@@ -1391,4 +1391,44 @@ describe("NodesListController", function () {
...
+          },
+        },
+      };
+      expect($scope.getUpgradeVersion(controller)).toBe("1.2.3");
+    });
+
+    it("handles a controller without versions", () => {
+      makeController();
+      const controller = {
+        versions: {},
```
Caleb-Ellis, line 1411: I think is is supposed to be removed

Caleb-Ellis (approved):
LGTM! Just one comment about a test

legacy/src/app/controllers/tests/test_nodes_list.js
huwshimi: Yep! Thanks.

Issue #2589, opened 2021-04-29
VLAN count


huwshimi:
Add number of VLANs to the VLAN column.

PR #2613, opened 2021-05-06
feat(legacy): display controller VLAN counts

huwshimi:
## Done

- Display the number of VLANs for each controller in the controller list.

### MAAS deployment

- [Bolla](/HACKING.md#bolla)

### Running the branch

You can run this branch by:

### QA steps

- Go to the list of controllers.
- In the VLAN column there should be a total VLAN count for each controller.
- Click on the count and it should take you to the VLANs tab in the controller details page.

## Fixes

Fixes: #2589.

### DIFF
--- a/legacy/src/app/controllers/nodes_list.js
+++ b/legacy/src/app/controllers/nodes_list.js
@@ -917,6 +917,12 @@ function NodesListController(
     );
   };
 
+  // Get the number of VLANs for a controller.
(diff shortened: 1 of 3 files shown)


Caleb-Ellis (approved):
LGTM!

Side note - should we be filtering the VLAN list to only those that use the controller as a primary/secondary rack? Seems odd for the VLAN count not to correspond to the number of rows in the VLAN list.

Issue #2568, opened 2021-04-23
Refactor KVM popover components + resource calculations


Caleb-Ellis:
We calculate KVM resources (cores, memory, storage) in many different places. We should unify how we're doing those calculations.

Caleb-Ellis:
I don't think this is valid (at least not anymore). There are a lot of similarities with how the KVM resource calculations are done but they're actually all different. We could refactor some commonalities but I think that already exists with the `resourceWithOverCommit()` util and anything extra might be a bit over-abstracted.

PR #2664, opened 2021-05-18
feat(ui): replace legacy pod websocket fields with 'resources' equivalents

### LINKED ISSUE #2652: Legacy fields in pod.get/list websocket can be dropped
Bug originally filed by ack at https://bugs.launchpad.net/bugs/1928442

Since we now provide a "resources" entry for pod.get/list websocket calls with details about resources usage, there are a few legacy fields that we should be able to drop, once the UI doesn't use them anymore.

Specifically:

- "total"/"used"/"available" (same info should be already available from "resources")
- "composed_machines_count" (replace with "resources.vm_count")
- "owners_count" (already unused?)
- "hints" (already unused?)

Caleb-Ellis:
## Done

- Removed all instances of `pod.composed_machines_count`, `pod.available`, `pod.used` and `pod.total` and replaced with equivalents found in the `pod.resources` field.

### MAAS deployment

- [Bolla](/HACKING.md#bolla)

### Running the branch

You can run this branch by:

### QA steps

- Go to the KVM list and check that the VM and storage cells are unchanged
- Check that you can still sort KVMs by VM count and storage
- Go to a KVM details page and check that the VM count in the header is unchanged

## Fixes

Fixes #2652

### DIFF
--- a/ui/src/app/kvm/views/KVMList/StorageColumn/StorageColumn.tsx
+++ b/ui/src/app/kvm/views/KVMList/StorageColumn/StorageColumn.tsx
@@ -18,9 +18,12 @@ const StorageColumn = ({ id }: Props): JSX.Element | null => {
   );
 
   if (pod) {
-    const availableStorage = formatBytes(pod.total.local_storage, "B");
-    const allocatedStorage = formatBytes(pod.used.local_storage, "B", {
-      convertTo: availableStorage.unit,
+    const { allocated_other, allocated_tracked, free } = pod.resources.storage;
+    const totalInBytes = allocated_other + allocated_tracked + free;
+    const totalStorage = formatBytes(totalInBytes, "B", { decimals: 1 });
+    const allocatedStorage = formatBytes(allocated_tracked, "B", {
+      convertTo: totalStorage.unit,
+      decimals: 1,
     });
 
     return (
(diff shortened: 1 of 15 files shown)


Caleb-Ellis:
We'll wait on 3.0 to land this

ui/src/app/kvm/views/KVMList/StorageColumn/StorageColumn.tsx
```diff
@@ -18,9 +18,12 @@ const StorageColumn = ({ id }: Props): JSX.Element | null => {
   );
 
   if (pod) {
-    const availableStorage = formatBytes(pod.total.local_storage, "B");
-    const allocatedStorage = formatBytes(pod.used.local_storage, "B", {
-      convertTo: availableStorage.unit,
+    const { allocated_other, allocated_tracked, free } = pod.resources.storage;
+    const totalInBytes = allocated_other + allocated_tracked + free;
+    const totalStorage = formatBytes(totalInBytes, "B", { decimals: 1 });
+    const allocatedStorage = formatBytes(allocated_tracked, "B", {
```
huwshimi, line 24: It seems like we're doing nearly the same in StorageResources, is this another case where we could use a util or hook or is it not quite similar enough?

ui/src/app/store/pod/types.ts
```diff
@@ -8,6 +8,7 @@ export enum PodType {
   VIRSH = "virsh",
 }
 
+// Remove when https://bugs.launchpad.net/maas/+bug/1928442 lands
```
huwshimi, line 11: Should we just remove all this now? We don't want to use these types.

huwshimi (approved):
A couple of small comments.

ui/src/app/kvm/views/KVMList/StorageColumn/StorageColumn.tsx
Caleb-Ellis: Yeah we could definitely use a hook or util, but I think I'll handle that in https://github.com/canonical-web-and-design/maas-ui/issues/2568 rather than here.

ui/src/app/store/pod/types.ts
Caleb-Ellis: I guess it feels a little weird to be ahead of the api, but yeah we're not gonna use these values again so they're not useful to keep.

PR #2671, opened 2021-05-18
ci(ci): run actions against the 2.9 branch

huwshimi:
## Done

- Run CI against the 2.9 branch.

### DIFF
--- a/.github/workflows/cypress.yml
+++ b/.github/workflows/cypress.yml
@@ -1,5 +1,8 @@
 name: Cypress
-on: [push]
+on:
+  push:
+    branches:
+      - 2.9
 jobs:
   cypress:
     name: Cypress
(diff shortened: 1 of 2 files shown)


anthonydillon (commented):
I wonder if we should use wildcarded tags? https://docs.github.com/en/actions/reference/workflow-syntax-for-github-actions#onpushpull_requestbranchestags

huwshimi:
I think I need to update the other workflows as well.

huwshimi:
> I wonder if we should use wildcarded tags? https://docs.github.com/en/actions/reference/workflow-syntax-for-github-actions#onpushpull_requestbranchestags

I think we want it set to a specific branch as we don't want to run the actions defined in one branch to be run against a different branch.

huwshimi:
This one is ready for review now.

Caleb-Ellis (approved):

Issue #1949, opened 2020-12-01
Create a util that determines if a machine is of type MachineDetails


Caleb-Ellis:
e.g.
``` typescript
export const isMachineDetails(machine: Machine | MachineDetails): machine is MachineDetails => {
  // something that 100% determines something is MachineDetails
};
```

We could use this instead of `if (machine && "property" in machine)` which is a bit less clear

PR #2681, opened 2021-05-20
refactor(ui): add a machine details type guard util

huwshimi:
## Done

- Move the type guarding for MachineDetails to a util.

## QA

As this is a type guard I don't think a QA is necessary.

## Fixes

Fixes: #1949.

### DIFF
--- a/ui/src/app/store/machine/utils/common.ts
+++ b/ui/src/app/store/machine/utils/common.ts
@@ -5,6 +5,6 @@ import type { Machine, MachineDetails } from "app/store/machine/types";
  * @returns Whether the machine is MachineDetails.
  */
 export const isMachineDetails = (
-  machine?: Machine | MachineDetails | null
+  machine?: Machine | null
   // Use "metadata" as the canary as it only exists for MachineDetails.
 ): machine is MachineDetails => !!machine && "metadata" in machine;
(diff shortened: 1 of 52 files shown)


ui/src/app/store/machine/utils/common.ts
```diff
@@ -0,0 +1,10 @@
+import type { Machine, MachineDetails } from "app/store/machine/types";
+/**
+ * Wether a machine has a Machine or MachineDetails type.
+ * @param machine - The machine to check
+ * @returns Whether the machine is MachineDetails.
+ */
+export const isMachineDetails = (
+  machine?: Machine | MachineDetails | null
```
Caleb-Ellis, line 8: Shouldn't it be either `Machine | null` or `BaseMachine | MachineDetails | null` since `Machine` is already `BaseMachine | MachineDetails`? Maybe we should rename the types to be a bit clearer what the differences are.

Caleb-Ellis (approved):
Just one question

ui/src/app/store/machine/utils/common.ts
huwshimi: Yeah good point, I think `Machine | null` is probably appropriate. That's my fault for copying and not thinking.

Issue #2377, opened 2021-03-14
Update machine slice to use util to generate params


huwshimi:
Use the `generateParams` function for all status handlers to reduce duplication.

PR #2682, opened 2021-05-20
refactor(ui): use util for optional params

huwshimi:
## Done

- Update the machine slice to use the util for optional params.

### MAAS deployment

N/A

## Fixes

Fixes: #2377.


Caleb-Ellis (approved):
Nice! Thanks for that!

Issue #2049, opened 2021-01-14
TypeScript strict mode errors for reducers using GenerateSlice 


squidsoup:
TypeScript issues strict mode errors for reducers using GenerateSlice that have additional custom reducers, e.g. `scriptresult/slice` and `pod/slice`.

from `scriptresult/slice`:
```
   Type 'SliceCaseReducers<ScriptResultState>' does not satisfy the constraint 'SliceCaseReducers<GenericState<ScriptResult, any>>'.
     Index signatures are incompatible.
       Type 'CaseReducer<ScriptResultState, { payload: any; type: string; }> | CaseReducerWithPrepare<ScriptResultState, PayloadAction<any, string, any, any>>' is not assignable to type 'CaseReducer<GenericState<ScriptResult, any>, { payload: any; type: string; }> | CaseReducerWithPrepare<GenericState<ScriptResult, any>, PayloadAction<...>>'.
         Type 'CaseReducer<ScriptResultState, { payload: any; type: string; }>' is not assignable to type 'CaseReducer<GenericState<ScriptResult, any>, { payload: any; type: string; }> | CaseReducerWithPrepare<GenericState<ScriptResult, any>, PayloadAction<...>>'.
           Type 'CaseReducer<ScriptResultState, { payload: any; type: string; }>' is not assignable to type 'CaseReducer<GenericState<ScriptResult, any>, { payload: any; type: string; }>'.
             Types of parameters 'state' and 'state' are incompatible.
               Property 'history' is missing in type 'WritableDraft<GenericState<ScriptResult, any>>' but required in type 'WritableDraft<ScriptResultState>'.
```

**maas-ui version**
tip

**To Reproduce**
1. Turn on typescript strict mode (edit .tsconfig)
2. Open a slice module in your IDE

PR #2694, opened 2021-05-27
refactor(ui): update slice generation to only generate reducers

huwshimi:
## Done

- Generate the common reducers instead of generating the whole slice as this provides much better type
support.

## How to review this PR

- The main changes are in [utils/slice.ts](https://github.com/canonical-web-and-design/maas-ui/pull/2694/files#diff-79f2f5b73d03afbd8158f5b7780ab97de74cb23247511bc89f56ab8fb3531445). Instead of generating the entire slice in the util, this just generates the common reducers.
- The next changes are updating the slice files for each model (e.g. [vlan/slice.ts](https://github.com/canonical-web-and-design/maas-ui/pull/2694/files#diff-d04eab3dac53d9dab4c839b45be5c5d664633f07d524bfe6368d9626ccc5ee7d). The slice is now defined in the file and common reducer util is used to added the reducers. This results in less code and better type support.
- The rest of the updates are largely fixing issues that weren't visible before.
- The status reducers generated by `generateStatusHandlers` still don't have good type support and will need to be revisited at some point.

### MAAS deployment

- [Bolla](/HACKING.md#bolla)

### Running the branch

You can run this branch by:

### QA steps

- Use redux dev tools to check that the data is being loaded into state.
- Try viewing and performing actions on machines and pods as they have the most complicated slices.

## Fixes

Fixes: #2049.


ui/src/app/store/utils/slice.ts
```diff
@@ -227,37 +178,45 @@ export const generateSlice = <
...
       } else {
-        state.items.push(action.payload);
+        // This should just be:
+        // state.items.push(action.payload);
+        // But because of a typscript issue when using some array methods on
+        // generics that extend array unions we instead added the item at the
+        // end of the array. This can be updated when the fix for the following
+        // issue has been released:
+        // https://github.com/microsoft/TypeScript/issues/13995
+        state.items[state.items.length] = action.payload;
```
Caleb-Ellis, line 215: this is pretty whack

ui/src/app/store/utils/slice.ts
```diff
@@ -198,23 +149,23 @@ export const generateSlice = <
...
     },
-    fetchSuccess: (state: GenericState<I, E>, action: PayloadAction<I[]>) => {
+    fetchSuccess: (state: S, action: PayloadAction<S["items"]>) => {
       state.loading = false;
       state.loaded = true;
       state.items = action.payload;
     },
     create: {
-      prepare: (params: TSFixMe) => ({
+      prepare: (params: Partial<S["items"][0]>) => ({
```
Caleb-Ellis, line 168: Is this type right? When creating something we usually only need to supply a subset of the model's fields - maybe even just a name. Plus in the case of creating machines you can actually pass through things that aren't machine fields, like "don't commission automatically". I think we might need a separate type for the create method if we're going to include it in this util

ui/src/app/store/utils/slice.ts
```diff
@@ -227,37 +178,45 @@ export const generateSlice = <
...
+        // generics that extend array unions we instead added the item at the
+        // end of the array. This can be updated when the fix for the following
+        // issue has been released:
+        // https://github.com/microsoft/TypeScript/issues/13995
+        state.items[state.items.length] = action.payload;
       }
     },
     update: {
-      prepare: (params: TSFixMe) => ({
+      prepare: (params: Partial<S["items"][0]>) => ({
```
Caleb-Ellis, line 219: I think update is fine to use this type, but I'm not certain

Caleb-Ellis (approved):
This is great! The diff is intimidating but the "final product" is much easier to make sense of. I just had one concern with typing the `create` function.

ui/src/app/store/utils/slice.ts
huwshimi: I did that mostly as a convenience, I think we have two choices, either have an extra (optional) generic to supply the type of the create and update params or we define these for each model (like I did here: https://github.com/canonical-web-and-design/maas-ui/pull/2694/files#diff-8d6b0d5a1c7fd502004350766887c23e6a2a8cddcaa8a0c4c3d6052f65c4af7eR241).

What do you reckon?

Caleb-Ellis: I think defining it for each model makes the most sense. I feel like adding another generic is the wrong abstraction but I can't really explain why - maybe because this PR already does a lot to remove some!

huwshimi: OK, are you happy if I unlink https://github.com/canonical-web-and-design/maas-ui/issues/2131 and then follow up by adding the create/update types?

Caleb-Ellis: Fine by me!

PR #2713, opened 2021-06-02
refactor(ui): improve status reducers

huwshimi:
## Done

- Replace the initial status reducers with a normal provide functions.

### MAAS deployment

- [Bolla](/HACKING.md#bolla)

### Running the branch

You can run this branch by:

### QA steps

- Perform some actions from the machine list and kvm list and they should work without errors.

## Fixes

Fixes: #2701.

### DIFF
--- a/ui/src/app/store/machine/slice.ts
+++ b/ui/src/app/store/machine/slice.ts
@@ -406,9 +406,9 @@ const machineSlice = createSlice({
         // No state changes need to be handled for this action.
       },
     },
-    abortStart: statusHandlers.abort.start,
-    abortSuccess: statusHandlers.abort.success,
-    abortError: statusHandlers.abort.error,
+    [`${NodeActions.ABORT}Error`]: statusHandlers.abort.error,
+    [`${NodeActions.ABORT}Start`]: statusHandlers.abort.start,
+    [`${NodeActions.ABORT}Success`]: statusHandlers.abort.success,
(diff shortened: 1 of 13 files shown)


ui/src/app/store/machine/slice.ts
```diff
@@ -359,867 +360,1338 @@ const statusHandlers = generateStatusHandlers<
...
+            extra: {},
+            system_id,
+          },
+        },
+      }),
+      reducer: () => {
+        // No state changes need to be handled for this action.
+      },
+    },
+    abortStart: statusHandlers.abort.start,
```
Caleb-Ellis, line 409: Should we be using the `NodeActions` enum for these as well?
``` typescript
[`${NodeActions.ABORT}Start`]: statusHandlers[NodeActions.ABORT].start,
```

Caleb-Ellis (approved):
So this is a bit more verbose, but I think it's a bit easier to reason about what the status reducers actually do and how they're formed.

I wonder if we should be enum-ing all the possible actions for all models, not just the ones that have been defined in `NodeActions`. If we did, it might be easier for the status keys to just be the enum instead of the continuous form (i.e. "abort" instead of "aborting"), and then there's only one place that defines an action name.

ui/src/app/store/machine/slice.ts
huwshimi: Done.

Issue #2504, opened 2021-04-13
Make SelectedAction generic so it can be used across different models


Caleb-Ellis:
The machine forms use the following:
``` typescript
export type SelectedAction = {
  name: MachineAction["name"];
  sentence?: MachineAction["sentence"];
  formProps?: {
    hardwareType?: HardwareType;
    applyConfiguredNetworking?: Scripts["apply_configured_networking"];
  };
};

export type SetSelectedAction = (
  action: SelectedAction | null,
  deselect?: boolean
) => void;
```

The KVM forms use the following:
``` typescript
export enum KVMAction {
  COMPOSE = "compose",
  DELETE = "delete",
  REFRESH = "refresh",
}
export type SelectedAction = KVMAction | null;
export type SetSelectedAction = Dispatch<SetStateAction<SelectedAction>>;
```

We should figure out what's needed for both cases and come up with a generic type.

PR #2728, opened 2021-06-07
refactor(ui): update selected action types to be generic

huwshimi:
## Done

- Make the selected action type generic and update KVM and machine components.

### MAAS deployment

- [Bolla](/HACKING.md#bolla)

### Running the branch

You can run this branch by:

### QA steps

- Check that you can still open the kvm and machine header forms.

## Fixes

Fixes: #2504.


ui/src/app/machines/components/ActionFormWrapper/ActionFormWrapper.tsx
```diff
@@ -45,13 +45,13 @@ const getErrorSentence = (action: SelectedAction, count: number) => {
     case NodeActions.UNLOCK:
       return `${machineString} cannot be unlocked`;
     default:
-      return `${machineString} cannot be ${action.sentence}`;
+      return `${machineString} cannot be ${action.extras?.sentence}`;
```
Caleb-Ellis, line 48: I wonder if we should actually just have a case for each action instead of relying on the api for this, since they're inconsistent with tenses. Followup maybe?

Caleb-Ellis (approved):
This looks great! Separating out `clearSelectedAction` is a good idea. Just one comment on something we could maybe follow up on, but otherwise LGTM!

ui/src/app/machines/components/ActionFormWrapper/ActionFormWrapper.tsx
huwshimi: Thanks, I've added an issue: https://github.com/canonical-web-and-design/maas-ui/issues/2730.

PR #2735, opened 2021-06-07
Add domainList header component

ClementChaumel:
## Done

- Add the header to the DomainsList page
image

## QA

go to http://0.0.0.0:8400/MAAS/r/domains and check that the header is present and displays the number of domains

## Fixes

Fixes: https://github.com/canonical-web-and-design/app-squad/issues/37

### DIFF
--- a/ui/src/app/domains/views/DomainsList/DomainListHeader/DomainListHeader.test.tsx
+++ b/ui/src/app/domains/views/DomainsList/DomainListHeader/DomainListHeader.test.tsx
@@ -1,0 +1,65 @@
+import { mount } from "enzyme";
+import { Provider } from "react-redux";
+import { MemoryRouter } from "react-router-dom";
+import configureStore from "redux-mock-store";
+
+import DomainListHeader from "./DomainListHeader";
+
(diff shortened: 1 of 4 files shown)


ClementChaumel:
Any idea why my header has a transparent background instead of a white one? I can't figure out what I did differently

bartaz:
> Any idea why my header has a transparent background instead of a white one? I can't figure out what I did differently

This seems to happen when using `yarn ui`, all is good with `yarn start`

hatched (approved):
Looks good!

huwshimi:
> > Any idea why my header has a transparent background instead of a white one? I can't figure out what I did differently
> 
> This seems to happen when using `yarn ui`, all is good with `yarn start`

Yeah the CSS is built slightly differently when running `yarn ui`, I think it might include the files in the wrong order somehow. It's much faster to work with `yarn ui` but CSS/visual needs to be checked with `yarn start` as that is how it's built for prod.

Caleb-Ellis (commented):
Looks great!

PR #2737, opened 2021-06-07
Add header component

bethcollins92:
## Done

- Add ZonesListHeader component

## QA

- Run `yarn ui` and go to: http://0.0.0.0:8400/MAAS/r/zones
- Check the header and tooltip are visible 

Screenshot 2021-06-07 at 14 33 21

### DIFF
--- a/ui/src/app/zones/views/ZonesListTitle/ZonesListTitle.tsx
+++ b/ui/src/app/zones/views/ZonesListTitle/ZonesListTitle.tsx
@@ -1,0 +1,21 @@
+import { Fragment } from "react";
+
+import { Tooltip, Icon } from "@canonical/react-components";
+
+const ZonesListTitle = (): JSX.Element => {
+  return (
+    <Fragment>
+      <span className="p-heading--four">Availability zones</span>
+      <span style={{ marginLeft: "1rem" }}>
+        <Tooltip
+          message="A representation of a grouping of nodes, typically by physical
+            location."
+        >
+          <Icon name="help" />
+        </Tooltip>
+      </span>
+    </Fragment>
+  );
+};
+
+export default ZonesListTitle;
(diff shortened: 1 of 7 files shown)


ClementChaumel:
The header should have a white background 
image
(I'm having the same issue, let me know if you figure it out")

ClementChaumel (approved):
LGTM :+1:

ui/src/app/zones/views/ZonesListTitle/ZonesListTitle.tsx
```diff
@@ -0,0 +1,21 @@
+import { Fragment } from "react";
+
+import { Tooltip, Icon } from "@canonical/react-components";
+
+const ZonesListTitle = (): JSX.Element => {
+  return (
+    <Fragment>
```
hatched, line 7: You can avoid `<Fragment>` and the associated import above by using `<>` and `</>` instead.

ui/src/app/zones/views/ZonesListTitle/ZonesListTitle.tsx
```diff
@@ -0,0 +1,21 @@
...
+      <span className="p-heading--four">Availability zones</span>
+      <span style={{ marginLeft: "1rem" }}>
```
Caleb-Ellis, line 9: We try not to use inline styles as much as possible, but it's super common to need to move stuff around a little bit. To that end we've got a [set of utils](https://github.com/canonical-web-and-design/maas-ui/blob/master/ui/src/scss/_utilities.scss#L66) (`u-nudge-right`, `u-nudge-left--small`) that simply add a bit of padding.

bethcollins92: ahh amazing! This is what we were hoping for but didn't think they existed. Thank you!

ui/src/app/zones/views/ZonesListTitle/ZonesListTitle.tsx
bethcollins92: Will do - thanks!

PR #2740, opened 2021-06-08
Rename root dashboard route to index

steverydz:
## Done

Changed the root dashboard route to `index` as it makes more sense

## QA

- Go to 
- Check that the page loads

## Issue
Fixes https://github.com/canonical-web-and-design/app-squad/issues/110


Caleb-Ellis (approved):
haha, nice - thanks for that! I actually deleted that comment because I realised we weren't consistent with the urls that we had already defined, but I'm glad to see the change anyway. :+1:

PR #2743, opened 2021-06-08
Build zones table

bethcollins92:
## Done

- Add sortable table with links to machines with tests
- Add counter to header

## QA

- run `yarn start` and go to http://0.0.0.0:8400/MAAS/r/zones
- Check the table is visible and the links are correct

Screenshot 2021-06-08 at 14 02 45


ui/src/app/zones/views/ZonesListHeader/ZonesListHeader.tsx
```diff
@@ -1,9 +1,18 @@
+import { useSelector } from "react-redux";
+
 import ZonesListTitle from "../ZonesListTitle";
 
 import SectionHeader from "app/base/components/SectionHeader";
+import zoneSelectors from "app/store/zone/selectors";
 
 const ZonesListHeader = (): JSX.Element => {
-  return <SectionHeader title={<ZonesListTitle />}></SectionHeader>;
+  const zones = useSelector(zoneSelectors.count);
```
bartaz, line 9: I guess you could add the fetch here as well?

At least `KVMListHeader` that we looked at has it.

sowasred2012: Yeah we just noticed that, we're gonna add it as part of the form issue, since we're working in the header there.

bartaz (approved):
LGTM

@sowasred2012 there you go 🥫 🍌 🥖 🧀

ui/src/app/zones/views/ZonesList/ZonesList.test.tsx
```diff
@@ -0,0 +1,53 @@
...
+          <ZonesList />
+        </MemoryRouter>
+      </Provider>
+    );
+    const expectedActions = ["zone/fetch"];
+    const actualActions = store.getActions();
+    expect(
+      expectedActions.every((expectedAction) =>
+        actualActions.some((action) => action.type === expectedAction)
+      )
```
huwshimi, line 32: Because you're only testing one action you could have done: `store.getActions().find((action) => action.type === "zone/fetch")`

ui/src/app/zones/views/ZonesListHeader/ZonesListHeader.tsx
```diff
@@ -1,9 +1,18 @@
...
+  return (
+    <SectionHeader
+      title={<ZonesListTitle />}
+      subtitle={`${zones} AZs available`}
+    ></SectionHeader>
```
huwshimi, line 14: This could have just closed the SectionHeader tag: `<SectionHeader ... />`.

ui/src/app/zones/views/ZonesListHeader/ZonesListHeader.tsx
huwshimi: Either is fine, but it's handy to fetch data in the component that needs it in case you end up using the component elsewhere that might not have fetched the data.

ui/src/app/zones/views/ZonesListTable/ZonesListTable.tsx
```diff
@@ -0,0 +1,81 @@
...
+    { content: "Machines", sortKey: "machines", className: "u-align--right" },
+    { content: "Devices", sortKey: "devices", className: "u-align--right" },
+    {
+      content: "Controllers",
+      sortKey: "controllers",
+      className: "u-align--right",
+    },
+  ];
+  const rows = zones.map((zone) => {
+    return {
```
huwshimi, line 31: A small convenience with arrow functions is that you can return directly by doing `const rows = zones.map((zone) => ({...`.

Note that you need `()` around the object otherwise it assumes the curlies are wrapping the function content.

ui/src/app/zones/views/ZonesList/ZonesList.tsx
```diff
@@ -1,13 +1,29 @@
...
+    dispatch(actions.fetch());
+  }, [dispatch]);
+
   return (
     <Section
       header={<ZonesListHeader />}
       headerClassName="u-no-padding--bottom"
-    />
+    >
+      {zonesCount > 0 && <ZonesListTable />}
```
huwshimi, line 25: Here you might want to show a spinner while the data is loading. You could get the loading state using `const loading = useSelector(zoneSelectors.loading);` and then you could display the Spinner component from react-components.

You might also want to add a message if the zones have loaded but there are no zones returned.

PR #2749, opened 2021-06-09
refactor(ui): update Formik secondarySubmit to not also run onSubmit

### LINKED ISSUE #2742: Update FormikForm secondarySubmit to be more flexible
Right now `secondarySubmit` always calls `onSubmit` as well. We should stop doing that.

Caleb-Ellis:
## Done

- Updated `FormikForm`'s `secondarySubmit` prop to not always run `onSubmit`. It also includes formik context so you can use all of Formik's functions in `secondarySubmit` (including `submitForm`).
- Updated the `onCancel` prop to follow the same format, so now all the button functions have the shape `(values, formikContext) => {}`
- Driveby - added `className` prop to `FormikFormContent`

### MAAS deployment

- [Bolla](/HACKING.md#bolla)

### Running the branch

You can run this branch by:

### QA steps

- Go to /MAAS/r/machines/add and check that the "Save and add another" button still works as before

## Fixes

Fixes #2742

### DIFF
--- a/ui/src/app/base/components/FormikFormButtons/FormikFormButtons.tsx
+++ b/ui/src/app/base/components/FormikFormButtons/FormikFormButtons.tsx
@@ -11,6 +11,11 @@ import classNames from "classnames";
 import type { FormikContextType } from "formik";
 import { useFormikContext } from "formik";
 
+export type FormikContextFunc<V> = (
+  values: V,
+  formikContext: FormikContextType<V>
+) => void;
+
 export type Props<V> = {
   buttonsAlign?: "left" | "right";
   buttonsBordered?: boolean;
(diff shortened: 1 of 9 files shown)


ui/src/app/base/components/FormikFormButtons/FormikFormButtons.tsx
```diff
@@ -115,7 +125,9 @@ export const FormikFormButtons = <V,>({
               })}
               data-test="cancel-action"
               disabled={cancelDisabled}
-              onClick={onCancel ? () => onCancel(formikContext) : undefined}
+              onClick={
+                onCancel ? () => onCancel(values, formikContext) : undefined
```
huwshimi, line 129: I can't remember if there was a reason why these were set to `undefined`, but does it work when they're `null` instead?

huwshimi (approved):
Looks good, one question.

ui/src/app/base/components/FormikFormButtons/FormikFormButtons.tsx
Caleb-Ellis: Ahh looks like it's because the `Button` component doesn't accept `null`. I'll file an issue there

PR #2752, opened 2021-06-09
Add discovery list table

steverydz:
## Done
Added basic discoveries table

## QA
- Go to http://0.0.0.0:8400/MAAS/r/dashboard
- Check that there is a table that matches the legacy one http://bolla.internal:5240/MAAS/l/dashboard
- **NOTE: There are no actions**

## Issue
Fixes https://github.com/canonical-web-and-design/app-squad/issues/66

### DIFF
--- a/ui/src/app/dashboard/views/DiscoveriesList/DiscoveriesList.tsx
+++ b/ui/src/app/dashboard/views/DiscoveriesList/DiscoveriesList.tsx
@@ -1,16 +1,75 @@
+import { useEffect } from "react";
+
+import { useSelector, useDispatch } from "react-redux";
+
 import DiscoveriesListHeader from "../DiscoveriesListHeader";
 
 import Section from "app/base/components/Section";
 import { useWindowTitle } from "app/base/hooks";
+import { actions } from "app/store/discovery";
+import discoverySelectors from "app/store/discovery/selectors";
 
 const DiscoveriesList = (): JSX.Element => {
   useWindowTitle("Dashboard");
+
+  const discoveries = useSelector(discoverySelectors.all);
+  const dispatch = useDispatch();
+
+  useEffect(() => {
+    dispatch(actions.fetch());
+  }, [dispatch]);
 
   return (
     <Section
       header={<DiscoveriesListHeader />}
       headerClassName="u-no-padding--bottom"
-    />
+    >
+      <table>
+        <thead>
+          <tr>
+            <th style={{ width: "20%" }}>Name</th>
+            <th style={{ width: "15%" }}>Mac address</th>
+            <th style={{ width: "15%" }}>IP</th>
+            <th style={{ width: "15%" }}>Rack</th>
+            <th>Last seen</th>
+            <th className="u-align--right" style={{ width: "75px" }}>
+              Action
+            </th>
+          </tr>
+        </thead>
+        <tbody>
+          {discoveries
+            ? discoveries.map((discovery) => (
+                <tr key={`discovery-${discovery.id}`}>
+                  <td style={{ overflow: "visible" }}>
+                    {discovery.hostname || "Unknown"}
+                    {discovery.is_external_dhcp ? (
+                      <div className="p-tooltip--top-center">
+                        <i className="p-icon--information ng-hide"></i>
+                        <span className="p-tooltip__message" role="tooltip">
+                          This device is providing DHCP
+                        </span>
+                      </div>
+                    ) : null}
+                  </td>
+                  <td>
+                    <div className="u-truncate">
+                      {discovery.mac_address || "Unknown"}
+                    </div>
+                    <div className="u-truncate">
+                      <small>{discovery.mac_organization}</small>
+                    </div>
+                  </td>
+                  <td className="u-truncate">{discovery.ip}</td>
+                  <td className="u-truncate">{discovery.observer_hostname}</td>
+                  <td className="u-truncate">{discovery.last_seen}</td>
+                  <td className="u-truncate">&nbsp;</td>
+                </tr>
+              ))
+            : null}
+        </tbody>
+      </table>
+    </Section>
   );
 };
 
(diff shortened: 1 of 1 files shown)


ui/src/app/dashboard/views/DiscoveriesList/DiscoveriesList.tsx
```diff
@@ -1,16 +1,75 @@
...
+    dispatch(actions.fetch());
+  }, [dispatch]);
+
   return (
     <Section
       header={<DiscoveriesListHeader />}
       headerClassName="u-no-padding--bottom"
-    />
+    >
+      <table>
```
hatched, line 27: You should probably stick with using the `MainTable` component that is used elsewhere in the application.

hatched (dismissed):

ui/src/app/dashboard/views/DiscoveriesList/DiscoveriesList.tsx
Caleb-Ellis: Yeah, give `MainTable` a go. What @bartaz did for the [domains list table](https://github.com/canonical-web-and-design/maas-ui/blob/master/ui/src/app/domains/views/DomainsList/DomainsTable/DomainsTable.tsx) is a good place to start.

When the table starts to get a bit more interactive (sorting, actions, etc), you can use the [machine tests table](https://github.com/canonical-web-and-design/maas-ui/blob/master/ui/src/app/machines/views/MachineDetails/MachineTests/MachineTestsTable/MachineTestsTable.tsx) or the [DHCP table](https://github.com/canonical-web-and-design/maas-ui/blob/master/ui/src/app/machines/views/MachineDetails/MachineNetwork/DHCPTable/DHCPTable.tsx) as guides. There's also things like the [available storage table](https://github.com/canonical-web-and-design/maas-ui/blob/master/ui/src/app/machines/views/MachineDetails/MachineStorage/AvailableStorageTable/AvailableStorageTable.tsx), but I don't think the discoveries list will get to that level of complexity.

ui/src/app/dashboard/views/DiscoveriesList/DiscoveriesList.tsx
```diff
@@ -1,16 +1,75 @@
...
+        <thead>
+          <tr>
+            <th style={{ width: "20%" }}>Name</th>
```
Caleb-Ellis, line 30: We typically create a `_index.scss` file for each table, and assign each `th` and `td` in that column a class (e.g. `name-col`). This way we can write media queries for dropping columns at certain breakpoints.

You can see [MachineTestsTable](https://github.com/canonical-web-and-design/maas-ui/blob/master/ui/src/app/machines/views/MachineDetails/MachineTests/MachineTestsTable/index.scss) does this. For now I wouldn't worry about using the `breakpoint-widths` mixin, just assigning static values will be enough and we can worry about the "responsiveness" later.

ui/src/app/dashboard/views/DiscoveriesList/DiscoveriesList.tsx
```diff
@@ -1,16 +1,75 @@
...
+            <th style={{ width: "15%" }}>IP</th>
+            <th style={{ width: "15%" }}>Rack</th>
+            <th>Last seen</th>
+            <th className="u-align--right" style={{ width: "75px" }}>
+              Action
+            </th>
+          </tr>
+        </thead>
+        <tbody>
+          {discoveries
```
Caleb-Ellis, line 41: The `discoverySelectors.all` selector should always return an array so I think this ternary might be unnecessary.

ui/src/app/dashboard/views/DiscoveriesList/DiscoveriesList.tsx
```diff
@@ -1,16 +1,75 @@
...
+            ? discoveries.map((discovery) => (
+                <tr key={`discovery-${discovery.id}`}>
+                  <td style={{ overflow: "visible" }}>
+                    {discovery.hostname || "Unknown"}
+                    {discovery.is_external_dhcp ? (
+                      <div className="p-tooltip--top-center">
```
Caleb-Ellis, line 47: We have [Tooltip component](https://github.com/canonical-web-and-design/react-components/blob/master/src/components/Tooltip/Tooltip.tsx) in react-components that you could use here.

ui/src/app/dashboard/views/DiscoveriesList/DiscoveriesList.tsx
```diff
@@ -1,16 +1,133 @@
...
+          content: (
+            <>
+              <div className="u-truncate">
+                {discovery.mac_address || "Unknown"}
+              </div>
+              <div className="u-truncate">
+                <small>{discovery.mac_organization}</small>
+              </div>
+            </>
+          ),
```
Caleb-Ellis, line 86: We actually have a component `DoubleRow` which would be appropriate here.
``` typescript
<DoubleRow
  primary={discovery.mac_address || "Unknown"}
  secondary={discovery.mac_organization}
/>
```

ui/src/app/dashboard/views/DiscoveriesList/DiscoveriesList.tsx
```diff
@@ -1,16 +1,133 @@
...
+      className: "p-table--network-discoveries__ip",
+    },
+    {
+      content: "Rack",
+      sortKey: "rack",
+      className: "p-table--network-discoveries__rack",
+    },
+    {
+      content: "Last seen",
+      sortKey: "last_seen",
```
Caleb-Ellis, line 46: The default sorting handled by `MainTable` is very basic so it's not really working as intended here. The api returns a string, so it's sorting by string value meaning all the Mondays are getting grouped together, all the Tuesdays etc. instead of actually being sorted by time. The angular version also has this problem.

For now I would probably remove the ability to sort by `last_seen` (so just remove the `sortKey` and `sortData` for this column) and we can fix this as a follow up, either by parsing the string using `date-fns` or requesting that the api include a unix timestamp in the discovery model.

Caleb-Ellis (approved):
LGTM! Just some very minor comments

Issue #2756, opened 2021-06-10
Update argPath arguments to use object keys


Caleb-Ellis:
`domainsURLs.details(null, true)` doesn't tell me what's happening here. I have to find the code where `details` is defined and inspect the function (and the resulting rabbit hole of the undocumented arguments of `argPath` ;) ). Instead of using positional arguments for boolean values it might be better for clarity to use an object with keys.

```js
domainsURLs.details({
  args: null,
  unmodified: true
})
```

_Originally posted by @hatched in https://github.com/canonical-web-and-design/maas-ui/pull/2754#discussion_r648540134_

huwshimi:
This could also add a docstring for the function it generates, this way your IDE will show descriptions for the function args.

hatched:
Right but a lot of time code isn't read in an IDE....like in a PR for example :D

But yeah even having proper docstrings would have helped here.

PR #2757, opened 2021-06-10
docs(ui): document the url function

huwshimi:
## Done

- Add docs to the generated URL function so that it's easier to use.

### QA steps

- Open `ui/src/app/Routes.tsx`.
- Hover a URL function e.g. `machineURLs.machine.index` on line 71. You should see a tooltip with the function args.

## Fixes

Fixes: #2756.

### DIFF
--- a/ui/src/app/utils/argPath.ts
+++ b/ui/src/app/utils/argPath.ts
@@ -5,10 +5,20 @@ type Args = {
 /**
  * Creates a function to transform a react-router style URL into a full path.
  * @param path - A path in react-router format e.g. /machine/:id
- * @returns A function to generate the full url.
+ * @template A - The type of the url args object.
+ * @returns A function to generate the full url or a url string for react-router.
  */
 export const argPath =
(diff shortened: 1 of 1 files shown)


ui/src/app/utils/argPath.ts
```diff
@@ -5,10 +5,20 @@ type Args = {
...
   <A extends Args>(path: string) =>
+  /**
+   *
+   * @param args An object of URL parameters. The object keys match the URL
+   * params, e.g. for `/machine/:id` you would pass `{id: 1}`. This should be
+   * `null` when you wish to return the unmodified react-router URL.
+   * @param unmodified Whether to return the URL in the react-router format
+   * .e.g. `/machine/:id`.
+   * @returns The complete or unmodified URL.
+   */
```
huwshimi, line 21: I didn't know you could nest docstrings like this, but it works!

hatched: Very cool

hatched (approved):

PR #2758, opened 2021-06-10
fix(ui): sprint cleanup Jun 10

Caleb-Ellis:
## Done

- Fixed up some minor issues in the domain and zone components

### MAAS deployment

- [Bolla](/HACKING.md#bolla)

### Running the branch

You can run this branch by:

### QA steps

- Go to /MAAS/r/zones
- Check that you can add a zone with a description
- Check that trying to add another zone with the same name results in error
- Check that closing and opening the form again removes the error
- Go to /MAAS/r/domains
- Check that you can add a domain with the normal "Save" button and the form closes
- Check that you can add a domain with the "Save and add another" button and the form does not close, but it resets the values
- Check that trying to add another domain with the same name results in an error, using either button
- Check that closing and opening the form again removes the error

### DIFF
--- a/ui/src/app/base/components/FormikFormContent/FormikFormContent.test.tsx
+++ b/ui/src/app/base/components/FormikFormContent/FormikFormContent.test.tsx
@@ -15,6 +15,7 @@ import {
   configState as configStateFactory,
   rootState as rootStateFactory,
 } from "testing/factories";
+import { waitForComponentToPaint } from "testing/utils";
 
 const mockStore = configureStore();
 
@@ -267,5 +268,57 @@ describe("FormikFormContent", () => {
     });
     wrapper.update();
     expect(wrapper.find("input[name='val1']").props().value).toBe("initial");
+  });
+
+  it("runs onSuccess function if successfully saved with no errors", async () => {
+    const onSuccess = jest.fn();
+    const store = mockStore(state);
+
+    const Proxy = ({ saved }: { saved: boolean }) => (
+      <Provider store={store}>
+        <MemoryRouter initialEntries={[{ pathname: "/", key: "testKey" }]}>
+          <Formik initialValues={{}} onSubmit={jest.fn()}>
+            <FormikFormContent onSuccess={onSuccess} saved={saved}>
+              <Field name="val1" />
+            </FormikFormContent>
+          </Formik>
+        </MemoryRouter>
+      </Provider>
+    );
+    const wrapper = mount(<Proxy saved={false} />);
+
+    expect(onSuccess).not.toHaveBeenCalled();
+
+    wrapper.setProps({ saved: true });
+    waitForComponentToPaint(wrapper);
+    expect(onSuccess).toHaveBeenCalled();
+  });
+
+  it("does not run onSuccess function if saved but there are errors", async () => {
+    const onSuccess = jest.fn();
+    const store = mockStore(state);
+
+    const Proxy = ({ errors, saved }: { errors?: string; saved: boolean }) => (
+      <Provider store={store}>
+        <MemoryRouter initialEntries={[{ pathname: "/", key: "testKey" }]}>
+          <Formik initialValues={{}} onSubmit={jest.fn()}>
+            <FormikFormContent
+              errors={errors}
+              onSuccess={onSuccess}
+              saved={saved}
+            >
+              <Field name="val1" />
+            </FormikFormContent>
+          </Formik>
+        </MemoryRouter>
+      </Provider>
+    );
+    const wrapper = mount(<Proxy errors={undefined} saved={false} />);
+
+    expect(onSuccess).not.toHaveBeenCalled();
+
+    wrapper.setProps({ errors: "Everything is ruined", saved: true });
+    waitForComponentToPaint(wrapper);
+    expect(onSuccess).not.toHaveBeenCalled();
   });
 });
--- a/ui/src/app/base/validation.ts
+++ b/ui/src/app/base/validation.ts
@@ -1,4 +1,11 @@
 /**
+ * Validate domain name e.g. "www.example.com"
+ * XXX This should be updated as it currently allows e.g. "www.example.com."
+ * https://github.com/canonical-web-and-design/maas-ui/issues/2755
+ */
+export const DOMAIN_NAME_REGEX = /^([a-z\d]|[a-z\d][a-z\d-.]*[a-z\d])*$/i;
+
+/**
  * Validate IPv4 address e.g 192.168.1.1
  */
 export const IPV4_REGEX =
--- a/ui/src/app/domains/views/DomainDetails/DomainsDetails.test.tsx
+++ b/ui/src/app/domains/views/DomainDetails/DomainsDetails.test.tsx
@@ -1,0 +1,56 @@
+import { mount } from "enzyme";
+import { Provider } from "react-redux";
+import { MemoryRouter, Route } from "react-router-dom";
+import configureStore from "redux-mock-store";
+
+import DomainDetails from "./DomainDetails";
+
+import {
+  domain as domainFactory,
+  domainState as domainStateFactory,
+  rootState as rootStateFactory,
+} from "testing/factories";
+
+const mockStore = configureStore();
+
+describe("DomainDetails", () => {
+  it("shows a spinner if domain has not loaded yet", () => {
+    const state = rootStateFactory({
+      domain: domainStateFactory({ items: [] }),
+    });
+    const store = mockStore(state);
+    const wrapper = mount(
+      <Provider store={store}>
+        <MemoryRouter
+          initialEntries={[{ pathname: "/domain/1", key: "testKey" }]}
+        >
+          <Route exact path="/domain/:id" component={() => <DomainDetails />} />
+        </MemoryRouter>
+      </Provider>
+    );
+
+    expect(wrapper.find("SectionHeader Spinner").exists()).toBe(true);
+  });
+
+  it("shows the domain name in the header if domain has loaded", () => {
+    const state = rootStateFactory({
+      domain: domainStateFactory({
+        items: [domainFactory({ id: 1, name: "domain-in-the-membrane" })],
+      }),
+    });
+    const store = mockStore(state);
+    const wrapper = mount(
+      <Provider store={store}>
+        <MemoryRouter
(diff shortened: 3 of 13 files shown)


ui/src/app/base/components/FormikFormContent/FormikFormContent.test.tsx
```diff
@@ -268,4 +269,56 @@ describe("FormikFormContent", () => {
     wrapper.update();
     expect(wrapper.find("input[name='val1']").props().value).toBe("initial");
   });
+
+  it("runs onSuccess function if successfully saved with no errors", async () => {
```
Caleb-Ellis, line 273: @sowasred2012 @bethcollins92 Instead of writing a test in the zone list header, I've written a much simpler one here that just asserts that the `onSuccess` function is called when expected.

The test you had started writing in the zone list header extended a bit beyond the boundaries of the component itself (i.e. mocking redux actions) and is really tricky to get right. I think in general, at least for this sprint, we can just write tests that assert that given some inputs (store/props) and optionally some internal actions (click a button), we can expect some outputs (action is dispatched from the component/form is closed).

ui/src/app/base/components/FormikFormContent/FormikFormContent.test.tsx
```diff
@@ -268,4 +269,56 @@ describe("FormikFormContent", () => {
...
+    const onSuccess = jest.fn();
+    const store = mockStore(state);
+
+    const Proxy = ({ saved }: { saved: boolean }) => (
```
Caleb-Ellis, line 277: This `Proxy` component is used as a workaround to set component props on components connected to the store, since you can only set props on the top-level component (which would otherwise be `Provider`).

ui/src/app/base/validation.ts
```diff
@@ -1,3 +1,10 @@
+/**
+ * Validate domain name e.g. "www.example.com"
+ * XXX This should be updated as it currently allows e.g. "www.example.com."
+ * https://github.com/canonical-web-and-design/maas-ui/issues/2755
+ */
+export const DOMAIN_NAME_REGEX = /^([a-z\d]|[a-z\d][a-z\d-.]*[a-z\d])*$/i;
```
Caleb-Ellis, line 6: @ClementChaumel I've just moved that domain regex here, which houses a few other regexes

ui/src/app/domains/views/DomainDetails/DomainsDetails.test.tsx
```diff
@@ -0,0 +1,56 @@
...
+
+import {
+  domain as domainFactory,
+  domainState as domainStateFactory,
+  rootState as rootStateFactory,
+} from "testing/factories";
+
+const mockStore = configureStore();
+
+describe("DomainDetails", () => {
```
Caleb-Ellis, line 16: @bartaz I've just added some basic tests to the `DomainDetails` component

ui/src/app/domains/views/DomainDetails/DomainsDetails.test.tsx
```diff
@@ -0,0 +1,56 @@
...
+    const state = rootStateFactory({
+      domain: domainStateFactory({ items: [] }),
+    });
+    const store = mockStore(state);
+    const wrapper = mount(
+      <Provider store={store}>
+        <MemoryRouter
+          initialEntries={[{ pathname: "/domain/1", key: "testKey" }]}
+        >
+          <Route exact path="/domain/:id" component={() => <DomainDetails />} />
```
Caleb-Ellis, line 27: @bartaz This is a bit of a weird pattern, but in order to get `useParams` to give you what you want in a test, you have to wrap the *actual* component in this `Route` component.

So in this case we're mocking that the url is at `/domain/1`, but since no there is no domain in state with an id of 1, a spinner should be showing

ui/src/app/domains/views/DomainsList/DomainListHeaderForm/DomainListHeaderForm.test.tsx
```diff
@@ -3,41 +3,39 @@ import { act } from "react-dom/test-utils";
 import { Provider } from "react-redux";
 import configureStore from "redux-mock-store";
 
-import DomainListHeader from "../DomainListHeader/DomainListHeader";
+import DomainListHeaderForm from "./DomainListHeaderForm";
```
Caleb-Ellis, line 6: @ClementChaumel generally we try to only run tests on the component in that directory. There are some older form components in the codebase that do the same as here however - just ignore those. :)

ui/src/app/domains/views/DomainsList/DomainListHeaderForm/DomainListHeaderForm.test.tsx
```diff
@@ -3,41 +3,39 @@ import { act } from "react-dom/test-utils";
...
 
 import type { RootState } from "app/store/root/types";
 import { rootState as rootStateFactory } from "testing/factories";
 
 const mockStore = configureStore();
-let state: RootState;
 
-beforeEach(() => {
```
Caleb-Ellis, line 14: @ClementChaumel Like Jeff said these make more sense in the `describe`

ui/src/app/domains/views/DomainsList/DomainListHeaderForm/DomainListHeaderForm.tsx
```diff
@@ -13,7 +16,7 @@ type Props = {
   closeForm: () => void;
 };
 
-export type CreateDomain = {
+export type CreateDomainValues = {
```
Caleb-Ellis, line 19: @ClementChaumel This is just to be consistent with other forms - forms get `Values` on the end, action parameters get `Params` on the end

ui/src/app/domains/views/DomainsList/DomainListHeaderForm/DomainListHeaderForm.tsx
```diff
@@ -23,52 +26,56 @@ const DomainListHeaderForm = ({ closeForm }: Props): JSX.Element => {
...
+  const cleanup = useCallback(() => domainActions.cleanup(), []);
+  const [shouldClose, setShouldClose] = useState(false);
 
-  // Pattern that matches a domainname.
-  // XXX 2016-02-24 lamont: This also matches "example.com.",
-  // which is wrong.
-  const domainnamePattern = /^([a-z\d]|[a-z\d][a-z\d-.]*[a-z\d])*$/i;
-
-  const domainNameSchema: SchemaOf<CreateDomain> = Yup.object()
+  const CreateDomainSchema: SchemaOf<CreateDomainValues> = Yup.object()
```
Caleb-Ellis, line 32: Same again, just consistency

ui/src/app/domains/views/DomainsList/DomainsTable/DomainsTable.tsx
```diff
@@ -2,8 +2,7 @@ import { MainTable } from "@canonical/react-components";
 import { useSelector } from "react-redux";
 import { Link } from "react-router-dom";
 
-import domainURLs from "../../../urls";
-
+import domainURLs from "app/domains/urls";
```
Caleb-Ellis, line 5: @bartaz If it's more than one level up or down it's probably safer to use absolute paths instead of relative paths

ui/src/app/zones/views/ZonesListForm/ZonesListForm.tsx
```diff
@@ -21,24 +22,24 @@ const ZonesListForm = ({ closeForm }: Props): JSX.Element => {
   const errors = useSelector(zoneSelectors.errors);
   const saved = useSelector(zoneSelectors.saved);
   const saving = useSelector(zoneSelectors.saving);
-
-  useEffect(() => {
-    dispatch(actions.fetch());
-  }, [dispatch]);
+  const cleanup = useCallback(() => zoneActions.cleanup(), []);
```
Caleb-Ellis, line 25: @sowasred2012 @bethcollins92 @ClementChaumel this cleanup function is important to pass to `FormikForm` because it resets the `saving`, `saved` and `errors` state when the form closes. It's wrapped in a `useCallback` so that the function doesn't get re-instantiated every time the component renders. While not absolutely necessary, it's an easy performance booster.

ui/src/app/zones/views/ZonesListTable/ZonesListTable.tsx
```diff
@@ -40,18 +44,26 @@ const ZonesListTable = (): JSX.Element => {
         },
         {
           content: (
-            <Link to={`${machineURLs.machines.index}?zone=%3D${zone.name}`}>
+            <Link to={`${machineURLs.machines.index}${machinesFilter}`}>
```
Caleb-Ellis, line 47: @bethcollins92 @sowasred2012 instead of writing out the query parameters manually, we have this `filtersToQueryString` util that handles all the million ways you can filter machines.

ui/src/app/zones/views/ZonesListTable/ZonesListTable.tsx
```diff
@@ -40,18 +44,26 @@ const ZonesListTable = (): JSX.Element => {
...
               {zone.machines_count}
             </Link>
           ),
           className: "u-align--right",
         },
         {
-          content: zone.devices_count,
+          content: (
+            <LegacyLink route={baseURLs.devices}>
```
Caleb-Ellis, line 55: We also have this `LegacyLink` component specifically for routing to the legacy app. I've noticed that the query params don't actually work in the legacy app, so don't worry about that for now

ui/src/app/zones/views/ZonesListForm/ZonesListForm.tsx
```diff
@@ -51,13 +52,13 @@ const ZonesListForm = ({ closeForm }: Props): JSX.Element => {
...
         <Col size="6">
           <FormikField
-            component={Input}
             label="Name"
             placeholder="Name"
             type="text"
             name="name"
             required
           />
+          <FormikField label="Description" type="text" name="description" />
```
Caleb-Ellis, line 61: I've added a Description form field here - seemed strange that this wasn't always the case. But like we've said before, if the angular app looks like garbage or it feels like something could be improved, don't feel like you have to implement things 1:1

huwshimi (approved):

huwshimi:
@Caleb-Ellis a QA issue when adding domains. If you open the form and enter a name a submit using the add-another button then enter the same name and the submit using the normal button it closes without an error.

PR #2760, opened 2021-06-10
feat(ui): add remaining bootresource actions to slice

Caleb-Ellis:
## Done

- Added remaining bootresource actions to slice

### MAAS deployment

- [Bolla](/HACKING.md#bolla)

### Running the branch

You can run this branch by:

### QA steps

- N/A. Some stuff might be wrong but we'll catch it when we start working with the actions for real.

## Fixes

Fixes canonical-web-and-design/app-squad#76


ui/src/app/store/bootresource/slice.ts
```diff
@@ -13,19 +19,90 @@ const bootResourceSlice = createSlice({
   initialState: {
     connectionError: false,
     eventErrors: [],
+    fetchedImages: null,
     otherImages: [],
     rackImportRunning: false,
     regionImportRunning: false,
     resources: [],
     statuses: {
-      poll: false,
```
huwshimi, line 21: This is totes fine, I just thought we discussed making these status names the same as the action, but maybe that was more musing that deciding.

ui/src/app/store/bootresource/slice.ts
```diff
@@ -66,8 +143,118 @@ const bootResourceSlice = createSlice({
       state.ubuntu = payload.ubuntu;
       state.ubuntuCoreImages = payload.ubuntu_core_images;
     },
-    cleanup(state: BootResourceState) {
```
huwshimi, line 69: Thanks!

ui/src/app/store/bootresource/types/actions.ts
```diff
@@ -0,0 +1,32 @@
...
+export type SaveUbuntuCoreParams = {
+  images: string[];
+};
+
+export type SaveUbuntuParams = {
+  osystems: UbuntuOsystem[];
+  source_type: BootResourceSourceType;
+};
+
+export type UbuntuOsystem = {
```
huwshimi, line 28: Should this type be before `SaveUbuntuParams`? It seems to work, but always seems odd when they're defined afterwards, but then it messes up the alphabet!

huwshimi (approved):
Nice! A couple of comments. I haven't validated the API calls with the MAAS code, because as you say we can improve these as needed when we're using them.

ui/src/app/store/bootresource/slice.ts
Caleb-Ellis: Oh, I thought we decided the opposite! I think this is what some would call a *complete and total breakdown of communication* hahaha. I actually like the idea of using the action name more, but maybe this is something we should update for all the models at once.

Added an issue here: https://github.com/canonical-web-and-design/maas-ui/issues/2767

ui/src/app/store/bootresource/types/actions.ts
Caleb-Ellis: Yeah I'm not sure how to reconcile this. Logically if you were to read top to bottom it makes sense for `SaveUbuntuParams` to be at the top. But if you were looking for something in particular, it's easier to scan things that are alphabetical. But if you were *actually* looking for something, you'd grep it. So I dunno! Maybe it doesn't matter.

PR #2761, opened 2021-06-10
Add zone detail route and basic component

bethcollins92:
## Done

- Add route and basic view for zone details

## QA

- run `yarn start` and go to 
- Check you can see the zone details header

Screenshot from 2021-06-10 09-57-41


huwshimi (approved):
Looks good!

PR #2762, opened 2021-06-10
Add "set default" action to the domain table

ClementChaumel:
## Done
image

1 - added `(default)` to the default domain name
2 - added the set default confirmation in an expandable row
3 - added the contextual menu to expand the row 

## QA

go to the table http://0.0.0.0:8400/MAAS/r/domains and switch around the default domain

## Fixes

Fixes: https://github.com/canonical-web-and-design/app-squad/issues/40

## Note

The background of the expanded row should be white (probably an artifact of yarn UI vs yarn start)
The alignment is a bit wonky there but we tried every "simple" solution and this was the best we could come up with without touching FormikForm component

## To do

- [x] Add tests for the table component


hatched (approved):

huwshimi (approved):
Thanks Clement, there are a few things here that I'll fix to keep you moving.

PR #2764, opened 2021-06-10
Add zone details header component

bethcollins92:
## Done

- Organise zone details and zone list into their own directories
- Add Zone details header component
- Add id to link in zone list table
- Add tests for header

## QA

- run `yarn start` and go to http://0.0.0.0:8400/MAAS/r/zone/3 - check you can see the zone name and button
- go to http://0.0.0.0:8400/MAAS/r/zone/334 check you see not found message

Screenshot 2021-06-10 at 14 50 07

Screenshot 2021-06-10 at 14 50 22


ui/src/app/zones/views/ZoneDetails/ZoneDetailsHeader/ZoneDetailsHeader.tsx
```diff
@@ -0,0 +1,46 @@
...
+  let title = "";
+
+  if (zonesLoaded && zone) {
+    title = `Availability zone: ${zone.name}`;
+  } else if (zonesLoaded) {
+    title = "Availability zone not found";
+    buttons = null;
+  }
+
+  useEffect(() => {
```
hatched, line 37: My personal preference is to keep these one shot use-effects before the component logic as they are arguably the first thing that needs to execute.

ui/src/app/zones/views/ZoneDetails/ZoneDetailsHeader/ZonesDetailsHeader.test.tsx
```diff
@@ -0,0 +1,77 @@
...
+            id: 1,
+            name: "zone-name",
+          }),
+        ],
+      }),
+    });
+  });
+
+  it("displays zone name in header if one exists", () => {
+    const state = { ...initialState };
```
hatched, line 37: I'm not sure this is necessary as you're getting a new `initialState` on every test run and a spread doesn't actually 'clone' any nested objects.

hatched (approved):

ui/src/app/zones/views/ZoneDetails/ZoneDetailsHeader/ZonesDetailsHeader.test.tsx
huwshimi: That's probably our fault as that's done in some old tests from before we hand the factories.

PR #2768, opened 2021-06-11
feat(ui): add new table confirm component

huwshimi:
## Done

- Add a new table confirm component. 
- Update the domain table to use the new component. 
- Domain improvements from review.

### MAAS deployment

- [Bolla](/HACKING.md#bolla)

### Running the branch

You can run this branch by:

### QA steps

- Go to /r/domains.
- Check that the domain confirmation has a white background.
- Check that you can change the default domain.

Screen Shot 2021-06-11 at 4 30 19 pm

### DIFF
--- a/ui/src/app/base/components/TableConfirm/TableConfirm.tsx
+++ b/ui/src/app/base/components/TableConfirm/TableConfirm.tsx
@@ -1,0 +1,88 @@
+import type { ReactNode } from "react";
+
+import {
+  ActionButton,
+  Button,
+  Col,
+  Notification,
+  Row,
+} from "@canonical/react-components";
+import type { Props as ButtonProps } from "@canonical/react-components/dist/components/Button";
+
+import { COL_SIZES } from "app/base/constants";
+import { useCycled } from "app/base/hooks";
+
+export type Props = {
+  confirmAppearance?: ButtonProps["appearance"];
+  confirmLabel: string;
+  errors?: string | Record<string, string[]>;
+  errorKey?: string;
+  finished: boolean;
+  inProgress: boolean;
+  message: ReactNode;
+  onClose: () => void;
+  onConfirm: () => void;
+  sidebar?: boolean;
+};
+
+const TableConfirm = ({
+  confirmAppearance = "positive",
+  confirmLabel,
+  errors,
+  errorKey,
+  finished,
+  inProgress,
+  message,
+  onClose,
+  onConfirm,
+  sidebar = true,
+}: Props): JSX.Element => {
+  useCycled(finished, () => {
+    onClose();
+  });
+  let errorMessage: string | null = null;
+  if (errors) {
+    if (typeof errors === "string") {
+      errorMessage = errors;
+    } else if (errorKey && typeof errors === "object") {
+      errorMessage = errors[errorKey].join(" ");
+    }
+  }
+
+  const { TABLE_CONFIRM_BUTTONS, SIDEBAR, TOTAL } = COL_SIZES;
+  return (
+    <Row>
+      {errorMessage && (
+        <Notification type="negative" status="Error:">
+          {errorMessage}
+        </Notification>
+      )}
+      <Col
+        size={
+          sidebar
+            ? TOTAL - SIDEBAR - TABLE_CONFIRM_BUTTONS
+            : TOTAL - TABLE_CONFIRM_BUTTONS
+        }
+      >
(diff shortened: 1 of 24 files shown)


ui/src/app/pools/views/Pools.js
```diff
@@ -69,7 +69,7 @@ const generateRows = (
                   "Cannot delete a pool that contains machines.")
               }
               editDisabled={!row.permissions.includes("edit")}
-              editPath={poolsURLs.edit({ edit: row.id })}
+              editPath={poolsURLs.edit({ id: row.id })}
```
huwshimi, line 72: This is just a driveby fix for going to the correct url.

ui/src/app/base/components/TableConfirm/TableConfirm.tsx
```diff
@@ -0,0 +1,88 @@
...
+  errorKey?: string;
+  finished: boolean;
+  inProgress: boolean;
+  message: ReactNode;
+  onClose: () => void;
+  onConfirm: () => void;
+  sidebar?: boolean;
+};
+
+const TableConfirm = ({
```
huwshimi, line 28: @ClementChaumel here's the new confirm component I added that fits in with the style of our delete confirmations and I also updated it to handle API errors.

ui/src/app/domains/views/DomainsList/DomainsTable/DomainsTable.tsx
```diff
@@ -113,6 +129,7 @@ const DomainsTable = (): JSX.Element => {
 
   return (
     <MainTable
+      className="p-table-expanding--light"
```
huwshimi, line 132: @ClementChaumel I figured out that this class was required for the confirmation to have the white background.

ui/src/app/store/domain/slice.ts
```diff
@@ -38,10 +43,19 @@ const domainSlice = createSlice({
...
+    setDefaultError: (
+      state: DomainState,
+      action: PayloadAction<SetDefaultErrors>
+    ) => {
       state.saving = false;
       // API seems to return the domain id in payload.error not an error message
-      state.errors = "There was an error when setting default domain.";
+      // when the domain can't be found. This override can be removed when the
+      // bug is fixed: https://bugs.launchpad.net/maas/+bug/1931654.
+      if (typeof action.payload === "number") {
```
huwshimi, line 54: @ClementChaumel here I'm only overriding API errors when the error is a number so that the other errors are passed through.

ui/src/app/domains/views/DomainsList/DomainsTable/DomainsTable.tsx
ClementChaumel: Nice one!

ui/src/app/base/components/TableConfirm/TableConfirm.tsx
```diff
@@ -0,0 +1,88 @@
...
+} from "@canonical/react-components";
+import type { Props as ButtonProps } from "@canonical/react-components/dist/components/Button";
+
+import { COL_SIZES } from "app/base/constants";
+import { useCycled } from "app/base/hooks";
+
+export type Props = {
+  confirmAppearance?: ButtonProps["appearance"];
+  confirmLabel: string;
+  errors?: string | Record<string, string[]>;
```
Caleb-Ellis, line 18: Should we be using the `FormErrors` type here? Maybe it needs to go in base types

ui/src/app/pools/views/Pools.js
Caleb-Ellis: whoops, how did that happen? :grimacing:

Caleb-Ellis (approved):
Nice, I think this works well.

I think later down the line we could consider using `FormikForm` as the base for all of our forms, and something like this new `TableConfirm` component would wrap around it, setting some reasonable prop defaults and maybe adding an extra layer of functionality. But at least at the core things like error handling and validation and analytics are all done the same way.

ui/src/app/pools/views/Pools.js
huwshimi: This is what happens when it's not a TypeScript file!

ui/src/app/base/components/TableConfirm/TableConfirm.tsx
huwshimi: Hmm yeah maybe, or maybe the `FormErrors` type should eventually become a generic type for the API error responses (kind of like I did for the setDefaultErrors case).

PR #2770, opened 2021-06-11
Add zone details page

sowasred2012:
## Done

- Added zone details content
- Added edit form
- Added form tests

## QA

- Visit http://0.0.0.0:8400/MAAS/r/zones
- Click any of the zones
- See that the details of the zone are listed there
- Click "Edit AZ"
- Update the form with whatever values you like, then click "Edit AZ", see that the changes are reflected in the details content page
- See that CI tests pass

Screenshot from 2021-06-11 10-39-40

### DIFF
--- a/ui/src/app/zones/views/ZoneDetails/ZoneDetailsContent/ZoneDetailsContent.tsx
+++ b/ui/src/app/zones/views/ZoneDetails/ZoneDetailsContent/ZoneDetailsContent.tsx
@@ -11,8 +11,7 @@ type Props = {
   id: number;
 };
 
-const ZoneDetailsContent = ({ id }: Props): JSX.Element => {
-  // const zonesLoaded = useSelector(zoneSelectors.loaded);
+const ZoneDetailsContent = ({ id }: Props): JSX.Element | null => {
   const zone = useSelector((state: RootState) =>
     zoneSelectors.getById(state, Number(id))
   );
--- a/ui/src/app/zones/views/ZoneDetails/ZoneDetailsContent/ZoneDetailsContent.tsx
+++ b/ui/src/app/zones/views/ZoneDetails/ZoneDetailsContent/ZoneDetailsContent.tsx
@@ -1,0 +1,58 @@
+import { useEffect } from "react";
+
+import { Row, Col } from "@canonical/react-components";
+import { useSelector, useDispatch } from "react-redux";
+
+import type { RootState } from "app/store/root/types";
+import { actions } from "app/store/zone";
+import zoneSelectors from "app/store/zone/selectors";
+
+type Props = {
+  id: number;
+};
+
+const ZoneDetailsContent = ({ id }: Props): JSX.Element => {
+  // const zonesLoaded = useSelector(zoneSelectors.loaded);
+  const zone = useSelector((state: RootState) =>
+    zoneSelectors.getById(state, Number(id))
+  );
+
+  const dispatch = useDispatch();
+  useEffect(() => {
+    dispatch(actions.fetch());
+  }, [dispatch]);
+
+  if (zone) {
+    return (
+      <>
+        <Row>
+          <Col size="2">
+            <p>Name:</p>
+          </Col>
+          <Col size="4">
+            <p>{zone.name}</p>
+          </Col>
+        </Row>
+        <Row>
+          <Col size="2">
+            <p>Description:</p>
+          </Col>
+          <Col size="4">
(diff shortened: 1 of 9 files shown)


ui/src/app/zones/views/ZoneDetails/ZoneDetailsContent/ZoneDetailsContent.tsx
```diff
@@ -0,0 +1,58 @@
...
+import type { RootState } from "app/store/root/types";
+import { actions } from "app/store/zone";
+import zoneSelectors from "app/store/zone/selectors";
+
+type Props = {
+  id: number;
+};
+
+const ZoneDetailsContent = ({ id }: Props): JSX.Element => {
+  // const zonesLoaded = useSelector(zoneSelectors.loaded);
```
ClementChaumel, line 15: Looks like some leftover comment 🧹

ui/src/app/zones/views/ZoneDetails/ZoneDetailsForm/ZoneDetailsForm.test.tsx
```diff
@@ -0,0 +1,81 @@
...
+  };
+  let initialState: RootState;
+
+  beforeEach(() => {
+    initialState = rootStateFactory({
+      zone: zoneStateFactory({
+        errors: {},
+        loading: false,
+        loaded: true,
+        items: [zoneFactory(testZone)],
```
Caleb-Ellis, line 31: I'm surprised this works! The way the test factories work is they set a bunch of defaults, but you can override them if you want, e.g:
``` js
zoneFactory(); // returns all defaults
zoneFactory({ name: "different-name" }); // returns all defaults, except name is now "different-name"
```
But `desc` is not a valid property of a zone - `description` is. So what this is doing is it's adding an extra key for `desc` and keeping the default `description` value.

``` js
{
  id: 1,
  desc: 'test description',
  name: 'zone-name',
  controllers_count: 273270286,
  created: 'Wed, 08 Jul. 2020 05:35:4',
  description: 'test description',
  devices_count: 1197561872,
  machines_count: 1150783015,
  updated: 'Wed, 08 Jul. 2020 05:35:4'
}
```

TS correctly complains if you try to set `desc` directly in `zoneFactory`, but defining `testZone` first seems to be ok. I'm not sure why that is. :confused:

ui/src/app/zones/views/ZoneDetails/ZoneDetailsForm/ZoneDetailsForm.tsx
```diff
@@ -0,0 +1,85 @@
...
+  const zone = useSelector((state: RootState) =>
+    zoneSelectors.getById(state, Number(id))
+  );
+
+  useEffect(() => {
+    dispatch(zoneActions.fetch());
+  }, [dispatch]);
+
+  let nameValue = null;
+  let descriptionValue = null;
```
Caleb-Ellis, line 37: These should be initialised as `""`, otherwise they're not `string` like what's defined in `CreateZoneValues`, but rather `string | null`.

But what I would actually do is render `null` for this component if `zone` hasn't loaded yet, otherwise `initialValues` gets set twice, once with the empty values and once again for the real values. This effectively means the component has to completely render twice.
``` typescript
if (zone) {
  return (
    <FormikForm<CreateZoneValues>
      initialValues={{
        description: zone.description,
        name: zone.name,
      }}
    />
  );
}
return null;

File: ui/src/app/zones/views/ZoneDetails/ZoneDetailsForm/ZoneDetailsForm.tsx [hunk 4]
```diff
@@ -0,0 +1,85 @@
...
+      cleanup={cleanup}
+      errors={errors}
+      initialValues={{
+        description: descriptionValue,
+        name: nameValue,
+      }}
+      onCancel={closeForm}
+      onSuccess={closeForm}
+      onSubmit={(values) => {
+        dispatch(
```
Caleb-Ellis, line 57: I would dispatch the cleanup function just before this too, to clear the errors while the update action occurs.
``` js
dispatch(cleanup());
```

Caleb-Ellis (approved):
This is looking great, thanks @bethcollins92 @sowasred2012!

I have some pretty minor comments but don't take them to be blockers. Don't forget you can always run`yarn betterer --update` if the strict mode errors are incomprehensible or too annoying to fix.

PR #2771, opened 2021-06-11
Add zone delete action

sowasred2012:
## Done

- Add delete zone action
- Add tests

Screenshot from 2021-06-11 16-54-02

## QA

- Visit http://0.0.0.0:8400/MAAS/r/zones
- Create a zone
- Click its name in the list
- Click "Delete AZ" and confirm when prompted
- See that the zone is not in the list

### DIFF
--- a/ui/src/app/base/components/FormikFormContent/FormikFormContent.tsx
+++ b/ui/src/app/base/components/FormikFormContent/FormikFormContent.tsx
@@ -79,7 +79,7 @@ const FormikFormContent = <V,>({
   inline,
   loading,
   onSaveAnalytics = {},
-  onSuccess,
+  onSuccess = () => void 0,
   onValuesChanged,
   resetOnSave,
   saved,
@@ -111,7 +111,7 @@ const FormikFormContent = <V,>({
 
   useEffect(() => {
     if (!errors && saved) {
-      onSuccess && onSuccess();
+      onSuccess();
     }
   }, [onSuccess, errors, saved]);
 
--- a/ui/src/app/zones/views/ZoneDetails/ZoneDetailsHeader/ZoneDetailsHeader.tsx
+++ b/ui/src/app/zones/views/ZoneDetails/ZoneDetailsHeader/ZoneDetailsHeader.tsx
@@ -1,35 +1,69 @@
-import { useEffect } from "react";
+import { useEffect, useState } from "react";
 
 import { Button } from "@canonical/react-components";
 import { useSelector, useDispatch } from "react-redux";
+import { useHistory } from "react-router-dom";
+
+import DeleteConfirm from "./DeleteConfirm";
 
 import SectionHeader from "app/base/components/SectionHeader";
 import type { RootState } from "app/store/root/types";
 import { actions } from "app/store/zone";
 import zoneSelectors from "app/store/zone/selectors";
+import zonesURLs from "app/zones/urls";
 
 type Props = {
   id: number;
 };
 
 const ZoneDetailsHeader = ({ id }: Props): JSX.Element => {
+  const [showConfirm, setShowConfirm] = useState(false);
   const zonesLoaded = useSelector(zoneSelectors.loaded);
   const zone = useSelector((state: RootState) =>
     zoneSelectors.getById(state, Number(id))
   );
   const dispatch = useDispatch();
+  const history = useHistory();
 
   useEffect(() => {
     dispatch(actions.fetch());
   }, [dispatch]);
 
+  const deleteZone = () => {
+    dispatch(actions.delete(id));
+    history.push({ pathname: zonesURLs.index });
+  };
+
+  const closeExpanded = () => setShowConfirm(false);
+
   let buttons: JSX.Element[] | null = [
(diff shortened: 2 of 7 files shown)


hatched:
@sowasred2012 @bethcollins92 After some digging...

- When you delete a zone the state that is returned by the [zoneSelectors.saved](https://github.com/canonical-web-and-design/maas-ui/blob/master/ui/src/app/zones/views/ZonesList/ZonesListForm/ZonesListForm.tsx#L23) is true 
- Which means that when you ultimately render the `FormikFormContent` on "Add AZ" it will pass `errors` that is `null` and `saved` which will be `true` which then makes the following conditional in this `useEffect` truthy and calls the supplied callback.
https://github.com/canonical-web-and-design/maas-ui/blob/4886dd778bdb02aaad18361264e41fb8b9c969e2/ui/src/app/base/components/FormikFormContent/FormikFormContent.tsx#L112-L116
- If you just click "Add AZ" without deleting first the `saved` value that's returned from the selector is `false` and so that `useEffect` never fires.

This ultimately appears to be a misunderstanding of how these selectors work and indeed it is confusing as `saved` doesn't reference a specific zone id as one would expect it does and repeated calls to it return different values. Which is why the second click to open the form works. cc @huwshimi @Caleb-Ellis 

I'm guessing that this `useEffect` was originally created so that the success callback would only be called after the api request was successful. So I think the fix here is to first understand how this `saved` state works and then make the appropriate changes.  Regardless of the outcome of that I think that this `useEffect` needs to be modified so that it's only ever run if the active form has an active request which is then completed (successfully or not).

ui/src/app/zones/views/ZoneDetails/ZoneDetailsHeader/ZoneDetailsHeader.tsx
```diff
@@ -1,45 +1,84 @@
...
+  useEffect(() => {
+    dispatch(zoneActions.fetch());
+  }, [dispatch]);
+
+  const deleteZone = () => {
+    dispatch(zoneActions.delete(id));
+    // we need to cleanup here but don't know how!
+    history.push({ pathname: zonesURLs.index });
+  };
+
```
ClementChaumel, line 37:
```suggestion
  const [showConfirm, setShowConfirm] = useState(false);
  const zonesLoaded = useSelector(zoneSelectors.loaded);
  const zonesSaved = useSelector(zoneSelectors.saved);
  const zone = useSelector((state: RootState) =>
    zoneSelectors.getById(state, Number(id))
  );
  const dispatch = useDispatch();
  const history = useHistory();

  useEffect(() => {
    dispatch(zoneActions.fetch());
  }, [dispatch]);

  useEffect(() => {
    if(zonesSaved){
      dispatch(zoneActions.cleanup());
       history.push({ pathname: zonesURLs.index });
     }
  }, [dispatch, zonesSaved ]);

  const deleteZone = () => {
    dispatch(zoneActions.delete(id));
  };

```

This is what I tried to suggest yesterday. Hope it helps

Caleb-Ellis:
@hatched is right on the money - `saved` is already `true` by the time the "Add AZ" form is rendered, so the `onSuccess` effect fires immediately.

The `cleanup` function reverts `saved` to false, amongst other things. The `FormikForm` component automatically dispatches that function when the component unmounts ([link](https://github.com/canonical-web-and-design/maas-ui/blob/master/ui/src/app/base/components/FormikFormContent/FormikFormContent.tsx#L131)), but this isn't the case with the new `DeleteConfirm` component (which is why I keep advocating to use `FormikForm` for everything so a lot of stuff is just "handled"!). So you could handle it in `DeleteConfirm`, orrrrrr use `FormikForm` and possibly update it so that it can have buttons inline with the children.

ui/src/app/zones/views/ZoneDetails/ZoneDetailsHeader/ZoneDetailsHeader.tsx
bethcollins92: Thanks! Makes sense

ui/src/app/base/components/FormikFormContent/FormikFormContent.tsx
```diff
@@ -79,7 +79,7 @@ const FormikFormContent = <V,>({
   inline,
   loading,
   onSaveAnalytics = {},
-  onSuccess,
+  onSuccess = () => void 0,
```
bartaz, line 82: Any specific reason for `void 0`? In other places I see simply `() => void`.

bartaz (approved):
LGTM, with one little question on `void 0`.

ui/src/app/base/components/FormikFormContent/FormikFormContent.tsx
sowasred2012: This is just setting a default value for the `onSuccess` parameter; `onSuccess` is optional, but we call it without checking whether anything was passed [here](https://github.com/canonical-web-and-design/maas-ui/blob/71aaa9f9f3b50ab697fcc29ff66cdecc8da42bcf/ui/src/app/base/components/FormikFormContent/FormikFormContent.tsx#L114). Looks neater, but requires this default value.

PR #2779, opened 2021-06-14
Add domains details header

ClementChaumel:
## Done

- Added the header to the domains details view complete with the number of hosts/resource records
- Added buttons to show the add and delete forms with dummy content
- Fixed a typo on "Authoriathitavtive" in the column name

## QA
Go to http://0.0.0.0:8400/MAAS/r/domain/0 and check that the header renders correctly

## Fixes

Fixes: #https://github.com/canonical-web-and-design/app-squad/issues/43

image


bethcollins92 (approved):
LGTM (and Scott) 👍🏼

ui/src/app/domains/views/DomainDetails/DomainDetailsHeader/DomainDetailsHeader.tsx
```diff
@@ -0,0 +1,79 @@
...
+  const domainsLoaded = useSelector(domainSelectors.loaded);
+
+  const [isDeleteFormOpen, setDeleteFormOpen] = useState(false);
+  const [isRecordFormOpen, setRecordFormOpen] = useState(false);
+
+  useEffect(() => {
+    dispatch(domainActions.fetch());
+  }, [dispatch]);
+
+  const getHostsString = () => {
```
hatched, line 29: I'd prefer if these functions were....functional.

Generally you want to avoid relying on the wrapper contexts for values when you can. So in this case and in the case below you could merge them into one function and that function would take a prefix and a value. This function could then be moved outside of the component so it's not being redefined all the time.

```js
function pluralizeString(prefix, count, emptyMessage) {
  if (count < 1) {
    return emptyMessage;
  }
return `${pluralize(prefix, count, true)}`;
}
```

ClementChaumel: Done! PTA

PR #2787, opened 2021-06-15
Domain details table

bartaz:
## Done

- Added links to nodes from domain details table (when available)
- Added sorting to domain resources table
- Added actions menu to domain resources table (with actions disabled until implemented)

### MAAS deployment

- [Bolla](/HACKING.md#bolla)

### Running the branch

You can run this branch by:

### QA steps

- `yarn start`
- go to http://0.0.0.0:8400/MAAS/r/domains
- click on different domains
- make sure resources table on domain details page is sortable and has action menu
- go to http://0.0.0.0:8400/MAAS/r/domain/0
- resources in the table should be linked to nodes
- actions menu should be disabled (as these are generated rows)

## Fixes

Fixes: canonical-web-and-design/app-squad#42

Screenshot 2021-06-15 at 16 16 57


bartaz:
@caleb @huwshimi Any tips on how to deal with betterer in this case?

ui/src/app/domains/views/DomainDetails/ResourceRecords/ResourceRecords.tsx
```diff
@@ -14,47 +26,121 @@ const ResourceRecords = ({ id }: Props): JSX.Element | null => {
...
+      sortKey: "data",
     },
     {
       content: "Actions",
+      className: "u-align--right",
     },
   ];
 
   const rows = domain.rrsets.map((resource) => {
+    const nodeType: NodeType = resource.node_type;
```
Caleb-Ellis, line 59: You're getting a strict mode error here because there's a disparity between the `NodeType` type that you've given this variable, and the `resource.node_type` type, which is `number | undefined` as defined in the `DomainResource` type.

A couple of things need to happen. First, `DomainResource.node_type` should use the type `NodeType` and not `number`, because `number` is too broad. Secondly, instead of being an optional parameter, it should actually be of type `NodeType | null` because the api actually explicitly sends `null` instead of not including the parameter at all. Then you shouldn't need to explicitly type `const nodeType` here since it'll be inferred to be `NodeType | null`.

I've had a quick look at the websocket code and it seems like the domain types should look something like this:
``` typescript
export enum RecordType {
  AAAA = "AAAA",
  A = "A",
  // etc
}

export type DomainResource = {
  dnsdata_id: number | null;
  dnsresource_id: number | null;
  name: string;
  node_type: NodeType | null;
  rrdata: string | null;
  rrtype: RecordType;
  system_id: BaseNode["system_id"] | null;
  ttl: number | null;
  user_id: User["id"] | null;
};
```

ui/src/app/domains/views/DomainDetails/ResourceRecords/ResourceRecords.tsx
```diff
@@ -14,47 +26,121 @@ const ResourceRecords = ({ id }: Props): JSX.Element | null => {
...
+    // We can't edit records that don't have a dnsresource_id.
+    // (If the row doesn't have one, it has probably been automatically
+    // generated by means of a deployed node, or some other reason.)
+    const isAutogenerated = !resource.dnsresource_id;
+    const canEdit = !isAutogenerated && isAdmin;
+
+    switch (nodeType) {
+      case NodeType.MACHINE:
+        nameCell = (
+          <Link to={machineURLs.machine.index({ id: resource.system_id })}>
```
Caleb-Ellis, line 71: The url function requires a string, but `resource.system_id` can be `string | null` (if using the type suggestions above). Intuitively we know that if `resource.node_type` is not `null`, then `resource.system_id` will also not be `null`, but the TS compiler doesn't know that. So what we actually want to do is check:
``` typescript
if (resource.node_type === NodeType.MACHINE && resource.system_id !== null) {
  // do the thing
}
```

Caleb-Ellis (approved):
LGTM! Just some ts strict mode suggestions

PR #2799, opened 2021-06-17
Add domain not found header

ClementChaumel:
## Done

- Add domain not found header to the domain detail page if the domain doesn't exist

## QA

go to http://0.0.0.0:8400/MAAS/r/domain/123456789
check that it's showing the correct header

Legacy : 
image

New and improved :sparkles: 
image


Caleb-Ellis (approved):
LGTM!

PR #2803, opened 2021-06-17
Add record updates

ClementChaumel:
## Done

- Updated Domain components to use an `id` prop instead of parsing it from the url in each component. This will make it easier if, for example, we want to render a component somewhere other than `/domain/{id}`.
- Separated `createRecord` action into two separate actions `createAddressRecord` and `createDNSData` (and typed params) to closer match the websocket api.
- Fixed a bug where TTL would not save for A and AAAA records.
- Added some help text for each record type.

## QA steps

- Go to /MAAS/r/domains and click "Add record"
- Check that you can successfully add each record type with TTL saved
- Check that you can still delete a domain that has no records


huwshimi (approved):
Nice work, the help text definitely helped me out!

PR #2805, opened 2021-06-17
WIP: Add discovery actions

steverydz:


steverydz:
@Caleb-Ellis @huwshimi 

For the discovery actions form, this is what is outstanding:

- The hostname isn't saved
- The type, domain, device name, IP assignment and parent fields aren't pre-populated with saved values
- Changed values aren't saved
- No delete discovery

huwshimi:
I've broken this work down into a number of PRs so I'll close this one.

PR #2806, opened 2021-06-18
feat(ui): add websocket polling

huwshimi:
## Done

- Add the ability to poll an action using the websocket.

### MAAS deployment

- [Bolla](/HACKING.md#bolla)

### Running the branch

You can run this branch by:

### QA steps

- Visit /r/images and open the redux tools to see the dispatches.
- You should see some start polling actions.
- Wait 10 seconds and you should see the next poll happen.
- Click on Settings in the header and you should see the stop polling actions.
- Wait 10 seconds and you should not see any more poll events.

## Fixes

Fixes: canonical-web-and-design/app-squad#87.

### DIFF
--- a/ui/src/app/base/sagas/websockets.ts
+++ b/ui/src/app/base/sagas/websockets.ts
@@ -35,7 +35,7 @@ import type {
 
 import { fileContextStore } from "app/base/file-context";
 
-const DEFAULT_POLL_INTERVAL = 10;
+const DEFAULT_POLL_INTERVAL = 10000;
 
 type ActionCreator = (...args: unknown[]) => WebSocketAction;
 
--- a/ui/src/app/base/sagas/websockets.ts
+++ b/ui/src/app/base/sagas/websockets.ts
@@ -10,6 +10,10 @@ import type { SagaGenerator } from "typed-redux-saga/macro";
 import {
   all,
   call,
+  cancel,
+  cancelled,
+  delay,
+  fork,
   put,
   take,
   takeEvery,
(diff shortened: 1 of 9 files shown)


ui/src/app/base/sagas/websockets.ts
```diff
@@ -337,6 +353,54 @@ export function* handleFileContextRequest({
...
+ * @param action - A websocket action.
+ */
+export function* pollAction(action: WebSocketAction): SagaGenerator<void> {
+  const endpoint = `${action.meta?.model}.${action.meta?.method}`;
+  try {
+    while (true) {
+      // The delay is put first as the action will have already been handled
+      // when it is first dispatched, so this should start by waiting until the
+      // next interval.
+      yield* delay((action.meta?.pollInterval || DEFAULT_POLL_INTERVAL) * 1000);
```
Caleb-Ellis, line 367: I feel like we should just work with milliseconds because that's how all JS timers work by default.

ui/src/websocket-client.ts
```diff
@@ -54,10 +54,14 @@ export type WebSocketResponseNotify = {
   type: WebSocketMessageType.NOTIFY;
 };
 
+export type WebSocketActionParams =
+  | Record<string, unknown>
```
Caleb-Ellis, line 58: could use the `AnyObject` type here

Caleb-Ellis (approved):
LGTM! Only thing I'll say is the "Loading" label in the header cycles between the "Step 1: Region... etc" label while polling is happening. Maybe we should just remove the "Loading" label? Or at least have it be tracking something other than `polling`?

ui/src/app/base/sagas/websockets.ts
huwshimi: Good idea.

ui/src/websocket-client.ts
huwshimi: Nice, thanks.

huwshimi:
> LGTM! Only thing I'll say is the "Loading" label in the header cycles between the "Step 1: Region... etc" label while polling is happening. Maybe we should just remove the "Loading" label? Or at least have it be tracking something other than `polling`?

Yeah, maybe I'll do something as a followup to keep this branch small.

PR #2807, opened 2021-06-18
feat(ui): only show the polling state on the first load

huwshimi:
## Done

- Only show the polling state in the images header on the first load.

### MAAS deployment

- [Bolla](/HACKING.md#bolla)

### Running the branch

You can run this branch by:

### QA steps

- Go to /r/images.
- You should see the loading state in the header.
- Wait 10 seconds and the poll will happen again, but you should not see it in the header.


Caleb-Ellis (approved):
LGTM! I was thinking we would add a `loaded` boolean to bootresource state but this is muuuuuch simpler

PR #2808, opened 2021-06-18
fix(ui): update dashboard file organisation + config form

Caleb-Ellis:
## Done

- Updated the dashboard files so that only the content below the header is routed instead of each tab being entirely different components
- Added tests + fixed some minor bugs with the dashboard config form

### MAAS deployment

- [Bolla](/HACKING.md#bolla)

### Running the branch

You can run this branch by:

### QA steps

- Go to /MAAS/r/dashboard/configuration and check that the values save correctly
- Refresh the page and check that there are no console errors


ui/src/app/dashboard/views/Dashboard.tsx
```diff
@@ -1,24 +1,31 @@
...
+          <DiscoveriesList />
+        </Route>
+        <Route exact path={[dashboardURLs.configuration]}>
+          <DashboardConfigurationForm />
+        </Route>
+        <Route path="*">
+          <NotFound />
+        </Route>
+      </Switch>
+    </Section>
```
Caleb-Ellis, line 28: @petesfrench @steverydz I've updated this component so that the Section and DashboardHeader components are consistent between routes, and only the content below the header changes with the route.

ui/src/app/dashboard/views/DashboardConfigurationForm/DashboardConfigurationForm.tsx
```diff
@@ -1,21 +1,16 @@
...
 
 const DashboardConfigurationForm = (): JSX.Element => {
-  useWindowTitle("Dashboard");
+  useWindowTitle("Dashboard configuration");
 
   return (
-    <Section
-      header={<DiscoveriesListHeader />}
-      headerClassName="u-no-padding--bottom"
-    >
```
Caleb-Ellis, line 15: So now we don't have to redefine it for each tab

ui/src/app/dashboard/views/DashboardConfigurationForm/DashboardConfigurationSubnetForm/DashboardConfigurationSubnetForm.test.tsx
```diff
@@ -0,0 +1,120 @@
...
+
+    expect(wrapper.find("a[data-test='subnet-link']").prop("href")).toBe(
+      "/MAAS/l/subnet/1"
+    );
+    expect(wrapper.find("a[data-test='fabric-link']").prop("href")).toBe(
+      "/MAAS/l/fabric/3"
+    );
+  });
+
+  it("dispatches actions to update subnet active discovery if they have changed", () => {
```
Caleb-Ellis, line 84: @petesfrench I've added some extra tests to the form, but this one in particular is important. We're testing that, given the Redux store is set up in some way and given that the form is submitted with these values, these Redux actions should have been dispatched. It's a bit cumbersome to have to set up the store for these kinds of tests but it's necessary for the whole React "UI is a function of state" thing.

ui/src/app/dashboard/views/DashboardConfigurationForm/DashboardConfigurationSubnetForm/DashboardConfigurationSubnetForm.tsx
```diff
@@ -0,0 +1,115 @@
...
+import LegacyLink from "app/base/components/LegacyLink";
+import baseURLs from "app/base/urls";
+import { actions as fabricActions } from "app/store/fabric";
+import fabricSelectors from "app/store/fabric/selectors";
+import { actions as subnetActions } from "app/store/subnet";
+import subnetSelectors from "app/store/subnet/selectors";
+import type { Subnet } from "app/store/subnet/types";
+
+type SubnetDiscoveryValues = {
+  [x: number]: Subnet["active_discovery"];
```
Caleb-Ellis, line 17: @petesfrench I've explicitly typed the form values here, otherwise you get a strict mode error. The type is a bit hard to parse, but it just means an object with numbered keys and boolean values (because subnet.active_discovery is a boolean).
``` typescript
{
  1: true,
  2: false,
}

File: ui/src/app/dashboard/views/DashboardConfigurationForm/DashboardConfigurationSubnetForm/DashboardConfigurationSubnetForm.tsx [hunk 5]
```diff
@@ -0,0 +1,115 @@
...
+  const fabricsLoaded = useSelector(subnetSelectors.loaded);
+  const saved = useSelector(subnetSelectors.saved);
+  const saving = useSelector(subnetSelectors.saving);
+
+  useEffect(() => {
+    dispatch(subnetActions.fetch());
+    dispatch(fabricActions.fetch());
+  }, [dispatch]);
+
+  const loaded = subnetsLoaded && fabricsLoaded;
```
Caleb-Ellis, line 34: @petesfrench Since this form requires both subnets and fabrics we should check that they're both loaded before rendering. Otherwise you might get the case where subnets have loaded but not fabrics, and the component doesn't know how to render the fabric link.

ui/src/app/dashboard/views/DashboardConfigurationForm/DashboardConfigurationSubnetForm/DashboardConfigurationSubnetForm.tsx
```diff
@@ -0,0 +1,115 @@
...
+                    <>
+                      <span> on </span>
+                      <LegacyLink
+                        data-test="fabric-link"
+                        route={baseURLs.fabric({ id: targetFabric.id })}
+                      >
+                        {targetFabric.name}
+                      </LegacyLink>
+                    </>
+                  )}
```
Caleb-Ellis, line 90: @petesfrench We use this `LegacyLink` component for all legacy links. Just means we don't have to hardcode in the whole routes. Also I've made the fabric link bit conditional, in the off chance the subnet's fabric can't be found from the list of fabrics.

ui/src/app/dashboard/views/DashboardConfigurationForm/DashboardConfigurationSubnetForm/DashboardConfigurationSubnetForm.tsx
```diff
@@ -0,0 +1,115 @@
...
+  }
+
+  return (
+    <Strip shallow>
+      <h3 className="p-heading--4">Subnet mapping</h3>
+      <p>
+        Active discovery (subnet mapping) can be enabled below on the configured
+        subnets. Each rack will scan the subnets that allow it. This helps
+        ensure discovery information is accurate and complete.
+      </p>
```
Caleb-Ellis, line 109: @petesfrench I've also added a little context about what the checkboxes actually mean (just lifted from the legacy version's help text)

ui/src/app/store/subnet/types/actions.ts
```diff
@@ -21,6 +21,20 @@ export type CreateParams = {
...
+  description?: Subnet["description"];
+  disabled_boot_architectures?: string;
+  dns_servers?: Subnet["dns_servers"];
+  fabric?: Fabric[FabricMeta.PK];
+  gateway_ip?: Subnet["gateway_ip"];
+  managed?: Subnet["managed"];
+  name?: Subnet["name"];
+  rdns_mode?: Subnet["rdns_mode"];
+  vid?: VLAN["vid"];
+  vlan?: Subnet["vlan"];
```
Caleb-Ellis, line 39: @huwshimi I think we might need to go through these UpdateParams because most of the time we only need the id then the rest are optional.

huwshimi: Sure, maybe this is a case for union types that match what the API expects? Although, looking at the subnet form it looks like there are [a lot of different related fields](https://github.com/maas/maas/blob/6d3edc8baffc56fad32c58ad6fecda3b0e19a966/src/maasserver/forms/subnet.py#L143) so maybe we'd introduce bugs when trying to mirror that here...

huwshimi (approved):
Looks good

PR #2826, opened 2021-06-23
feat(ui): add saga for updating a DNS resource then updating the record

### LINKED ISSUE #2818: [Domains] Add saga for changing record name then updating record data
Seems like something in the api requires the name of an address record to be able to update its address. So we need to first run `update_dnsresource` to change the name, then with that name changed we can run `update_address_record`.

Caleb-Ellis:
## Done

- Migrated actions saga to TypeScript (badly - added 4 `TSFixMe`s)
- Added a saga `updateDomainRecord` for handling what to do when updating a record that could have its name changed and be either an address record or non-address record

### MAAS deployment

- [Bolla](/HACKING.md#bolla)

### Running the branch

You can run this branch by:

### QA steps

- Do to /MAAS/r/domains and click on a domain
- Add an A record and then open the edit form for it
- Change all the fields and submit
- Check that the `updateDNSResource` action is dispatched first, then when that is complete `updateAddressRecord` is dispatched
- Add a TXT record and then open the edit form for it
- Change all the fields and submit
- Check that the `updateDNSResource` action is dispatched first, then when that is complete `updateDNSData` is dispatched
- Check that all the values update correctly

## Fixes

Fixes #2818


ui/src/app/base/sagas/actions.ts
```diff
@@ -0,0 +1,127 @@
...
+ * @param sendMessage - The saga that handles sending websocket messages.
+ * @param action - The redux action with pool and machine data.
+ */
+export function* createPoolWithMachines(
+  socketClient: WebSocketClient,
+  sendMessage: TSFixMe,
+  { payload }: PayloadAction<{ params: CreateWithMachinesParams }>
+): SagaGenerator<void> {
+  const { machineIDs, pool } = payload.params;
+  const actionCreators = yield call(
```
huwshimi, line 50: To have these return the correct types we need to use `yield* call` from `import { call } from "typed-redux-saga/macro";`

huwshimi (approved):
As mentioned I'll address these (and some more) in a commit.

PR #2831, opened 2021-06-25
feat(ui): add discovery form

huwshimi:
## Done

- Add discovery form for creating devices/interfaces.

### MAAS deployment

- [Bolla](/HACKING.md#bolla)

### Running the branch

You can run this branch by:

### QA steps

- Visit /r/dashboard.
- Open the action menu for a discovery and click "Add discovery".
- Try adding a device (don't choose a parent, this doesn't currently work: https://bugs.launchpad.net/maas/+bug/1933408).
- Repeat the process, but this time create an interface.

## Fixes

Fixes: canonical-web-and-design/app-squad#71.

### DIFF
--- a/ui/src/app/dashboard/views/DiscoveriesList/DiscoveriesList.tsx
+++ b/ui/src/app/dashboard/views/DiscoveriesList/DiscoveriesList.tsx
@@ -13,6 +13,8 @@ import classNames from "classnames";
 import { useSelector, useDispatch } from "react-redux";
 import type { Dispatch } from "redux";
 
+import DiscoveryAddForm from "../DiscoveryAddForm";
+
 import DoubleRow from "app/base/components/DoubleRow";
 import TableDeleteConfirm from "app/base/components/TableDeleteConfirm";
 import { useWindowTitle } from "app/base/hooks";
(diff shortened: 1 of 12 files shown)


ui/src/app/dashboard/views/DiscoveriesList/DiscoveriesList.tsx
```diff
@@ -71,74 +84,80 @@ const generateRows = (
...
-              toggleAppearance="base"
-            />
-          ),
-          className: "u-align--right",
-        },
-      ],
+      columns:
+        // Hide the columns when the add form is visible.
+        isExpanded && expandedRow?.type === ExpandedType.ADD
+          ? []
```
Caleb-Ellis, line 90: I know this is how the legacy version works but it seems like a weird pattern to hide the column data and then show it again in a different place. We could also probably add the network details to the columns if we used double rows. And the name field being above the rest of the fields, for some reason? Just a bit weird.

Caleb-Ellis (approved):
LGTM!

ui/src/app/dashboard/views/DiscoveriesList/DiscoveriesList.tsx
huwshimi: I actually started to rework it to be consistent with newer parts of the app and then saw that domains does that same name field at the top thing. I'll have another look, but might be good to get some design guidance on it.

PR #2842, opened 2021-06-28
feat(ui): add delete action to images table

Caleb-Ellis:
## Done

- Added delete action for images
- Updated bootresource action reducers to handle success responses (which are identical to the poll success response)

### MAAS deployment

- [Bolla](/HACKING.md#bolla)

### Running the branch

You can run this branch by:

### QA steps

- Go to /MAAS/r/images
- Check that images that have already been downloaded or are in the process of being downloaded (i.e. they exist in `bootresource.resources`) have a bin button
- Check that you can delete an image if it does not use the default commissioning release, which will always be an Ubuntu release (so Other images should always be deletable). The resource will stay in the table until state updates when the server is polled again.
- Check that you can't delete an image that uses the default commissioning release.

## Fixes

Fixes canonical-web-and-design/app-squad#142

### DIFF
--- a/ui/src/app/images/components/ImagesTable/ImagesTable.tsx
+++ b/ui/src/app/images/components/ImagesTable/ImagesTable.tsx
@@ -1,10 +1,18 @@
+import { useState } from "react";
+
 import { Icon, MainTable, Spinner } from "@canonical/react-components";
+import classNames from "classnames";
 import { useFormikContext } from "formik";
+import { useSelector } from "react-redux";
+
+import DeleteImageConfirm from "./DeleteImageConfirm";
 
 import DoubleRow from "app/base/components/DoubleRow";
+import TableActions from "app/base/components/TableActions";
 import type { ImageValue } from "app/images/types";
 import type { BootResource } from "app/store/bootresource/types";
 import { splitResourceName } from "app/store/bootresource/utils";
+import configSelectors from "app/store/config/selectors";
 
 type Props = {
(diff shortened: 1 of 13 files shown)


huwshimi:
There seems to be a lot of odd statuses being reported, probably most of it come from the API. The only thing I can reliable reproduce is that when you delete an image it shows "Selected for download" and never gets removed from the list.

Screen Shot 2021-06-29 at 9 20 06 am

ui/src/app/images/components/ImagesTable/ImagesTable.tsx
```diff
@@ -64,14 +71,26 @@ const generateImageRow = (image: ImageValue) => {
...
+const generateResourceRow = (
+  resource: BootResource,
+  commissioningRelease: string | null,
+  expanded: BootResource["id"] | null,
+  setExpanded: (id: BootResource["id"] | null) => void,
+  unchecked = false
+) => {
+  const { os, release } = splitResourceName(resource.name);
+  const canBeDeleted = !(os === "ubuntu" && release === commissioningRelease);
+  const isExpanded = expanded === resource.id;
```
huwshimi, line 89: The row also needs to add the active class when it is expanded to show the white background:
```typescript
className: classNames("p-table__row", {
        "is-active": isExpanded,
      }),
```

huwshimi (approved):
Looks good, one comment.

PR #2844, opened 2021-06-29
feat(ui): add discovery success message

huwshimi:
## Done

- Show a message when a device has been added.

### MAAS deployment

- [Bolla](/HACKING.md#bolla)

### Running the branch

You can run this branch by:

### QA steps

- Go to /r/dashboard.
- Open a row menu and click "Add discovery" and click the green button.
- You should see a success message at the top of the list.

## Fixes

Fixes: canonical-web-and-design/app-squad#132.

### DIFF
--- a/ui/src/app/base/components/FormikFormContent/FormikFormContent.tsx
+++ b/ui/src/app/base/components/FormikFormContent/FormikFormContent.tsx
@@ -38,7 +38,7 @@ export type Props<V> = {
     category?: string;
     label?: string;
   };
-  onSuccess?: () => void;
+  onSuccess?: (values: V) => void;
   onValuesChanged?: (values: V) => void;
   resetOnSave?: boolean;
   saved?: boolean;
@@ -81,7 +81,7 @@ const FormikFormContent = <V,>({
   inline,
   loading,
   onSaveAnalytics = {},
-  onSuccess = () => void 0,
+  onSuccess,
   onValuesChanged,
   resetOnSave,
   saved = false,
@@ -98,6 +98,7 @@ const FormikFormContent = <V,>({
     allowUnchanged,
   });
   const hasSaved = useCycled(saved);
+  const hasErrors = !!errors;
 
   // Run onValuesChanged function whenever formik values change.
   useEffect(() => {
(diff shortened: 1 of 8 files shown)


ui/src/app/base/components/FormikFormContent/FormikFormContent.tsx
```diff
@@ -113,10 +114,10 @@ const FormikFormContent = <V,>({
   }, [initialValues, resetForm, resetOnSave, saved]);
 
   useEffect(() => {
-    if (!errors && hasSaved) {
-      onSuccess();
+    if (!hasErrors && hasSaved) {
+      onSuccess && onSuccess(values);
     }
-  }, [onSuccess, errors, hasSaved]);
+  }, [onSuccess, hasErrors, hasSaved, values]);
```
Caleb-Ellis, line 120: Do you think there could be unintended side-effects if this `useEffect` runs every time `values` changes?

Caleb-Ellis (approved):
LGTM! Just one question about the `useEffect` but I think it might be fine

ui/src/app/base/components/FormikFormContent/FormikFormContent.tsx
huwshimi: It's going too run quite a lot, but I don't think that will be a problem? Was there anything you were thinking of? Or can you think of a better way to do this?

Caleb-Ellis: I'd just been considering the case where the form doesn't close on success, i.e. the cleanup function doesn't get run to reset `saved`. In that scenario the `onSuccess` function will run every time values changes even though we only really want it to run when `hasSaved` changes from false to true.

I don't think it's an issue with any of our forms now (although maybe the "Save and add another" ones?) but it might be safer to only run the `onSuccess` function if `hasSaved` changes from false to true and not just when `hasSaved` is true. I guess one way of handling it would be with an internal `hasSaved` state that you explicitly set to false in the `useEffect`.

huwshimi: Ah yeah I see. `hasSaved` does currently only get set when `saved` has cycled from `false` to `true`, but once it has cycled there's nothing to stop it from firing again, even if `saved` goes back to `false`.

How about when it calls `onSuccess` it sets a ref so that `onSuccess` can only run once. Forms that get reused (e.g. the save-and-add-another forms that you mentioned) call `resetOnSave` so when it resets the form it could also reset `onSuccessCalled`  and `hasSaved`?

Caleb-Ellis: Yeah that sounds good!

huwshimi: I've pushed that change. I've left it as a separate commit so you can just review the changes.

Caleb-Ellis: LGTM

PR #2845, opened 2021-06-29
feat(ui): build custom images table

Caleb-Ellis:
## Done

- Built Custom Images table
- Added selectors for the rest of the tables, and fixed the selectors used for Ubuntu images and Other images (they're only supposed to be synced resources)

### MAAS deployment

- [Bolla](/HACKING.md#bolla)

### Running the branch

You can run this branch by:

### QA steps

- Connect to labmaas and go to /MAAS/r/images
- Check that there is a section for Custom images that lists the data correctly

## Fixes

Fixes canonical-web-and-design/app-squad#83


ui/src/app/images/views/ImageList/CustomImages/CustomImages.tsx
```diff
@@ -0,0 +1,46 @@
...
+  }, []);
+
+  return (
+    <>
+      <hr />
+      <Strip shallow>
+        <h4>Custom images</h4>
+        {/* ImagesTable requires formik context in order to be used in conjunction
+          with the image selection checkboxes for synced resources,
+          so we wrap it in Formik despite these resources being uploaded */}
```
huwshimi, line 37: I think maybe instead we should split up ImagesTable into two components, one that has the table and accepts an array of images and another array of unchecked resources (maybe just the ids) and a second component that contains the formik parts which would wrap the first component. Then CustomImages can just use the first component.

Caleb-Ellis: I could probably just make `images` a prop instead of using `useFormikContext` in the component. Think that would work?

huwshimi: Yeah I think so.

Caleb-Ellis: @huwshimi ok should be ready for another look now. I've moved `images` to be a `prop` and also removed the `onSuccess` from `DeleteImageConfirm`. Turns out just using the Formik `enableReinitialize` prop achieves the same thing (removing an image from the form values if it no longer exists)

huwshimi (approved):
Looks good, I think that makes more sense now.

PR #2850, opened 2021-06-30
refactor(ui): generic node filtering

huwshimi:
## Done

- Update the search and filter code to work with any node type.

### MAAS deployment

- [Bolla](/HACKING.md#bolla)

### Running the branch

You can run this branch by:

### QA steps

- Visit the machine list and check you can still search and filter.
- Visit a KVM details page and check that you can still search the VMs.

## Fixes

Fixes: canonical-web-and-design/app-squad#136.


ui/src/app/utils/search/filter-handlers.ts
```diff
@@ -0,0 +1,230 @@
...
+  );
+};
+
+export const isFilterValueArray = (
+  nodeValue: unknown
+): nodeValue is FilterValue[] => {
+  return Array.isArray(nodeValue) && isFilterValue(nodeValue[0]);
+};
+
+export default class FilterHandlers {
```
huwshimi, line 25: I moved these utilities to a class so that the `prefixedFilters` (e.g. for workload annotations) are available to all utilities.

ui/src/app/utils/search/filter-nodes.ts
```diff
@@ -0,0 +1,177 @@
+import FilterHandlers from "./filter-handlers";
+import type { Filters, FilterValue, PrefixedFilter } from "./filter-handlers";
+
+export type GetValue<N> = (
+  node: N,
+  filter: string
+) => FilterValue | FilterValue[] | null;
+
+export default class FilterNodes<N, PK extends keyof N> extends FilterHandlers {
```
huwshimi, line 9: These filter utilities are also now in a class so that the generics and underlying filters for `prefixedFilters` (workload annotations) are set up in one place.

ui/src/app/utils/search/filter-handlers.ts
```diff
@@ -0,0 +1,230 @@
...
+   * Toggles a filter on or off based on type and value.
+   * @param filters - The initial filters.
+   * @param type - The filter key.
+   * @param value - The filter value to toggle.
+   * @param exact - Optional value for whether the value should
+   * exactly match.
+   * @param shouldExist - An optional value for whether the value should
+   * exist or not i.e. if true and the value exists there will be no change and
+   * vice versa.
+   * @returns Tag loading state.
```
Caleb-Ellis, line 170: No it doesn't!

ui/src/app/utils/search/filter-nodes.ts
Caleb-Ellis, line 9: Since the plan is for this to be used with discoveries as well, I think `FilterNodes` might be a bit of a misnomer since nodes in MAAS-land are specifically `Machine | Device | Controller`. Maybe `FilterObjects` if that's not too vague?

ui/src/app/utils/search/filter-nodes.ts
```diff
@@ -0,0 +1,177 @@
...
+            }
+            // If an exclude was found then exit the loop.
+            return exclude;
+          });
+        });
+      });
+      return matched && !exclude;
+    });
+
+  filterNodes = (nodes: N[], search: string, selectedIDs: N[PK][]): N[] => {
```
Caleb-Ellis, line 135: Same here - "nodes" feels like it could be misinterpreted.

Caleb-Ellis (approved):
Just a naming suggestion but otherwise LGTM! I guess we'll see when these generic functions are used in earnest whether they work across model types

ui/src/app/utils/search/filter-handlers.ts
huwshimi: Hah, thanks.

ui/src/app/utils/search/filter-nodes.ts
huwshimi: For the generic utils we use `item` and `I` so maybe we should carry that over here (also on the state objects they're stored in an `items` array). Does that sound OK?

Caleb-Ellis: Yeah that sounds good

PR #2861, opened 2021-07-02
feat(ui) build fetched images table

Caleb-Ellis:
## Done

- Built FetchedImages table which is used to sync images with a different source

### MAAS deployment

- [Bolla](/HACKING.md#bolla)

### Running the branch

You can run this branch by:

### QA steps

- Go to /MAAS/r/images and click "Change source"
- Use a source type that's different to the current selection and click "Connect"
- Check that an image selection table shows
- Select an image and click "Update selection"
- It might take a while but eventually you should get taken back to the main image list, and the header should say "Syncing images from {the source URL you chose}"
- Wait until images start downloading and check that the "Change source" button becomes disabled
- Click "Stop import".
- It might take a while but eventually downloading will stop, then check that the "Change source" button becomes enabled again.

## Fixes

Fixes canonical-web-and-design/app-squad#141

### DIFF
--- a/ui/src/app/images/components/UbuntuImageSelect/ArchSelect/ArchSelect.tsx
+++ b/ui/src/app/images/components/UbuntuImageSelect/ArchSelect/ArchSelect.tsx
@@ -4,6 +4,7 @@ import { useSelector } from "react-redux";
 
 import type { ImageValue } from "app/images/types";
 import type {
+  BaseImageFields,
   BootResourceUbuntuArch,
   BootResourceUbuntuRelease,
 } from "app/store/bootresource/types";
@@ -11,7 +12,7 @@ import configSelectors from "app/store/config/selectors";
 
 type Props = {
   arches: BootResourceUbuntuArch[];
-  release: BootResourceUbuntuRelease | null;
+  release: BootResourceUbuntuRelease | BaseImageFields | null;
 };
 
 const ArchSelect = ({ arches, release }: Props): JSX.Element => {
@@ -45,6 +46,7 @@ const ArchSelect = ({ arches, release }: Props): JSX.Element => {
     );
 
   const archUnsupported = (arch: BootResourceUbuntuArch) =>
+    "unsupported_arches" in release &&
     release.unsupported_arches.includes(arch.name);
 
   const isLastCommissioningArch = (arch: BootResourceUbuntuArch) => {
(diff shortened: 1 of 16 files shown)


ui/src/app/images/components/UbuntuImageSelect/ArchSelect/ArchSelect.tsx
```diff
@@ -4,14 +4,15 @@ import { useSelector } from "react-redux";
...
+  BaseImageFields,
   BootResourceUbuntuArch,
   BootResourceUbuntuRelease,
 } from "app/store/bootresource/types";
 import configSelectors from "app/store/config/selectors";
 
 type Props = {
   arches: BootResourceUbuntuArch[];
-  release: BootResourceUbuntuRelease | null;
+  release: BootResourceUbuntuRelease | BaseImageFields | null;
```
Caleb-Ellis, line 15: I've had to update a lot of the `UbuntuImageSelect` props to work with data coming from both `ubuntu.releases` and `fetchedImages.releases`. I think this is a bug in the backend as they should be the same, i.e both have the key `unsupported_arches`

huwshimi: Might be worth leaving a comment about when you get the different types.

huwshimi (approved):
This looks good. I think this is starting to take shape nicely.

One thing I found the first time I went through the flow is that when I got to the "Showing images fetched from..." it wasn't immediately obvious that I was still in the flow and not back at the main list. I wonder if the forms should be shown in the form card and then it's clear when you're still working through the flow (the machine list etc. use the card when you're shown the add forms).

One other thing was that when I pressed cancel from the second form (the "showing images..." screen) it went back to the "Choose source" form, but I think maybe it makes more sense to cancel the flow completely (otherwise you have to press cancel twice).

ui/src/app/images/components/UbuntuImageSelect/ArchSelect/ArchSelect.tsx
Caleb-Ellis: Yeah good call

Caleb-Ellis:
> This looks good. I think this is starting to take shape nicely.
> 
> One thing I found the first time I went through the flow is that when I got to the "Showing images fetched from..." it wasn't immediately obvious that I was still in the flow and not back at the main list. I wonder if the forms should be shown in the form card and then it's clear when you're still working through the flow (the machine list etc. use the card when you're shown the add forms).
> 
> One other thing was that when I pressed cancel from the second form (the "showing images..." screen) it went back to the "Choose source" form, but I think maybe it makes more sense to cancel the flow completely (otherwise you have to press cancel twice).

I can fix the second comment now, but maybe the first comment can be part of the UX images rework issue? I remember Lyubo saying we're trying to move away from using cards for forms (or maybe it was just the highlighted cards). But yeah I can see how right now there's really not much difference between the main list and the fetched list.

huwshimi:
> > This looks good. I think this is starting to take shape nicely.
> > One thing I found the first time I went through the flow is that when I got to the "Showing images fetched from..." it wasn't immediately obvious that I was still in the flow and not back at the main list. I wonder if the forms should be shown in the form card and then it's clear when you're still working through the flow (the machine list etc. use the card when you're shown the add forms).
> > One other thing was that when I pressed cancel from the second form (the "showing images..." screen) it went back to the "Choose source" form, but I think maybe it makes more sense to cancel the flow completely (otherwise you have to press cancel twice).
> 
> I can fix the second comment now, but maybe the first comment can be part of the UX images rework issue? I remember Lyubo saying we're trying to move away from using cards for forms (or maybe it was just the highlighted cards). But yeah I can see how right now there's really not much difference between the main list and the fetched list.

Sure.

Issue #2881, opened 2021-07-19
The correct VLAN is not selected when creating bonds


huwshimi:
When adding a bond the VLAN should match the VLAN of the parent interfaces, but the select box is defaulting to the first option.

To reproduce:
- Open the network tab for a machine.
- Create two interfaces with the same VLAN, but make sure that the VLAN isn't the default (i.e. choose a fabric that has multiple VLANs and choose a VLAN that is not the first option).
- Tick both interfaces and click 'Create bond'.
- See that the VLAN selected in the disabled field is the first option and not the VLAN that was set above.

PR #2882, opened 2021-07-20
fix(ui): select correct VLAN when creating a bond

huwshimi:
## Done

- Fix selecting the correct VLAN when adding a bond.

### MAAS deployment

- [Bolla](/HACKING.md#bolla)

### Running the branch

You can run this branch by:

### QA steps

- Open the network tab for a machine (at the time of writing 'top owl' on Bolla is set up correctly)
- If not using 'top owl' then create two interfaces with the same VLAN, but make sure that the VLAN isn't the default (i.e. choose a fabric that has - multiple VLANs and choose a VLAN that is not the first option).
- Tick both interfaces and click 'Create bond'.
- See that the VLAN selected in the disabled field is the VLAN that was set above.

## Fixes

Fixes: #2881.

## Launchpad issue

## Backports


Caleb-Ellis (approved):
LGTM!

PR #2905, opened 2021-07-27
fix(cypress): fix images and intro Cypress tests

Caleb-Ellis:
## Done

- Updated intro and images cypress tests to point to the react versions
- Updated intro pages to take cookies into account when redirecting

## QA

- N/A - tests should pass

### DIFF
--- a/integration/cypress/integration/base/header.spec.ts
+++ b/integration/cypress/integration/base/header.spec.ts
@@ -19,7 +19,7 @@ context("Header", () => {
   it("navigates to the machine list when clicking on the logo as a non admin", () => {
     cy.get(".p-navigation__logo .p-navigation__link").click();
     cy.location("pathname").should("eq", generateNewURL("/machines"));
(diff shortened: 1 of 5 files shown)


integration/cypress/integration/base/header.spec.ts
```diff
@@ -19,7 +19,7 @@ context("Header", () => {
   it("navigates to the machine list when clicking on the logo as a non admin", () => {
     cy.get(".p-navigation__logo .p-navigation__link").click();
     cy.location("pathname").should("eq", generateNewURL("/machines"));
-    cy.get(".p-navigation__link.is-selected a").should("not.exist");
+    cy.get(".p-navigation__link.is-selected a").contains("Machines");
```
Caleb-Ellis, line 22: I'm not sure how this ever worked before? If the url is at "/machines" then the highlighted nav link has to be "Machines"

huwshimi (approved):
Looks good.

PR #2934, opened 2021-08-06
feat(ui): update Placeholder to show real content when loaded

### LINKED ISSUE #2933: Update Placeholder component to be able to show text when loaded
Currently `Placeholder` is very simple and only shows a loading grey bar with a length proportional to its children. It would be nicer if it showed the real text after being loaded instead of having to have a separate component.

Caleb-Ellis:
## Done

- Updated `Placeholder` component to show real content when loaded

### MAAS deployment

- [Bolla](/HACKING.md#bolla)

### Running the branch

You can run this branch by:

### QA steps

- Open the clone form and select a source machine
- Check that the placeholder content shows while the machine details are loading
- Go to a PCI devices tab of a machine has them
- Check that the placeholder content shows while the PCI devices are loading

## Fixes

Fixes #2933

### DIFF
--- a/ui/src/app/base/components/Placeholder/Placeholder.test.tsx
+++ b/ui/src/app/base/components/Placeholder/Placeholder.test.tsx
@@ -15,4 +15,11 @@ describe("Placeholder", () => {
     const wrapper = shallow(<Placeholder>Placeholder text</Placeholder>);
     expect(wrapper).toMatchSnapshot();
   });
+
+  it("does not return placeholder styling if loading is false", () => {
+    const wrapper = shallow(
+      <Placeholder loading={false}>Placeholder text</Placeholder>
+    );
+    expect(wrapper.find("[data-test='placeholder']").exists()).toBe(false);
+  });
 });
--- a/ui/src/app/base/components/Placeholder/Placeholder.tsx
+++ b/ui/src/app/base/components/Placeholder/Placeholder.tsx
@@ -5,18 +5,27 @@ import classNames from "classnames";
 type Props = {
   children: ReactNode;
   className?: string;
+  loading?: boolean;
 };
 
-const Placeholder = ({ children, className }: Props): JSX.Element => {
+const Placeholder = ({
+  children,
+  className,
+  loading = true,
+}: Props): JSX.Element => {
   const delay = Math.floor(Math.random() * 750);
-  return (
-    <span
-      className={classNames("p-placeholder", className)}
-      style={{ animationDelay: `${delay}ms` }}
-    >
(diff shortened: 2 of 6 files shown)


Caleb-Ellis:
@huwshimi instead of also adding a `content` prop I've handled that outside the scope of the `Placeholder` component, otherwise it gets a bit messy having to check that `machine` exists in every `Placeholder`. I think this is the most flexible option that also means we don't have to have two separate lists.

huwshimi:
> @huwshimi instead of also adding a `content` prop I've handled that outside the scope of the `Placeholder` component, otherwise it gets a bit messy having to check that `machine` exists in every `Placeholder`. I think this is the most flexible option that also means we don't have to have two separate lists.

Wouldn't you only have to check for `loading` in the placeholder? So [this case](https://github.com/canonical-web-and-design/maas-ui/pull/2934/files#diff-75f19c37d1b430af0b17856e3aa34eb64ed7d6a3077ba8d9cfab1f6a35443680R24) would return `{content}` and the non loading state would return `{children}`? Maybe I misunderstand a bit about how this component works.

huwshimi (approved):
Looks much better I reckon, thanks!

Caleb-Ellis:
@huwshimi it's not so much about the component internals but rather the type errors when trying to access something in `children` that might not exist, e.g.
``` typescript
<Placeholder loading={!machine} content="Placeholder CPU model">
  {machine.cpu_model} // This will return `"cpu_model" does not exist on "MachineDetails | null"`
  {machine?.cpu_model} // This works, but it can get messy if the thing you're accessing isn't just a string
  {machine?.cpu_model || "Placeholder CPU model"} // This also works, and means the "content" prop is redundant
</Placeholder>
```
So to keep things simple I just kept `children` as the thing being rendered, and what it is is determined before it gets given to the `Placeholder` component

huwshimi:
> @huwshimi it's not so much about the component internals but rather the type errors when trying to access something in `children` that might not exist, e.g.
> 
> ```ts
> <Placeholder loading={!machine} content="Placeholder CPU model">
>   {machine.cpu_model} // This will return `"cpu_model" does not exist on "MachineDetails | null"`
>   {machine?.cpu_model} // This works, but it can get messy if the thing you're accessing isn't just a string
>   {machine?.cpu_model || "Placeholder CPU model"} // This also works, and means the "content" prop is redundant
> </Placeholder>
> ```
> 
> So to keep things simple I just kept `children` as the thing being rendered, and what it is is determined before it gets given to the `Placeholder` component

Ah yes I see, that's a shame.

PR #2942, opened 2021-08-09
fix(ui): infer correct prop types for formik fields

huwshimi:
## Done

- Update FormikField to infer correct prop types.
- Update type errors for components using FormikField.

## QA

N/A CI should pass.

Fixes: #2943.


Caleb-Ellis (approved):
LGTM, thanks!

PR #2944, opened 2021-08-09
fix(cypress): fix cookies being set in cypress tests

Caleb-Ellis:
## Done

- Properly set and clear cookies in Cypress tests. Originally it was only setting one of the cookies.
- Moved cookie-handling to custom hooks rather than in each individual component.

## QA
- N/A - Cypress tests should pass

### DIFF
--- a/integration/cypress/integration/login/login.spec.ts
+++ b/integration/cypress/integration/login/login.spec.ts
@@ -1,4 +1,6 @@
-import { generateLegacyURL, generateNewURL } from "@maas-ui/maas-ui-shared";
+import { generateNewURL } from "@maas-ui/maas-ui-shared";
+
+import { clearCookies } from "../utils";
 
 context("Login page", () => {
   beforeEach(() => {
@@ -6,8 +8,7 @@ context("Login page", () => {
   });
 
   afterEach(() => {
-    cy.clearCookie("skipintro");
-    cy.clearCookie("skipsetupintro");
+    clearCookies();
   });
 
   it("is disabled by default", () => {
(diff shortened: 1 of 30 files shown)


Caleb-Ellis:
@huwshimi so it turns out basically all the cypress tests have been giving a sort of false positive. One of the cookies was never set, but since intro used to be in legacy it would go:

1. Navigate to a React URL
2. Check that a thing exists
2a. It does - Cypress test passes and moves on
2b. But if it doesn't, half a second later the user redirects to the legacy intro

But now that intro is in React it doesn't have that lag time between the React app having loaded and the legacy app loading.

integration/cypress/integration/login/login.spec.ts
```diff
@@ -48,22 +49,18 @@ context("Login page", () => {
...
-    // Confirm skipping setup intro - should redirect to user intro.
-    cy.get("button[data-test='action-confirm']").click();
-    cy.location("pathname").should("eq", generateNewURL("/intro/user"));
-
     // Log out.
     cy.get(".p-navigation__link a:contains(Log out)").click();
 
+    // Set cookie to skip setup intro.
+    cy.setCookie("skipsetupintro", "true");
+
```
Caleb-Ellis, line 63: I've replaced this bit with just setting the cookie to keep state between tests as consistent as possible (maybe a fool's errand)

huwshimi (approved):
I think this is OK, it moves the overrides for the intros into once place (the selectors) so theoretically if we always use those selectors we shouldn't have to worry about the intros appearing when the cookies are set.

I do think it breaks the idea of selectors a bit though. You always expect the selector to return the data that's in the store and could be potentially very confusing if it's being manipulated by an outside source (i.e. the cookies).

At least this should only ever be a problem for the Cypress tests. Maybe we should add some big warnings to those selector descriptions.

ui/src/app/base/hooks.ts
```diff
@@ -472,3 +473,19 @@ export const useScrollOnRender = <T extends HTMLElement>(): ((
...
+  return !!completedIntro || !!getCookie("skipsetupintro");
+};
+
+/**
+ * Returns whether the user intro has been completed or skipped.
+ */
+export const useCompletedUserIntro = (): boolean => {
+  const completedUserIntro = useSelector(authSelectors.completedUserIntro);
+  return completedUserIntro || !!getCookie("skipintro");
+};
```
huwshimi, line 491: Nice, I think this works well now.

huwshimi (approved):
Thanks @Caleb-Ellis this looks good now.

PR #2962, opened 2021-08-17
refactor(ui): migrate TableHeader to TypeScript

huwshimi:
## Done

- migrate TableHeader to TypeScript

## QA

N/A

### DIFF
--- a/ui/src/app/base/types.ts
+++ b/ui/src/app/base/types.ts
@@ -1,13 +1,15 @@
+import type { ValueOf } from "@canonical/react-components";
+
 export type TSFixMe = any; // eslint-disable-line @typescript-eslint/no-explicit-any
 
-export enum SortDirection {
-  ASCENDING = "ascending",
-  DESCENDING = "descending",
-  NONE = "none",
-}
+export const SortDirection = {
+  ASCENDING: "ascending",
+  DESCENDING: "descending",
+  NONE: "none",
+} as const;
 
 export type Sort<K extends string | null = string> = {
-  direction: SortDirection;
+  direction: ValueOf<typeof SortDirection>;
   key: K | null;
 };
 
--- a/ui/src/app/base/types.ts
+++ b/ui/src/app/base/types.ts
@@ -1,8 +1,14 @@
 export type TSFixMe = any; // eslint-disable-line @typescript-eslint/no-explicit-any
 
-export type Sort = {
-  direction: "ascending" | "descending" | "none";
-  key: string;
+export enum SortDirection {
+  ASCENDING = "ascending",
+  DESCENDING = "descending",
+  NONE = "none",
+}
+
+export type Sort<K extends string | null = string> = {
+  direction: SortDirection;
+  key: K | null;
 };
 
 export type RouteParams = {
--- a/ui/src/app/base/components/TableHeader/TableHeader.test.tsx
+++ b/ui/src/app/base/components/TableHeader/TableHeader.test.tsx
@@ -2,6 +2,8 @@ import { shallow } from "enzyme";
 
 import TableHeader from "./TableHeader";
 
+import { SortDirection } from "app/base/types";
+
 describe("TableHeader ", () => {
   it("renders a div if no onClick prop is present", () => {
     const wrapper = shallow(<TableHeader>Text</TableHeader>);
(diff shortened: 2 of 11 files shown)


petermakowski (approved):
Nice. Awesome to see more TS adoption 🎉

ui/src/app/base/types.ts
```diff
@@ -1,8 +1,14 @@
 export type TSFixMe = any; // eslint-disable-line @typescript-eslint/no-explicit-any
 
-export type Sort = {
-  direction: "ascending" | "descending" | "none";
-  key: string;
+export enum SortDirection {
+  ASCENDING = "ascending",
+  DESCENDING = "descending",
+  NONE = "none",
+}
```
petermakowski, line 7: I may be missing some context here, what was your reasoning for choosing `enum`? Have you considered using an object with `as const`?

https://www.typescriptlang.org/docs/handbook/enums.html#objects-vs-enums

```suggestion
export const SortDirection = {
  ASCENDING: "ascending",
  DESCENDING: "descending",
  NONE: "none",
} as const
```

ui/src/app/base/types.ts
```diff
@@ -1,8 +1,14 @@
...
+
+export type Sort<K extends string | null = string> = {
+  direction: SortDirection;
```
petermakowski, line 10: As above, using an object with `as const` could result in better TS suggestions.
```suggestion
  direction: typeof SortDirection[keyof typeof SortDirection]
```
before
```
(property) direction: SortDirection
```
after
```
(property) direction: "ascending" | "descending" | "none"
```

ui/src/app/base/types.ts
Caleb-Ellis: I'm personally a fan of this method also - we've started using it in our [react-components library](https://github.com/canonical-web-and-design/react-components/blob/master/src/components/Notification/Notification.tsx#L10) in conjunction with a helper type [ValueOf](https://github.com/canonical-web-and-design/react-components/blob/master/src/types/index.ts#L41) (`typeof Foo[keyof typeof Foo]` is a bit of a mouthful).

Still, one of the benefits (and caveats) of using `enum`s like this though is that you *have* to use them - you can't just use the object values directly. So then there's literally only one place you ever have to write the underlying value.

huwshimi: @petermakowski this is a great question and like @Caleb-Ellis said we've started to adopt that pattern in some places.

I think we need to investigate it a bit more as it hasn't always worked brilliantly (i.e. having to add `as const` when you're using strings in some places).

I've updated this to use `as const`.

ui/src/app/base/types.ts
```diff
@@ -1,13 +1,15 @@
+import type { ValueOf } from "@canonical/react-components";
+
 export type TSFixMe = any; // eslint-disable-line @typescript-eslint/no-explicit-any
 
-export enum SortDirection {
-  ASCENDING = "ascending",
-  DESCENDING = "descending",
-  NONE = "none",
-}
+export const SortDirection = {
```
petermakowski, line 5: Just nitpicking here. Is there any reason why the constant name and its properties wouldn't follow the same naming convention (either all UPPERCASE or PascalCase)?

Seems like UPPERCASE has been already adopted in some places in this repo so might be worth following the same convention for consistency.
https://github.com/huwshimi/maas-ui/blob/ef53877bc99212d81693756fb75b78bf4e726051/ui/src/app/base/constants.ts#L11-L20

Otherwise, If you want to think of`SortDirection` as enum-like, might want to change it all to PascalCase.

> - Use PascalCase for enum names
> - Use PascalCase for enum member
https://basarat.gitbook.io/typescript/styleguide#enumhttps://basarat.gitbook.io/typescript/styleguide#enum

ui/src/app/base/types.ts
```diff
@@ -1,8 +1,16 @@
+import type { ValueOf } from "@canonical/react-components";
```
petermakowski, line 1: 👍  for using existing tools

petermakowski (approved):
Thanks for taking the time to make these changes. Added one more comment below, but it's nothing blocking.

ui/src/app/base/types.ts
huwshimi: There is a definitely a bit of inconsistency but we've mostly been treating enums like types (as they're somewhat interchangeable) so we've been using PascalCase for types and enums, but enum properties have been uppercase, a good example of this is this file:

https://github.com/canonical-web-and-design/maas-ui/blob/master/ui/src/app/store/types/enum.ts.

I suspect we've adopted this style mostly because we inherited the syntax from the legacy codebase, e.g. https://github.com/canonical-web-and-design/maas-ui/blob/master/legacy/src/app/enum.js.

We could definitely revisit the syntax, especially as there is already some inconsistency. That styleguide looks pretty handy, I haven't come across that before.

PR #2974, opened 2021-08-19
refactor(ui): migrate add machine and chassis forms to TypeScript

huwshimi:
## Done

- migrate add machine and chassis forms to TypeScript

## QA

N/A


Caleb-Ellis (approved):
Man it really was loosey goosey with variable types before wasn't it

huwshimi:
> Man it really was loosey goosey with variable types before wasn't it

haha yeah I'm definitely finding bugs as I go.

PR #2996, opened 2021-08-24
refactor(ui): set all errors to the api error type

huwshimi:
## Done

- Use the `APIErrors` type for all errors instead of `TSFixMe`.
- Update form components to handle event errors that have specific types.

## QA

N/A


Caleb-Ellis (approved):
LGTM!

PR #3000, opened 2021-08-25
refactor(ui): update tests to find components so that their props are inferred

huwshimi:
… inferred

## Done

- Update tests to find components so that their props are inferred.


Caleb-Ellis (approved):
oh nice, I didn't realise this was a thing!

PR #3011, opened 2021-08-26
chore(ui): turn on strict mode and remove betterer

huwshimi:
## Done

- remove betterer and all betterer ci task etc.
- turn on strict mode


phenomnomnominal:
Hey there! Sorry for the random comment, I was just doing some searching for Betterer stuff (it's my project!) and came across this PR. Looks like you successfully went from no strict mode, to strict mode enabled and that's really neat! If you have some time, I'd love to know more about how the migration went, if Betterer was actually useful, and what could've been better?

huwshimi:
> Hey there! Sorry for the random comment, I was just doing some searching for Betterer stuff (it's my project!) and came across this PR. Looks like you successfully went from no strict mode, to strict mode enabled and that's really neat! If you have some time, I'd love to know more about how the migration went, if Betterer was actually useful, and what could've been better?

Hi @phenomnomnominal thanks so much for your project, it did help us get to strict mode. Early on we didn't pay too much attention to the warnings, but as we got more familiar with TypeScript we started to take more notice.

I think without Betterer we might not really have thought too much about checking for strict mode errors and so when the time came the amount of work to enable strict mode was manageable. This project is in the order of 100k LOC so we were pretty happy to get the project to 100% strict TypeScript.

I'm not sure what could have been better(er). In some senses we had a bit of a love/hate relationship with Betterer because it was constantly yelling at us, but that's precisely what it was there for!

PR #3032, opened 2021-09-06
feat(ui): add ability to filter machines by clone error

Caleb-Ellis:
## Done

- Added filtering to clone results table

### MAAS deployment

- [Bolla](/HACKING.md#bolla)

### Running the branch

You can run this branch by:

### QA steps

- Go to the machine list, select some machines and select "Clone from" from the action menu
- Choose a source with incompatible networking/storage and submit the form
- Check that you can filter the machine list by clicking the "Show" link, and that opening the link in a new tab will also filter the machine list
- Check the same things from the KVM details page, using VMs
- Go to the details page of a machine and try to clone from an incompatible source machine
- Check that there is no column for "Affected machines" in the error table

## Fixes

Fixes canonical-web-and-design/app-squad#230

### DIFF
--- a/ui/src/app/machines/components/ActionFormWrapper/CloneForm/CloneResults/CloneResults.tsx
+++ b/ui/src/app/machines/components/ActionFormWrapper/CloneForm/CloneResults/CloneResults.tsx
@@ -11,9 +11,9 @@ import {
 } from "@canonical/react-components";
 import pluralize from "pluralize";
 import { useSelector } from "react-redux";
-import { Link } from "react-router-dom";
+import { Link, useLocation } from "react-router-dom";
 
-import type { APIError } from "app/base/types";
+import type { APIError, SetSearchFilter } from "app/base/types";
 import machineURLs from "app/machines/urls";
 import machineSelectors from "app/store/machine/selectors";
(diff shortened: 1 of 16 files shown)


ui/src/app/machines/components/ActionFormWrapper/CloneForm/CloneResults/CloneResults.tsx
```diff
@@ -171,6 +175,9 @@ export const CloneResults = ({
         },
         []
       ).length;
+  const viewingMachineDetails = pathname.startsWith(
```
huwshimi, line 178: I'm always a little worried that using the URL will result in a bug later on. What do you think about passing a prop down from MachineHeader?

huwshimi (approved):
Looks good, one question.

ui/src/app/machines/components/ActionFormWrapper/CloneForm/CloneResults/CloneResults.tsx
Caleb-Ellis: There’s a test that should fail if we ever change the machine details URL, but I think you’re right that it’s a bit brittle. I’ll add a ‘viewingDetails’ prop or something

PR #3050, opened 2021-09-14
feat(ui): add ability to generate client certificate when adding LXD VM host

Caleb-Ellis:
## Done

- Added `generateCertificate` actions/reducers/selectors
- Dispatch action to generate certificate in the Add LXD form and display in the authentication step

### MAAS deployment

- [Bolla](/HACKING.md#bolla)

### Running the branch

You can run this branch by:

### QA steps

- Go to the KVM list and click "Add KVM"
- Choose LXD, then input a name and address and make sure "Generate certificate" is selected
- Check that actions are dispatched to generate a certificate, and once loaded it shows up in the next step of the form

## Fixes

Fixes canonical-web-and-design/app-squad#257


ui/src/app/store/general/slice.ts
```diff
@@ -93,18 +70,79 @@ const generalSlice = createSlice({
...
-    ...generateGeneralReducers("bondOptions", "bond_options"),
-    ...generateGeneralReducers("componentsToDisable", "components_to_disable"),
-    ...generateGeneralReducers("defaultMinHweKernel", "default_min_hwe_kernel"),
-    ...generateGeneralReducers("hweKernels", "hwe_kernels"),
-    ...generateGeneralReducers("knownArchitectures", "known_architectures"),
-    ...generateGeneralReducers("machineActions", "machine_actions"),
-    ...generateGeneralReducers("osInfo", "osinfo"),
-    ...generateGeneralReducers("pocketsToDisable", "pockets_to_disable"),
-    ...generateGeneralReducers("powerTypes", "power_types"),
-    ...generateGeneralReducers("version", "version"),
```
Caleb-Ellis, line 106: I'll be honest - I'm completely stumped on how best to handle computed objects/object keys in TypeScript.

In order to get this to work and for TypeScript to be happy I've removed the larger `generateGeneralReducers()` helper util and replaced with four smaller utils, one for each reducer type. This way I can be explicit about exactly what the `reducers` object consists of. It' obviously led to a lot of repetition in the reducer list, but it's at least accurate whereas before it was basically evaluating as `{ [x: string] : ReducerType }`. Anyway I'm open to suggestions to make this better.

ui/src/app/kvm/components/KVMHeaderForms/AddKVM/AddLxd/AddLxd.tsx
```diff
@@ -48,21 +51,38 @@ export const AddLxd = ({
...
     }
-  }, [authenticated, step]);
+  }, [shouldGoToProjectStep]);
 
   // We run the cleanup function here rather than in each form component
   // because we want the errors to be able to persist across forms.
   useEffect(() => {
     return () => {
       dispatch(podActions.cleanup());
+      dispatch(generalActions.clearGeneratedCertificate());
```
huwshimi, line 85: It might be helpful to leave a comment as to why this is necessary. Something like: "The generated certificate state only holds one certificate at a time so the state can be cleaned up before fetching another certificate", but I'm sure you can make it make more sense than that :).

ui/src/app/kvm/components/KVMHeaderForms/AddKVM/AddLxd/AuthenticationForm/AuthenticationFormFields/AuthenticationFormFields.tsx
```diff
@@ -39,17 +42,16 @@ export const AuthenticationForm = ({
...
-              code: "certificate data",
-            },
-            { code: "EOF" },
-          ]}
-        />
+        <div className="certificate-wrapper">
+          <CodeSnippet
+            blocks={[
+              { code: "lxd config trust add - << EOF" },
+              { code: generatedCertificate?.certificate || "" },
```
huwshimi, line 49: Is this going to pop-in when it loads? Do we need to display something different if it doesn't exist?

ui/src/app/store/general/slice.ts
huwshimi: We've had this same issue with the other slices, so while not ideal this will have to do for now.

huwshimi (approved):
This looks good, a couple of tiny comments. Also I noticed that the certificate data includes `private_key` but it doesn't look like we have a need for it. Should we double check to see if it shouldn't be returned by the API?

ui/src/app/kvm/components/KVMHeaderForms/AddKVM/AddLxd/AuthenticationForm/AuthenticationFormFields/AuthenticationFormFields.tsx
Caleb-Ellis: It shouldn't. This step should only show if `generatedCertificate` exists, but the selector doesn't know that.

PR #3054, opened 2021-09-16
feat(ui): add ability to poll LXD server for authentication status

Caleb-Ellis:
## Done

- Added `pollLxdServer` to pod slice, which continuously checks authentication status with a LXD server
- Moved `setStep` functionality out of `AddLxd` and in to the individual step components. Having it all handled in the parent was a bit too inconsistent.

### MAAS deployment

- [Bolla](/HACKING.md#bolla)

### Running the branch

You can run this branch by:

### QA steps

- Go to the KVM list and click "Add KVM"
- Enter some details, make sure to keep "Generate new certificate" selected and click "Next"
- Make note of the certificate that was generated
- Keep "Add trust to LXD via command line" selected and click "Check authentication"
- Check that an action is dispatched to poll the server. If this is a new LXD server that MAAS doesn't know about, the poll action should be continuous until the code snippet is run on the server. Otherwise authentication should happen quickly.
- Select a project and click "Save"
- Check that the newly created LXD VM host's certificate is the same as what was generated before.

## Fixes

Fixes canonical-web-and-design/app-squad#253

## Screenshot
Peek 2021-09-16 13-04

### DIFF
--- a/ui/src/app/base/components/FormikFormContent/FormikFormContent.tsx
+++ b/ui/src/app/base/components/FormikFormContent/FormikFormContent.tsx
@@ -69,6 +69,7 @@ const generateNonFieldError = <V, E = null>(
 const FormikFormContent = <V, E = null>({
   allowAllEmpty,
   allowUnchanged,
+  cancelDisabled,
   children,
   className,
   cleanup,
@@ -161,7 +162,7 @@ const FormikFormContent = <V, E = null>({
       {editable && (
         <FormikFormButtons
           {...buttonsProps}
-          cancelDisabled={saving}
+          cancelDisabled={cancelDisabled === false ? false : saving}
           inline={inline}
           saved={saved}
           saving={saving}
--- a/ui/src/app/kvm/components/KVMHeaderForms/AddKVM/AddLxd/CredentialsForm/CredentialsForm.tsx
+++ b/ui/src/app/kvm/components/KVMHeaderForms/AddKVM/AddLxd/CredentialsForm/CredentialsForm.tsx
@@ -4,7 +4,12 @@ import { useDispatch, useSelector } from "react-redux";
 import * as Yup from "yup";
 
 import type { SetKvmType } from "../../AddKVM";
-import type { CredentialsFormValues, NewPodValues } from "../types";
+import { AddLxdSteps } from "../AddLxd";
+import type {
+  AddLxdStepValues,
+  CredentialsFormValues,
+  NewPodValues,
+} from "../types";
 
 import CredentialsFormFields from "./CredentialsFormFields";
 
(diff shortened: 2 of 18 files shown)


ui/src/app/base/components/FormikFormContent/FormikFormContent.tsx
```diff
@@ -161,7 +162,7 @@ const FormikFormContent = <V, E = null>({
       {editable && (
         <FormikFormButtons
           {...buttonsProps}
-          cancelDisabled={saving}
+          cancelDisabled={cancelDisabled === false ? false : saving}
```
Caleb-Ellis, line 165: I can't remember why we made `cancelDisabled` tied to `saving`? I've made it so that you can override that behaviour if you explicitly set `cancelDisabled` to false, but I'm wondering if we should remove the `saving` bit...

ui/src/app/store/pod/slice.ts
```diff
@@ -228,6 +228,71 @@ const podSlice = createSlice({
     ) => {
       state.errors = action.payload;
     },
+    pollLxdServer: {
```
Caleb-Ellis, line 231: This action uses `get_projects` as the websocket method, but I've split it into its own action instead of making `getProjects` more flexible. The separation is a bit cleaner and the intention makes a bit more sense, imo.

huwshimi: I think that makes sense.

huwshimi (approved):
A couple of comments/questions.

ui/src/app/base/components/FormikFormContent/FormikFormContent.tsx
huwshimi: Also, I meant to say this could probably do with a test.

Caleb-Ellis: Good call!

ui/src/app/kvm/components/KVMHeaderForms/AddKVM/AddLxd/CredentialsForm/CredentialsForm.tsx
```diff
@@ -43,6 +55,29 @@ export const CredentialsForm = ({
     }
   }, [errors]);
 
+  // Once a generated certificate for the new pod exists in state, the user
+  // should be sent to the auth trust step.
+  const shouldGoToAuthStep = !!generatedCertificate?.CN.includes(
```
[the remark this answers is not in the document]
Caleb-Ellis: Good call - yeah it should only match the first part. I'll add a util to split the certificate name

PR #3070, opened 2021-09-24
feat(ui): add custom certificate fields in LXD machine config

Caleb-Ellis:
## Done

- Created `CertificateDetails` component which shows read-only info of an existing certificate, used in both machine and KVM details
- Implemented custom certificate fields in LXD machine config - if the user is not editing the form they see the certificate data, if the user is editing the form they see the radios for generating/updating a certificate
- Changed `AuthenticationFields` a bit. I've renamed it to `CertificateFields` which I think makes a bit more sense, but also I've removed the password field entirely from it. This makes it so I can reuse the component in exactly the same way in 4 different places, otherwise there's a bit of bespoke field placement for each case:

**Add machine**

**Edit machine**

**Add KVM**

**Edit KVM**

- @amylily1011 the main consequence here is that in the "Add machine" form the password no longer shows below the "Generate new certificate" radio if its selected. Let me know if this is a deal breaker.

### QA steps

- Go to the config page of a LXD machine that was added manually, i.e. not composed from a VM host. Add a new one if it doesn't exist.
- Check that certificate data shows in the power section.
- Click "Edit" for the power config form and check that the certificate data changes into certificate fields.
- Select "Generate new certificate" and submit the form.
- Refresh the page (there's a bug in the backend where no machine update notify message is sent when updating machine power parameters).
- Check that the certificate updated successfully.
- Click "Edit" and try uploading a certificate/key manually.
- Refresh the page and check that the certificate updated successfully.
- Go to the config page of a LXD machine that was composed from a VM host. Compose a new one if it doesn't exist.
- Check that certificate data shows in the power section.
- Click "Edit" and check that you do not get the option to edit the certificate, since that is handled by the LXD VM host.

## Fixes

Fixes canonical-web-and-design/app-squad#247

## Launchpad issue

[lp#1944847](https://bugs.launchpad.net/maas/+bug/1944847)

### Added LXD machine
Peek 2021-09-24 18-11

### Composed LXD machine
Peek 2021-09-24 18-12


amylily1011:
@Caleb-Ellis that works, it's not a deal breaker. At first I thought it might be a little misleading if we don't want to encourage people to use the password but still have it there above the certificate, but considering that `adding machines` are only going to be a minor use case, I think we can do this way 😄

ui/src/app/base/components/CertificateDetails/CertificateDetails.tsx
```diff
@@ -0,0 +1,83 @@
...
+        <Textarea
+          className="certificate-details__textarea"
+          data-test="private-key"
+          id="lxd-key"
+          readOnly
+          rows={5}
+          value={privateKey}
+        />
+      )}
+      <Button data-test="toggle-key" onClick={() => setShowKey(!showKey)}>
```
huwshimi, line 67: I know this wasn't added in this PR, but this button needs `type="button"` as it currently has an error:

```
Warning: You submitted a Formik form using a button with an unspecified `type` attribute.  Most browsers default button elements to `type="submit"`. If this is not a submit button, please add `type="button"`.
```

ui/src/app/store/general/utils/certificates.ts
```diff
@@ -0,0 +1,16 @@
...
+): { name: string; host: string } | null => {
+  if (!certificateName) {
+    return null;
+  }
+  const split = certificateName.split("@");
+  if (split.length < 2) {
+    return { name: certificateName, host: "" };
+  }
+  const host = split[split.length - 1];
+  const name = split.slice(0, split.length - 1).join("@");
```
huwshimi, line 14: It might be good to leave a comment about what this is extracting/what the format is that the data is in.

huwshimi (approved):
Nice work.

ui/src/app/base/components/CertificateDetails/CertificateDetails.tsx
Caleb-Ellis: Thanks, done!

ui/src/app/store/general/utils/certificates.ts
Caleb-Ellis: Done!

PR #3220, opened 2021-11-03
fix(single-spa): fix for the ui not loading in Safari

huwshimi:
## Done

- Fix a regex that was causing the UI to not display in Safari.

### MAAS deployment

- [Bolla](/HACKING.md#bolla)

### Running the branch

You can run this branch by:

### QA steps

- Run the app with `yarn start` so that you are viewing the site with single-spa.
- Open the ui in Safari, it should load.
- Double check that it still works in Firefox/Chrome.

## Fixes

Fixes: #3045.


huwshimi:
@hatched maybe would you mind having a look at this as you've got Safari?

ui/src/single-spa-entry.js
```diff
@@ -27,10 +27,18 @@ const reactLifecycles = singleSpaReact({
 
 export const { bootstrap } = reactLifecycles;
 export const mount = (props) => {
-  // Get the full path, querystring and hash. The regex gets the first
-  // forward slash that is not a double forward slash and everything after it.
-  const matches = window.location.href.match(/(?<!\/)\/(?!\/).+/);
-  // The app should never reach this entrypoint witout the basename and react
+  // Get the full path, querystring and hash (location.pathname can't be used as
+  // it does not contain the querystring and hash).
+  const matches = window.location.href
```
hatched, line 32: You could make this a bit easier to understand for those that don't know regex by using a string replace to remove the protocol and hostname. 

```js
const loc = window.location;
const matches = loc.href.replace(`${loc.protocol}//${loc.host}`, '');
```
or you can reconstruct it
```js
const loc = window.location;
const matches = loc.pathname + loc.search + loc.hash;
```

hatched (approved):
Implementers choice :)

hatched:
QA OK in FF and Safari.

ui/src/single-spa-entry.js
huwshimi: I've done this and it's a lot cleaner, would you mind doing another review/QA?

hatched (approved):

hatched:
Looks good, thanks for the updates!

Issue #3197, opened 2021-10-27
Generate certificate errors not shown in add lxd host form


huwshimi:
When there is an error generating a certificate in the add lxd host form the error is displayed.

To reproduce:
- Go to /LXD.
- Click Add LXD host.
- Enter a really long name (e.g. "thishosthasalongnamethatisallonewordonereallyreallyreallylongword").
- Enter a valid address and click Next.
- You will stay on that form and if you look in the redux dev tools there will be an error like: `"[('asn1 encoding routines', 'ASN1_mbstring_ncopy', 'string too long')]"`

PR #3223, opened 2021-11-03
fix(ui): display cert and polling errors when adding an LXD KVM

huwshimi:
## Done

- Update the add LXD host wizard to display errors when generating a cert or authenticating with LXD.

### MAAS deployment

- [Bolla](/HACKING.md#bolla)

### Running the branch

You can run this branch by:

### QA steps

- Go to /LXD.
- Click Add LXD host.
- Enter a really long name (e.g. "thishosthasalongnamethatisallonewordonereallyreallyreallylongword").
- Enter a valid address and click "Next".
- An error should appear about the name being too long.
- Now enter a short name.
- Enter a valid address but change the port (e.g. "10.1.238.1:8442").
- Click "Next".
- Click "Check authentication".
- An error should appear about not connecting to LXD.
- Close the form and open it again.
- Enter a valid name and use the single-host address ("10.1.238.1:8443").
- Click "Next".
- Click "Check authentication".
- The new cert you generated won't have been added to the host so you should see the spinner, but no error should appear about the cert not being trusted (that error should be there in the pod.errors state).

## Fixes

Fixes: canonical-web-and-design/app-squad#483.
Fixes: canonical-web-and-design/maas-ui#3197.


ui/src/app/store/general/slice.ts
```diff
@@ -8,7 +8,7 @@ const generateInitialState = <K extends keyof GeneralState>(
   initialData: GeneralState[K]["data"]
 ) => ({
   data: initialData,
-  errors: {},
+  errors: null,
```
huwshimi, line 11: I changed this so that we don't get false positives that the errors exist (because an empty object is truthy - we've made this update for all other models). It doesn't appear that the general errors are being used anywhere (until this branch).

hatched:
If we know the name is going to be too long then why don't we provide immediate error feedback to the user in the UI?

hatched:
https://bugs.launchpad.net/maas/+bug/1949625 request to improve the messaging from maas.

amylily1011:
Thanks @hatched, this is good. Last time I check they didn't have any condition for the name field, so I was a little puzzled to see this error message. But thank you for filing the bug 😄

huwshimi:
> QA Issue:
> 
> * [ ]  The error persists after navigating away.
>   Follow steps 1-5 then navigate away and then back to adding a new LXD cluster.
> 

Thanks, that should be fixed now.

> 
> * [ ]  The error message is ugly can we display just the message content instead? We need to request better error messages from MAAS. "string too long"....which string? This should instead be something like "lxd cluster name must be between 1 and 18 characters". But I feel like we shouldn't let them submit values we know will be invalid.

I think this error is a bit of misdirection as it's a failure from generating the cert rather than a problem with the project name so I've replied to the Launchpad bug about the backend handling that without failing. I think for this branch we can consider it good if it exposes errors from the API (the name error being just one such failure). Does that sound reasonable?

ui/src/app/kvm/components/KVMHeaderForms/AddLxd/CredentialsForm/CredentialsForm.test.tsx
```diff
@@ -305,9 +306,50 @@ describe("CredentialsForm", () => {
...
+          />
+        </MemoryRouter>
+      </Provider>
+    );
+    expect(setStep).not.toHaveBeenCalled();
+    expect(wrapper.find(FormikForm).prop("errors")).toBe("name too long");
   });
 
-  it("clears the submission errors when unmounting", () => {
+  it.only("clears the submission errors when unmounting", () => {
```
Caleb-Ellis, line 352: 😲

Caleb-Ellis (approved):
LGTM! Just a lingering `only` in there

PR #3224, opened 2021-11-03
fix(ui): use resources endpoints for pod and vmcluster storage pools

Caleb-Ellis:
## Done

- Replaced the use of `pod.storage_pools` in UI components with `pod.resources.storage_pools`, which includes `allocated_other` data. The "other" data has been included in the KVM list in this PR.
- Removed the cluster storage pool selector, which was incorrectly aggregating each cluster member's pools. Instead it's been replaced with `vmcluster.total_resources.storage_pools`, which takes a similar (but not identical) shape to `pod.resources.storage_pools`.
- Updated components to be able to work with either pod or vmcluster resource.storage_pools
- Since the new data shape does not include the pod id, I've had to remove the "default" label and sorting for now. https://bugs.launchpad.net/maas/+bug/1949681

### MAAS deployment

- [Bolla](/HACKING.md#bolla)

### Running the branch

You can run this branch by:

### QA steps

- In the KVM list, hover over a host that uses "other" storage and check that it is reflected in the storage popover.
- Go to the details page of a LXD KVM that uses "other" storage.
- Check that the "Storage" section in the VMs card shows the correct data (for now, allocated is the sum of "tracked" and "other". Next step will be to reflect "other" data in the details pages separately).
- Click "Add virtual machine" and scroll down to the storage table
- Click the location dropdown and check that the "Allocated" amount is the sum of "tracked" and "other"
- Close the form and go to the "Resources" tab
- Check that the storage cards show "Allocated" as the sum of "tracked" and "other"

## Fixes

Fixes canonical-web-and-design/app-squad#487


Caleb-Ellis:
Note that this can't actually be QA'd properly until the bolla snap is updated, otherwise you'll need to build a MAAS from source and who wants to do that

Caleb-Ellis:
It appears the data we get in `pod.resources.storage_pools` is now the same shape as that in `vmcluster.total_resources.storage_pools` so we should just use that instead.

Caleb-Ellis:
Blocked until we get the cluster back on bolla

huwshimi (approved):
This looks good. I know this might be a little pedantic for some fairly simple calculations but I wonder if we should create utils (possibly as a followup) for calculating the allocated and free values for a pool as we're doing the same two calculations in lots of places and it might make it less error prone if we centralise those calculations.

Allocated: `pool.allocated_tracked + pool.allocated_other`.

Free: `pool.total - pool.allocated_tracked - pool.allocated_other`.

PR #3247, opened 2021-11-08
Update config to group minor dependency updates.

hatched:
## Done

Update the renovate config to mirror the config used in some other app projects in order to properly group minor dependencies. 

## QA

No QA necessary


renovate.json
```diff
@@ -1,25 +1,16 @@
...
-  "extends": [
-    "config:base"
-  ],
+  "extends": ["config:base", "group:allNonMajor", "schedule:weekly"],
   "labels": ["Maintenance 🔨"],
   "packageRules": [
     {
+      "groupName": "internal dependencies",
+      "groupSlug": "internal",
       "packagePatterns": [
```
huwshimi, line 8: I can't find `packagePatterns` in the docs, do you happen to know what this rule does?

renovate.json
```diff
@@ -1,25 +1,16 @@
...
+        "^@canonical",
+        "^canonicalwebteam",
+        "^vanilla-framework"
       ],
-      "excludePackagePatterns": [
-        "@canonical/react-components",
-        "jquery",
-        "prettier",
-        "vanilla-framework"
-      ],
```
huwshimi, line 17: I think we probably still want to exclude these as they've all caused breakages at minor versions.

renovate.json
huwshimi: It looks like it's been replaced by `matchPackagePatterns`, so would this only group the packages below?

huwshimi (changes requested):

renovate.json
hatched: I've updated this, we'll have to see if it still works after landing :) 

The previous approach is working on the other sites still so it might just be undocumented now. In any event if this works then we can make this change elsewhere.

renovate.json
hatched: done

huwshimi (approved):
Thanks @hatched, hopefully the "at any time" rule overrides the weekly rule. Like you say, we'll have to see how this works.

hatched:
What I've found is that it just creates a PR and constantly updates it as new updates come in and then it can be addressed at any time. But yeah we can tweak it over time if we find it's not fitting our needs.

PR #3251, opened 2021-11-09
feat(ui): only warn if the boot disk is below 8GB

huwshimi:
## Done

- Updated the compose form to only show a warning if the boot disk is below 8GB.

### MAAS deployment

- [Bolla](/HACKING.md#bolla)

### Running the branch

You can run this branch by:

### QA steps

- Go to a LXD cluster or single host and open the compose form.
- Scroll down to 'Storage configuration' and click 'Add disk'.
- Change the size on both disks to be less than 8GB.
- You should see size warning on the first disk but not the second.
- Change the boot radio button to the second disk.
- You should now see the size warning on the second disk but not the first.

## Fixes

Fixes: #3039.

## Screenshots

Screen Shot 2021-11-09 at 12 58 15 pm


Caleb-Ellis (approved):
LGTM!

PR #3256, opened 2021-11-09
ensure no-only-tests rule is enforced in all tests

petermakowski:
## Done

- ensure no-only-tests rule is enforced in all tests
  - add `no-only-tests` to `legacy`, `ui`, `cypress`
- setup `eslint` and `npmPkgJsonLint` for cypress

## QA
- verify lint errors occurred in the "demonstrate lint errors" commit: https://github.com/canonical-web-and-design/maas-ui/runs/4150402341?check_suite_focus=true
- locally: add `.only` to any `describe`/`it` block in a cypress/ui/legacy test and make sure running `yarn lint` reports an error


Caleb-Ellis (approved):
Awesome, thanks for this :+1:

PR #3322, opened 2021-11-24
feat(ui): add selectors for getting the active and selected devices

huwshimi:
## Done

- Add selectors for getting the active and selected devices.

## Fixes

Fixes: canonical-web-and-design/app-tribe#519.

### DIFF
--- a/ui/src/app/store/device/selectors.ts
+++ b/ui/src/app/store/device/selectors.ts
@@ -108,8 +108,8 @@ const eventErrorsForDevices = createSelector(
 
 /**
  * Returns currently active device's system_id.
- * @param {RootState} state - The redux state.
- * @returns {Device["system_id"]} Active device system_id.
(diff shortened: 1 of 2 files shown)


ui/src/app/store/device/selectors.ts
```diff
@@ -106,10 +106,62 @@ const eventErrorsForDevices = createSelector(
   }
 );
 
+/**
+ * Returns currently active device's system_id.
+ * @param {RootState} state - The redux state.
```
Caleb-Ellis, line 111: We don't need these type annotations anymore

Caleb-Ellis (approved):

ui/src/app/store/device/selectors.ts
huwshimi: Thanks, I'll remove them.

PR #3361, opened 2021-11-29
fix(ui): handle machine null status_message

petermakowski:
## Done
- fix Uncaught Type error for null machine `status_message` 
  - fix breakLines behaviour for null text value

## Before
https://pastebin.canonical.com/p/9XMPk5xQkP/
image

https://pastebin.canonical.com/p/tmF4SwWdbc/
image

## After
image

## QA
- if the machine on http://bolla.internal:5240/MAAS/r/ is still in the same state state that causes the UI to break - verify that the error no longer occurs
- otherwise, manually add/update a machine with the state below (`"status_message": null,`) to verify there's no runtime error

[Example machine that's causing the runtime error (text is null)](https://pastebin.canonical.com/p/GH56Vpk2jq/)

### DIFF
--- a/ui/src/app/utils/breakLines.ts
+++ b/ui/src/app/utils/breakLines.ts
@@ -3,12 +3,13 @@
  * that is inserted into a <pre> block.
  */
 export const breakLines = (
-  text: string | null,
+  text?: string | null,
   breakAtSpaces = true,
   lineLength = 52
 ): string => {
   let chunks = [];
 
+  // Check if text is `null` or `undefined`
   if (text == null) {
     return "";
   }
--- a/ui/src/app/utils/breakLines.ts
+++ b/ui/src/app/utils/breakLines.ts
@@ -8,9 +8,13 @@ export const breakLines = (
   lineLength = 52
 ): string => {
   let chunks = [];
+
+  if (text == null) {
+    return "";
+  }
   // Check that the text includes whitespace, otherwise we'll treat it as if we
   // are not breaking at spaces.
-  if (breakAtSpaces && text?.includes(" ")) {
+  if (breakAtSpaces && text.includes(" ")) {
     let remainingText = text.trim();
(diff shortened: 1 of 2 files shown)


ui/src/app/utils/breakLines.ts
```diff
@@ -3,14 +3,14 @@
...
-  text: string,
+  text: string | null,
   breakAtSpaces = true,
   lineLength = 52
 ): string => {
   let chunks = [];
   // Check that the text includes whitespace, otherwise we'll treat it as if we
   // are not breaking at spaces.
-  if (breakAtSpaces && text.includes(" ")) {
+  if (breakAtSpaces && text?.includes(" ")) {
```
hatched, line 13: I think it might be better to handle this failure case at the top of the fn and then treat it as a valid value for the rest of it...something like (untested):

```javascript
if (!text && text.length > 0) {
  return [];
}
```

then this line and the following one can assume that text is a string.

hatched (changes requested):

ui/src/app/utils/breakLines.ts
```diff
@@ -3,11 +3,15 @@
...
  */
 export const breakLines = (
-  text: string,
+  text: string | null,
   breakAtSpaces = true,
   lineLength = 52
 ): string => {
   let chunks = [];
+
+  if (text == null) {
```
hatched, line 12: === ;)

hatched (approved):

ui/src/app/utils/breakLines.ts
petermakowski: I specifically wanted to target both null and undefined in case this ever comes as undefined from the API. Can change to `===`.

petermakowski:
Opened another PR from my forked repo here https://github.com/canonical-web-and-design/maas-ui/pull/3363

PR #3370, opened 2021-12-01
refactor(ui): make the DHCP table work for all nodes

huwshimi:
## Done

- Update the DHCP table to handle all node types.

### MAAS deployment

- [Bolla](/HACKING.md#bolla)

### Running the branch

You can run this branch by:

### QA steps

- Go to settings and add a DHCP snippet to a machine.
- Go to the network tab for the machine you added the DHCP snippet to.
- You should see the DHCP table at the bottom.
- You should be able to edit the DHCP snippet.

## Fixes

Fixes: canonical-web-and-design/app-tribe#568.


ui/src/app/base/components/DHCPTable/DHCPTable.tsx
```diff
@@ -70,25 +69,23 @@ const generateRows = (
...
+        target: node.fqdn,
         type,
       },
     };
   });
 
 type Props = {
-  systemId: Machine["system_id"];
+  node: Node;
+  nodeType: NodeModels;
```
Caleb-Ellis, line 80: I'm not sure about using this type here. I think the fact that it's giving us the correct string at the moment is purely incidental, though this wouldn't be the case for e.g "pod". IMO it would be safer if we just used `string`, or when the API updates we use the `NodeType` enum that would then map to a display name (or something).

Caleb-Ellis (approved):
Just one comment on types

ui/src/app/base/components/DHCPTable/DHCPTable.tsx
huwshimi: Yeah, fair point. Thanks.

PR #3391, opened 2021-12-07
refactor(ui): improve action form composability

Caleb-Ellis:
## Done

- Removed some workarounds in `ActionForm` to get the machine batched actions to work. Now the form remains rendered if an action has started as opposed to being 100% locked to the processing machines in state. This change could also lead us to better handle errors returned from the API, like with the clone form that has a results page that lists the different errors for the different machines.
- Updated `useProcessing` hook to return completed state and added tests. Note we don't actually use the `onComplete` or `onError` functions in the hook anymore, but it could still be useful so I've kept them in.

### MAAS deployment

- [Bolla](/HACKING.md#bolla)

### Running the branch

You can run this branch by:

### QA steps

- I've set up a bunch of Manual machines in bolla so it might be easier to use bolla here
- Go to the machine list, select all the machines then select an action that can't be performed by all of them
- Check that a warning shows and you can filter the machines to only those that can perform the action
- Unselect the machines and select a bunch of "New" machines
- Commission them and check that the warning doesn't show up at any point
- With the same machines selected, mark them broken while they're commissioning
- Now, try and mark them fixed
- Check that an error shows and the form does not close

## Fixes

Fixes canonical-web-and-design/app-tribe#582

### DIFF
--- a/ui/src/app/base/hooks/base.ts
+++ b/ui/src/app/base/hooks/base.ts
@@ -87,32 +87,38 @@ export const useCycled = (
  * @param onComplete - The function to call when all the items have been processed.
  * @param hasErrors - Whether there are any item errors in state.
  * @param onError - The function to call when an error occurs.
+ * @returns Whether the list of items has been processed successfully.
  */
-export const useProcessing = (
-  processingCount: number,
-  onComplete: () => void,
+export const useProcessing = ({
   hasErrors = false,
-  onError: () => void
-): void => {
-  const processingStarted = useRef(false);
-  if (processingStarted.current === false && processingCount > 0) {
-    processingStarted.current = true;
-  }
+  onComplete,
+  onError,
+  processingCount,
+}: {
+  hasErrors?: boolean;
+  onComplete?: () => void;
+  onError?: () => void;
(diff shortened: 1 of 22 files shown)


ui/src/app/base/hooks/base.ts
```diff
@@ -87,32 +87,38 @@ export const useCycled = (
...
+export const useProcessing = ({
   hasErrors = false,
-  onError: () => void
-): void => {
-  const processingStarted = useRef(false);
-  if (processingStarted.current === false && processingCount > 0) {
-    processingStarted.current = true;
-  }
+  onComplete = () => null,
+  onError = () => null,
```
huwshimi, line 95: I think maybe we shouldn't set default functions here. It might be easier to understand that these are optional if we check for the existence of the function when we call them below.

huwshimi (approved):
Looks good. One minor comment.

ui/src/app/base/hooks/base.ts
Caleb-Ellis: Fair enough! Updated

PR #3399, opened 2021-12-08
refactor(ui): update networking utils to work with devices

huwshimi:
## Done

- Update the networking utils to work with all node types.

## Fixes

Fixes: canonical-web-and-design/app-tribe#538.


ui/src/app/store/utils/node/base.ts
```diff
@@ -1,7 +1,11 @@
 import type { Controller } from "app/store/controller/types";
+import { isControllerDetails } from "app/store/controller/utils";
 import type { Device } from "app/store/device/types";
+import { isDeviceDetails } from "app/store/device/utils";
 import type { Machine } from "app/store/machine/types";
-import type { Node } from "app/store/types/node";
+// Import from the common utils to prevent an import loop in machine/utils/index.ts.
+import { isMachineDetails } from "app/store/machine/utils/common";
```
Caleb-Ellis, line 7: Is this necessary? As far as I can tell all the node utils are an extension of the individual machine/device/controller utils - I'm not sure where the import loop would happen?

Caleb-Ellis (approved):
LGTM just one question on an import

ui/src/app/store/utils/node/base.ts
huwshimi: What's happening is that we have a [large index file](https://github.com/canonical-web-and-design/maas-ui/blob/67e27f9e251186e283a9417d89da7de425e9c445/ui/src/app/store/utils/index.ts) that exports all the base utils and then we also have a [large index file](https://github.com/canonical-web-and-design/maas-ui/blob/67e27f9e251186e283a9417d89da7de425e9c445/ui/src/app/store/machine/utils/index.ts) that exports all the machine utils.

I introduced the import into the node utils: `import { isMachineDetails } from "app/store/machine/utils";`, but inside the machine utils we have imports like `import { getLinkInterface, hasInterfaceType } from "app/store/utils";` so because those files are both exported from the index files they end up importing from each other (when compiled).

When I try to run the code I get errors like "TypeError: Cannot read property 'generateBaseSelectors' of undefined" and the whole thing collapses.

Or at least that's my understanding of what's going on. Open to suggestions if you think there might be another solution other than changing the import.

PR #3400, opened 2021-12-08
remove redundant clearCookies in afterEach hook

petermakowski:
## Done

- remove redundant `clearCookies` in `afterEach` hook
  - initally I planned to move state reset from `afterEach` to `beforeEach` hook to follow best practices:
    **[Resetting the state in afterEach hook is an anti-pattern](https://docs.cypress.io/guides/references/best-practices#Using-after-or-afterEach-hooks)**
  - [cypress clears cookies automatically before each test](https://docs.cypress.io/api/commands/clearcookies) - so there's no need to do this manually


hatched (approved):
Looks good, thanks!

PR #3412, opened 2021-12-09
refactor login to custom cypress command

petermakowski:
## Done

- refactor login to custom cypress command
  - move `integration/cypress/support/index` to TypeScript and add type declarations

Fixes #593


integration/cypress/support/commands.ts
```diff
@@ -0,0 +1,18 @@
+import { BASENAME } from "@maas-ui/maas-ui-shared";
+
+Cypress.Commands.add(
+  "login",
+  (username = Cypress.env("username"), password = Cypress.env("password")) => {
```
hatched, line 5: Should this be an async function? Is there any risk that the "post" doesn't complete before the function ends and the test continues on?

petermakowski: Because Cypress commands are already promise-like, you don't need to wrap them or return your own promise.

Cypress will resolve your command with whatever the final Cypress command yields.

This will actually throw an error if you try to do this.

huwshimi (approved):
Nice, thanks for the cleanup.

PR #3423, opened 2021-12-15
feat(ui): edit device interface form

huwshimi:
## Done

- Add a form for editing a device's physical interfaces.

### MAAS deployment

- [Bolla](/HACKING.md#bolla)

### Running the branch

You can run this branch by:

### QA steps

- Go to /r/devices.
- Click on a device and then go to the network tab.
- Open the action menu for one of the devices and click "Edit physical".
- Make some changes and save.
- Your changes should get reflected in the table.

## Fixes

Fixes: canonical-web-and-design/app-tribe#572.

### DIFF
--- a/ui/src/app/devices/views/DeviceDetails/DeviceNetwork/EditInterface/EditInterface.test.tsx
+++ b/ui/src/app/devices/views/DeviceDetails/DeviceNetwork/EditInterface/EditInterface.test.tsx
@@ -1,0 +1,229 @@
+import { mount } from "enzyme";
+import { Provider } from "react-redux";
+import configureStore from "redux-mock-store";
+
+import EditInterface from "./EditInterface";
+
+import { actions as deviceActions } from "app/store/device";
+import deviceSelectors from "app/store/device/selectors";
+import type { DeviceNetworkInterface } from "app/store/device/types";
+import { DeviceIpAssignment } from "app/store/device/types";
+import type { RootState } from "app/store/root/types";
+import {
+  device as deviceFactory,
+  deviceDetails as deviceDetailsFactory,
+  deviceEventError as deviceEventErrorFactory,
+  deviceInterface as deviceInterfaceFactory,
+  deviceState as deviceStateFactory,
+  deviceStatus as deviceStatusFactory,
+  deviceStatuses as deviceStatusesFactory,
+  rootState as rootStateFactory,
+  subnet as subnetFactory,
+  subnetState as subnetStateFactory,
+} from "testing/factories";
+import { submitFormikForm } from "testing/utils";
+
+const mockStore = configureStore();
+
+describe("EditInterface", () => {
+  let state: RootState;
+  let nic: DeviceNetworkInterface;
+  beforeEach(() => {
+    nic = deviceInterfaceFactory();
+    state = rootStateFactory({
+      device: deviceStateFactory({
+        items: [
+          deviceDetailsFactory({
+            interfaces: [nic],
+            system_id: "abc123",
+          }),
+        ],
+        loaded: true,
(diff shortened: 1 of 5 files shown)


ui/src/app/devices/views/DeviceDetails/DeviceNetwork/EditInterface/EditInterface.test.tsx
```diff
@@ -0,0 +1,229 @@
...
+    state.device.items[0] = deviceFactory({ system_id: "abc123" });
+    const store = mockStore(state);
+    const wrapper = mount(
+      <Provider store={store}>
+        <EditInterface nicId={nic.id} systemId="abc123" closeForm={jest.fn()} />
+      </Provider>
+    );
+    expect(
+      wrapper.find("[data-testid='loading-device-details']").exists()
+    ).toBe(true);
```
petermakowski, line 67: I find this test misleading. It implies that an element with a`data-testid='loading-device-details` exists.

In reality, it will never appear anywhere in the DOM - [Spinner doesn't accept the `data-testid` prop](https://github.com/canonical-web-and-design/react-components/blob/main/src/components/Spinner/Spinner.tsx#L21) (https://github.com/canonical-web-and-design/react-components/issues/651). 

I'd suggest to check for Spinner component instance, like it's often done in enzyme-style unit tests, or use react testing library to assert the real DOM (I believe unit tests as such provide less value than BDD-style integration tests, but you already know that 😉 ).

ui/src/app/devices/views/DeviceDetails/DeviceNetwork/EditInterface/EditInterface.test.tsx
```diff
@@ -0,0 +1,229 @@
...
+    const wrapper = mount(<Proxy />);
+    const formValues = {
+      ip_address: "192.168.1.1",
+      ip_assignment: DeviceIpAssignment.STATIC,
+      mac_address: "11:22:33:44:55:66",
+      name: "eth123",
+      subnet: 2,
+      tags: ["tag1", "tag2"],
+    };
+    submitFormikForm(wrapper, formValues);
```
petermakowski, line 195: Did you consider testing the edit interface in the way the user would use it?

It would give more confidence the UI works as expected and have the advantage of providing a good overview of how the component is expected to behave at a glance, whilst this is arguably quite difficult to read.

We have no way of knowing if the form actually works correctly or if it's accessible other that testing it manually  - and there's no end-to-end test.

Correct me if I'm wrong, but looks like the form submission on the `EditInterface` component could be broken for the user and the test would still pass (you're calling the `onSubmit` handler directly, and not testing the actual UI anywhere).

huwshimi: We could consider adding some integration tests for the forms in the future, but this is a unit test for the surface area that this component interacts with without needing to test the internals of the components it wraps. We've found this approach to be a reliable way to have confidence in the component and has a very low maintenance burden.

A small number of integration tests could help to raise that confidence but with the move to TypeScript it is hard to mis-use these components by e.g. passing the wrong type of data.

ui/src/app/devices/views/DeviceDetails/DeviceNetwork/EditInterface/EditInterface.test.tsx
huwshimi: Thanks, I've added the reference to the Spinner component, but I think the test id is still worthwhile here as we're not doing a shallow compare so there could be other Spinners that could provide false positives.

Caleb-Ellis (approved):
LGTM!

PR #3424, opened 2021-12-15
test: add cypress-axe accessibility tests

petermakowski:
## Done
- add accessibility tests with cypress-axe 
  - skip failures by default (can be set to false once fixed)
  - extract and log useful information from each failure (violation level, summary, link with info how to fix, and HTML node that caused the failure):
  image

- add [cypress-wait-until](https://www.npmjs.com/package/cypress-wait-until) to ensure loading UI elements disappear before continuing
- add `data-testid="section-header-title"` to login, intro, network list titles so that they can be targeted consistently with other page titles (this would ideally be refactored so that they're all `h1`)
  

## Fixes

Fixes: #592 .

### DIFF
--- a/integration/cypress/integration/accessibility/accessibility.spec.ts
+++ b/integration/cypress/integration/accessibility/accessibility.spec.ts
@@ -33,6 +33,9 @@ pages.forEach(({ heading, url }: Page) => {
     cy.visit(pageUrl);
 
     cy.waitUntil(
+      // using `Cypress.$` instead of `cy.get` to prevent test failure
+      // - `cy.get` itself is an assertion and would throw an error when element is not found
+      // https://github.com/NoriSte/cypress-wait-until/issues/75#issuecomment-572685623
       () => Cypress.$("[data-testid='section-header-title-spinner']").length < 1
     );
 
--- a/integration/cypress/integration/accessibility/accessibility.spec.ts
+++ b/integration/cypress/integration/accessibility/accessibility.spec.ts
@@ -2,38 +2,49 @@ import { generateLegacyURL, generateNewURL } from "@maas-ui/maas-ui-shared";
 
 const legacyPages = ["/devices", "/controllers", "/networks"];
 
-const newPages = [
-  { title: "Devices", url: "/devices" },
-  { title: "Controllers", url: "/controllers" },
-  { title: "Subnets", url: "/networks" },
-  { title: "Machines", url: "/machines" },
-  { title: "KVM", url: "/kvm/lxd" },
-  { title: "Images", url: "/images" },
-  { title: "DNS", url: "/domains" },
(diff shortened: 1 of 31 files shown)


integration/cypress/integration/accessibility/accessibility.spec.ts
```diff
@@ -0,0 +1,51 @@
+import { generateLegacyURL, generateNewURL } from "@maas-ui/maas-ui-shared";
+
+const legacyPages = ["/devices", "/controllers", "/networks"];
+
+type Page = { heading: string; url: string };
+const pages: Page[] = [
+  { heading: "Login", url: "/accounts/login" },
+  { heading: "SSH keys for admin", url: "/intro/user" },
```
Caleb-Ellis, line 8: We should include the MAAS intro page at `/intro` with the heading "Welcome to MAAS" that is only accessible if the MAAS hasn't been set up for the first time (or when `shouldSkipIntro` is false

integration/cypress/support/commands.ts
```diff
@@ -1,18 +1,83 @@
...
+    url: `${BASENAME}/accounts/login/`,
+    form: true,
+    body: {
+      username,
+      password,
+    },
+  });
+});
+
+type Violation = {
```
Caleb-Ellis, line 30: We should use the type exposed by `axe-core`:
``` ts
import type { Result } from "axe-core";`
```

integration/cypress/support/index.ts
```diff
@@ -1,12 +1,22 @@
 /// <reference types="cypress" />
 
+import type { Options } from "cypress-axe";
 import "cypress-axe";
+import type { ContextObject } from "axe-core";
```
Caleb-Ellis, line 5: Neither `Options` nor `ContextObject` are used in this file so should be removed.

Caleb-Ellis (approved):
LGTM! Just some minor suggestions on types and including the MAAS initial intro page

integration/cypress/integration/accessibility/accessibility.spec.ts
```diff
@@ -0,0 +1,51 @@
+import { generateLegacyURL, generateNewURL } from "@maas-ui/maas-ui-shared";
+
+const legacyPages = ["/devices", "/controllers", "/networks"];
```
Caleb-Ellis, line 3: We can move `/devices` out of legacy pages now that we're using the React version

integration/cypress/integration/accessibility/accessibility.spec.ts
petermakowski: Seems like `shouldSkipIntro = false` isn't enough, this would require running this test without any users (like we do in `no-users` directory). Doesn't seem worth the effort, I'd rather skip this for now if that's okay with you?

Issue #615, opened 2020-01-06
Current nav item not selected for legacy


huwshimi:
**Describe the bug**
In the angular app the current page is not selected in the top nav.

PR #3483, opened 2022-01-10
refactor(ui): setup Subnets React page

petermakowski:
## Done
Subnets React page setup
- add cypress tests for the main subnets view
  - based on user journeys descriptions - more in [MAAS Subnets React migration plan - User Journeys](https://docs.google.com/document/d/1Vlih_9RyfNKKwLeBXS6tnRnNZTTR1BtXG_MHHC-Bi1s/edit#heading=h.h5ndvhge2k7u)
- fix "Group by" label (`cypress-testing-library` reported this,  HTML`for` attribute was pointing to the `name` attribute instead of the `id`) - a good example how user-centric testing (and `byRole` queries) help identify UI bugs
- add "Subnets" aria-label to table - describes it for screen readers and allowis to target specifically in tests

https://docs.google.com/document/d/1Vlih_9RyfNKKwLeBXS6tnRnNZTTR1BtXG_MHHC-Bi1s/edit#heading=h.faeke7zfhcfj

### QA steps

- Manually enter `/MAAS/r/networks` URL and verify "Subnets" h1 element is displayed

## Screenshots
image
*New Subnets page*

### Before
Pasted Graphic
*no label for "group by" combobox*

combobox
*cypress error*

### After
Pasted Graphic 1
*correct label*

margin--bottom ng-pristine ng-untouched ng-valid ng-not-
*cypress success*

## Cypress tests
https://user-images.githubusercontent.com/7452681/148773257-0c9f6e9d-f600-4980-bc5b-9ab03e83c6e1.mp4

## Fixes

Fixes: [#615 .](https://github.com/canonical-web-and-design/app-tribe/issues/615)

### DIFF
--- a/integration/cypress/integration/legacy/subnets.spec.ts
+++ b/integration/cypress/integration/legacy/subnets.spec.ts
@@ -35,18 +35,25 @@ context("Subnets", () => {
       "Available IPs",
       "Space",
     ];
-    expectedHeaders.forEach((name) => {
-      cy.findByRole("columnheader", { name }).should("exist");
+    cy.findByRole("table", { name: "Subnets" }).within(() => {
+      expectedHeaders.forEach((name) => {
+        cy.findByRole("columnheader", { name }).should("exist");
+      });
     });
   });
 
   it("allows filtering by 'fabrics' or 'spaces' via the 'Group by' drop-down", () => {
-    cy.findAllByRole("columnheader").first().should("have.text", "Fabric");
-    cy.findByRole("combobox", { name: "Group by" }).within(() => {
-      cy.findByRole("option", { selected: true }).contains("Fabrics");
-      cy.findByRole("option", { name: "Spaces" }).should("exist");
+    cy.findByRole("table", { name: "Subnets" }).within(() => {
+      cy.findAllByRole("columnheader").first().should("have.text", "Fabric");
     });
-    cy.findByRole("combobox", { name: "Group by" }).select("Spaces");
-    cy.findAllByRole("columnheader").first().should("have.text", "Space");
+    cy.findByRole("combobox", { name: "Group by" })
+      .within(() => {
+        cy.findByRole("option", { selected: true }).contains("Fabrics");
+        cy.findByRole("option", { name: "Spaces" }).should("exist");
+      })
+      .select("Spaces");
+    cy.findByRole("table", { name: "Subnets" }).within(() => {
+      cy.findAllByRole("columnheader").first().should("have.text", "Space");
+    });
   });
 });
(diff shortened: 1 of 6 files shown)


ui/src/app/subnets/views/SubnetsList.tsx
```diff
@@ -0,0 +1,13 @@
+import React from "react";
+
+const SubnetsList = (): React.ReactElement => (
+  <header className="p-strip--light is-shallow u-no-padding--bottom">
+    <div className="row">
+      <div className="col-medium-4 col-8">
+        <h1 data-testid="section-header-title">Subnets</h1>
+      </div>
+    </div>
+  </header>
```
petermakowski, line 10: This will be replaced with a correctly formatted header in a next PR

integration/cypress/integration/legacy/subnets.spec.ts
```diff
@@ -16,5 +17,43 @@ context("Subnets", () => {
...
+      "Space",
+    ];
+    cy.findByRole("table", { name: "Subnets" }).within(() => {
+      expectedHeaders.forEach((name) => {
+        cy.findByRole("columnheader", { name }).should("exist");
+      });
+    });
+  });
+
+  it("allows filtering by 'fabrics' or 'spaces' via the 'Group by' drop-down", () => {
```
Caleb-Ellis, line 45: It seems like an odd approach to write tests for the legacy version only to have to port them over to the React version soon afterwards. It also sets a precedent of using the legacy app as a goalpost, when we really are trying to replace it wholesale - we use it as a starting point, but we try not to replicate things that are obviously broken or have some awkward UX that can be easily fixed.

It obviously doesn't hurt to include these tests now even if they do get replaced and changed a bit, but I'm curious of the reasoning.

Caleb-Ellis (approved):
Just a question on the cypress tests, but otherwise LGTM!

integration/cypress/integration/legacy/subnets.spec.ts
petermakowski: I cannot imagine having no end to end coverage for a feature that I'm working on therefore I would write them anyway - but doing this first gives an ability to almost do TDD approach moving forward - writing the new implementation and watching the tests go green as you go.

> It seems like an odd approach to write tests for the legacy version only to have to port them over to the React version soon afterwards.

The way I think about it is - I didn't write tests for the _legacy version_. The goal was to write tests for critical user journeys, which will remain the same after migration or even some redesigns (e.g. link inside of a primary navigation will still be a link inside of primary navigation even after we move to using [application panels](https://vanillaframework.io/docs/layouts/application#panels-example) in the future, which I know is planned).

>  we try not to replicate things that are obviously broken or have some awkward UX that can be easily fixed.

That's also the reason I decided to write tests first - user-centric testing approach is excellent in identifying accessibility issues/HTML that has questionable structure.

> It obviously doesn't hurt to include these tests now even if they do get replaced and changed a bit, but I'm curious of the reasoning.

I didn't want to throw away everything but analyse the current implementation and keep the good stuff (if it ain't broke, don't fix it).

This is a side-effect of me doing preparatory work - going through code and UI, analysing the feature and writing them to serve as a documentation as I go. Just a part of my regular process.

PR #3492, opened 2022-01-11
feat(ui): add controller instructions

huwshimi:
## Done

- Add instructions for adding a controller.

### MAAS deployment

- [Bolla](/HACKING.md#bolla)

### Running the branch

You can run this branch by:

### QA steps

- Visit the controllers list.
- Click on "Add controller".
- You should see instructions on how to add a controller.
- Click "Close" and the header should close.

## Fixes

Fixes: canonical-web-and-design/app-tribe#612.

### DIFF
--- a/ui/src/app/controllers/components/ControllerHeaderForms/AddController/AddController.test.tsx
+++ b/ui/src/app/controllers/components/ControllerHeaderForms/AddController/AddController.test.tsx
@@ -1,0 +1,58 @@
+import { mount } from "enzyme";
+import { Provider } from "react-redux";
+import { MemoryRouter } from "react-router-dom";
+import configureStore from "redux-mock-store";
+
+import AddController from "./AddController";
+
+import type { RootState } from "app/store/root/types";
+import {
+  configState as configStateFactory,
+  rootState as rootStateFactory,
+} from "testing/factories";
+
+const mockStore = configureStore();
+
+describe("AddController", () => {
+  let state: RootState;
+
+  beforeEach(() => {
+    state = rootStateFactory({
+      config: configStateFactory({
+        items: [
+          { name: "maas_url", value: "http://1.2.3.4/MAAS" },
+          { name: "rpc_shared_secret", value: "veryverysecret" },
+        ],
+      }),
+    });
+  });
+
+  it("includes the config in the instructions", () => {
+    const store = mockStore(state);
+    const wrapper = mount(
+      <Provider store={store}>
+        <MemoryRouter>
+          <AddController clearHeaderContent={jest.fn()} />
+        </MemoryRouter>
+      </Provider>
+    );
+    const instructions = wrapper.find("[data-testid='instructions']").text();
+    expect(instructions.includes("http://1.2.3.4/MAAS")).toBe(true);
+    expect(instructions.includes("veryverysecret")).toBe(true);
+  });
+
+  it("can close the instructions", () => {
+    state.zone.loaded = false;
+    const store = mockStore(state);
+    const clearHeaderContent = jest.fn();
+    const wrapper = mount(
+      <Provider store={store}>
+        <MemoryRouter>
+          <AddController clearHeaderContent={clearHeaderContent} />
+        </MemoryRouter>
+      </Provider>
+    );
+    wrapper.find("Button[data-testid='close']").simulate("click");
+    expect(clearHeaderContent).toHaveBeenCalled();
+  });
+});
--- a/ui/src/app/controllers/components/ControllerHeaderForms/AddController/AddController.test.tsx
+++ b/ui/src/app/controllers/components/ControllerHeaderForms/AddController/AddController.test.tsx
@@ -36,13 +36,14 @@ describe("AddController", () => {
         </MemoryRouter>
       </Provider>
     );
-    const instructions = wrapper.find("[data-testid='instructions']").text();
+    const instructions = wrapper
+      .find("[data-testid='register-snippet']")
+      .text();
     expect(instructions.includes("http://1.2.3.4/MAAS")).toBe(true);
     expect(instructions.includes("veryverysecret")).toBe(true);
   });
 
   it("can close the instructions", () => {
-    state.zone.loaded = false;
     const store = mockStore(state);
     const clearHeaderContent = jest.fn();
     const wrapper = mount(
@@ -52,7 +53,9 @@ describe("AddController", () => {
         </MemoryRouter>
       </Provider>
     );
-    wrapper.find("Button[data-testid='close']").simulate("click");
+    wrapper
+      .find("Button[data-testid='add-controller-close']")
+      .simulate("click");
     expect(clearHeaderContent).toHaveBeenCalled();
   });
 });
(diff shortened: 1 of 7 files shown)


ui/src/app/controllers/components/ControllerHeaderForms/AddController/AddController.test.tsx
```diff
@@ -0,0 +1,58 @@
...
+  beforeEach(() => {
+    state = rootStateFactory({
+      config: configStateFactory({
+        items: [
+          { name: "maas_url", value: "http://1.2.3.4/MAAS" },
+          { name: "rpc_shared_secret", value: "veryverysecret" },
+        ],
+      }),
+    });
+  });
```
petermakowski, line 28: This is a minor thing, but I noticed you often use this pattern. It seems unnecessary (unless there's something that I'm missing, please explain).

You could get rid of reassignement of state in `beforeEach` and nesting all tests in `describe` by returning a new state object in each test (like you do with `store`). This would simplify this test and make easier to reason about.

https://kentcdodds.com/blog/avoid-nesting-when-youre-testing

ui/src/app/controllers/components/ControllerHeaderForms/AddController/AddController.test.tsx
```diff
@@ -0,0 +1,58 @@
...
+    const store = mockStore(state);
+    const clearHeaderContent = jest.fn();
+    const wrapper = mount(
+      <Provider store={store}>
+        <MemoryRouter>
+          <AddController clearHeaderContent={clearHeaderContent} />
+        </MemoryRouter>
+      </Provider>
+    );
+    wrapper.find("Button[data-testid='close']").simulate("click");
```
petermakowski, line 55: Doesn't using a generic string ("close") defeat the purpose of using a testid?

This could be re-written with react testing library: `userEvent("click", getByRole("button", { name: "Close" }))`. It would the advantage of testing actual DOM nodes that users interact with instead of enzyme wrappers.

ui/src/app/controllers/components/ControllerHeaderForms/AddController/AddController.test.tsx
```diff
@@ -0,0 +1,58 @@
...
+      <Provider store={store}>
+        <MemoryRouter>
+          <AddController clearHeaderContent={jest.fn()} />
+        </MemoryRouter>
+      </Provider>
+    );
+    const instructions = wrapper.find("[data-testid='instructions']").text();
+    expect(instructions.includes("http://1.2.3.4/MAAS")).toBe(true);
+    expect(instructions.includes("veryverysecret")).toBe(true);
+  });
```
petermakowski, line 42: 👍 for going in depth and testing config strings specifically

ui/src/app/controllers/components/ControllerHeaderForms/AddController/AddController.tsx
```diff
@@ -0,0 +1,47 @@
...
+sudo maas init rack --maas-url ${maasUrl} --secret ${rpcSharedSecret}`}</code>
+      </pre>
+      <Row>
+        <Col size={6}>
+          <a href="https://maas.io/docs/rack-controller">
+            Help with adding a rack controller
+          </a>
+        </Col>
+        <Col className="u-align--right" size={6}>
+          <Button data-testid="close" onClick={clearHeaderContent}>
```
petermakowski, line 38: As above, please remove this testid and target a button with text in tests or rename to be more explicit.

petermakowski (changes requested):
Looks good, but I have a few comments on the approach to testing. Could you please address the testid issue that I explain in more details below.

ui/src/app/controllers/components/ControllerHeaderForms/AddController/AddController.tsx
```diff
@@ -0,0 +1,47 @@
...
+};
+
+export const AddController = ({ clearHeaderContent }: Props): JSX.Element => {
+  const maasUrl = useSelector(configSelectors.maasUrl);
+  const rpcSharedSecret = useSelector(configSelectors.rpcSharedSecret);
+
+  return (
+    <>
+      <pre>
+        <code data-testid="instructions">{`# To add a new rack controller, SSH into the rack controller.
```
petermakowski, line 18: Is there a particular reason for wrapping the entire content along with instructions in a code block?

It makes it more difficult to read and has an awkward horizontal scroll when the content doesn't fit in (e.g. larger font or smaller screens):

image

I know you just copied it over from the legacy version, but what about we try to improve the UX as we go?

Unless there's a good reason for it I'd suggest to move everything out of the code formatted block except the commands themselves.

ui/src/app/controllers/components/ControllerHeaderForms/AddController/AddController.test.tsx
huwshimi: The pattern we've adopted is to set up the root state in beforeEach that is common to all the tests (which can be quite complex) and then override the parts of state that is required for a single test.

For a simple example in the App tests we have:

```
describe("App", () => {
  let state: RootState;

  beforeEach(() => {
    state = rootStateFactory({
      config: configStateFactory({
        items: [{ name: "completed_intro", value: true }],
      }),
    });
  });

  it("renders routes if logged in", () => {
    state.status.connected = true;
    state.status.authenticated = true;
    const store = mockStore(state);
    const wrapper = mount(
     ...
  });
```

This way we don't have to override the completed_intro config in every test, but it's also explicit what state is required for a single test.

ui/src/app/controllers/components/ControllerHeaderForms/AddController/AddController.test.tsx
huwshimi: Thanks, yep I suppose one of the child components could also include that generic string. The RTL example you've provided wouldn't make a conflict any less likely that a child component might provide a conflicting element.

I've added a prefix to the test id.

ui/src/app/controllers/components/ControllerHeaderForms/AddController/AddController.test.tsx
```diff
@@ -0,0 +1,58 @@
+import { mount } from "enzyme";
```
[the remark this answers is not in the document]
huwshimi: I appreciate your enthusiasm, but I'm not sure how productive it is to have this same discussion on every PR :).

ui/src/app/controllers/components/ControllerHeaderForms/AddController/AddController.test.tsx
huwshimi: Thanks, good suggestion. I've updated it so it looks like this now:

Screen Shot 2022-01-12 at 9 46 08 am

ui/src/app/controllers/components/ControllerHeaderForms/AddController/AddController.tsx
huwshimi: Thanks, good suggestion. I've updated it so it looks like this now:

Screen Shot 2022-01-12 at 9 46 08 am

Caleb-Ellis (approved):
Nice, this looks much better than legacy now

ui/src/app/controllers/components/ControllerHeaderForms/AddController/AddController.tsx
petermakowski: Awesome, looks much better.

PR #3503, opened 2022-01-12
Explore alternative state shape to better handle actions and errors

Caleb-Ellis:
## Problem

Our current state shape looks like this for each generic model, e.g.:
``` ts
{
  errors: null, // This can take a variety of different shapes
  items: [], // A list of instances of the model type
  loaded: false, // Whether the model type has been fetched once
  loading: false, // Whether the model type is currently fetching
  saved: false, // Whether some arbitrary action was successful - this is different depending on the model type
  saving: false, // Whether that arbitrary action is in progress
}
```

This was fine at the start but it has become outdated and unreliable.

Firstly, we don't have a way to link an error with the action or model that caused it. `error` in the state object is the raw data returned from the API when an error occurs without any of the necessary identifying data.

Secondly, there is no nuance with `saved`/`saving` - for some model types this just means "created"/"creating" but for some others it can mean "any action performed on this particular model type".

These issues have been solved somewhat with the inclusion of `eventErrors` and `statuses` for some model types like machines and devices, which look something like this:
``` ts
{
  eventErrors: [ // A list of errors with an "event" that caused it as well as the raw API error
    {
      error: {
        mac_addresses: [
          'MAC address 89:78:97:89:78:97 already in use on test-device.'
        ]
      },
      event: 'create'
    }
  ],
  statuses: { // A dict of primary keys for the model type with all the possible actions that can be performed on them
    qkfath: {
      creatingInterface: false,
      creatingPhysical: false,
      deleting: false,
      deletingInterface: false,
      linkingSubnet: false,
      unlinkingSubnet: false,
      updatingInterface: false,
      settingZone: false
    },
    f8xsss: {
      creatingInterface: false,
      creatingPhysical: false,
      deleting: false,
      deletingInterface: false,
      linkingSubnet: false,
      unlinkingSubnet: false,
      updatingInterface: false,
      settingZone: false
    },
}
```

The issue with this approach is that now we house errors in two places: `errors` and `eventErrors`, where `eventErrors` is far more useful. Also, the dict of `statuses` could get huge, especially for machines that can reach in the thousands, where each machine records 50 entries for each possible action. Finally, we don't know for sure whether any of those actions were truly successful - we only know whether it's currently in progress or not. In a few places we use a workaround that essentially checks that if the status transitions from `true` to `false` and there are no `eventErrors` for that action, then that means it was probably successful. E.g.:
``` ts
const updatingInterface = useSelector((state: RootState) =>
  deviceSelectors.getStatusForDevice(state, systemId, "updatingInterface")
);
const updateInterfaceErrored =
  useSelector((state: RootState) =>
    deviceSelectors.eventErrorsForDevices(state, systemId, "updateInterface")
  ).length > 0;
const [updatedInterface, resetUpdatedInterface] = useCycled(
  !updatingInterface
);
const saved = updatedInterface && !updateInterfaceErrored;
```

## Proposal

This PR aims to simplify and extend the ideas of `eventErrors` and `statuses` and make the base state shape generic across all model types (with the exclusion of some wildly different model types like `bootresource`). My proposal is for it to look something like this for the absolute simplest case (e.g zones, which only has CRUD actions):

``` ts
{
  zone: {
    errors: [ // "eventErrors" is renamed to "errors" and the old "errors" is removed
      {
        action: 'update', // The name of the action that caused the error
        error: { // The error returned from the API
          name: [
            'Physical zone with this Name already exists.'
          ]
        },
        modelPK: 82, // The primary key of the model if applicable, otherwise null for generic actions
      }
    ],
    genericActions: { // A dict of all the actions that are not model-specific
      create: {
        failed: false,
        processing: false,
        successful: false
      },
      fetch: {
        failed: false,
        processing: false,
        successful: false
      }
    },
    items: [], // Unchanged
    modelActions: { // A dict of all the actions that are model-specific
      delete: {
        failed: [], // A list of model primary keys
        processing: [],
        successful: []
      },
      update: {
        failed: [],
        processing: [],
        successful: []
      }
    }
  }
}
```

Firstly, `errors` is removed and `eventErrors` is renamed `errors`. It also optionally includes a `modelPK` so we can confidently relate an error from the API with the action and model that caused it.

Secondly, `loaded`, `loading`, `saved` and `saving` have been replaced with `genericActions`. These are all the actions for a model type that are not performed on an instance of that model, e.g fetching and creating. In the above example, `genericActions.fetch.processing` is the same as `loading`, `genericActions.create.successful` is the same as `saved` etc, but there is a clear link between what the original action was and what state that action is in. Also this structure allows us to easily extend the list of generic actions in a uniform way, for example `discovery.clear` which clears the discovery list would follow the exact same pattern.

Lastly, `statuses` has been replaced with `modelActions`. These are all the actions for a model type that are performed on an instance of that model, e.g. updating and deleting. Note I've opted to refer to an action in the same way everywhere in the codebase instead of using the continuous form (e.g. "updating") to keep everything consistent. Each action status (failed, processing, successful) contains a list of primary keys (e.g zone IDs), in which an ID can only be in one (or zero) status at a time, i.e. a zone cannot have successfully updated and also failed to update. This shape doesn't run into the same problem as growing linearly as with `statuses` while also being clear about the state of the actions.

The code example above could now be rewritten as something like:
``` ts
const updatingInterface = useSelector((state: RootState) =>
  deviceSelectors.getModelActionState(
    state,
    DEVICE_ACTIONS.updateInterface,
    ACTION_STATUS.processing,
    systemId
  )
);
const updatedInterface = useSelector((state: RootState) =>
  deviceSelectors.getModelActionState(
    state,
    DEVICE_ACTIONS.updateInterface,
    ACTION_STATUS.successful,
    systemId
  )
);
const updateInterfaceError = useSelector((state: RootState) =>
  deviceSelectors.getActionError(
    state,
    DEVICE_ACTIONS.updateInterface,
    systemId
  )
);
// Or we could abstract some parts away and use a hook, like:
const { error, saving, saved } = useDeviceAction(
  DEVICE_ACTIONS.updateInterface,
  systemId
);
```

In the new version we have 100% confidence of the status of each action and we don't have to use any workarounds/make any assumptions.

## Further work

There is obviously more work to be done on this PR, but following that we should aim to be rolling out the changes to all the other model types. The next step would be to add a `formId` to each action just like with `modelPK` in order to be able to relate an error with the specific form that was submitted. That way we aren't duplicating errors in places where we call an action in multiple places on the same page.


ui/src/app/base/sagas/websockets.ts
```diff
@@ -487,14 +487,14 @@ export function* handleMessage(
         }
         if (error) {
           yield* put({
-            meta: { item },
+            meta: { item, modelPK: action.meta.modelPK },
```
Caleb-Ellis, line 490: We need the `modelPK` to be present throughout the websocket saga in order to relate an error with a model, so we include it for each of the `Start`, `Success` and `Error` messages

ui/src/app/store/zone/constants.ts
```diff
@@ -0,0 +1,26 @@
...
+export const ACTION_STATUS = {
+  failed: "failed",
+  processing: "processing",
+  successful: "successful",
+} as const;
+/***************************/
+
+export const ZONE_PK = "id";
+
+export const ZONE_MODEL = "zone";
```
Caleb-Ellis, line 11: I've replaced the `ZoneMeta` enum with these two constants. I'm not married to this change, but my general preference is not to use enums since I feel like they trip me up by being both types and variables.

ui/src/app/store/zone/constants.ts
```diff
@@ -0,0 +1,26 @@
...
+
+export const ZONE_ACTIONS = {
+  cleanup: "cleanup",
+  create: "create",
+  delete: "delete",
+  fetch: "fetch",
+  update: "update",
+} as const;
```
Caleb-Ellis, line 19: This is the single source of all the possible zone actions names as we refer to them in the codebase. This makes many things simpler as we should no longer need to map between the static and continuous forms of the action name, e.g. "update" => "updating".

ui/src/app/store/zone/constants.ts
```diff
@@ -0,0 +1,26 @@
...
+
+export const ZONE_WEBSOCKET_METHODS = {
+  create: "create",
+  delete: "delete",
+  list: "list",
+  update: "update",
+} as const;
```
Caleb-Ellis, line 26: We typically only use these strings once ever, but I think housing it in a `ZONE_WEBSOCKET_METHODS` const object makes it clear when we're referring to an "action" (as we refer to it in the UI) and when we're referring to a "websocket method" (as expected by the API)

ui/src/app/zones/views/ZoneDetails/ZoneDetailsForm/ZoneDetailsForm.tsx
```diff
@@ -21,13 +22,29 @@ export type CreateZoneValues = {
...
+    )
+  );
+  const updating = useSelector((state: RootState) =>
+    zoneSelectors.getModelActionState(
+      state,
+      id,
+      ZONE_ACTIONS.update,
+      ACTION_STATUS.processing
+    )
+  );
```
petermakowski, line 47: When using a single status string this could be simplified like below:
```jsx
const status: 'idle' | 'processing' | 'success' | 'error' =
 useSelector(zoneSelectors.getModelActionstate(state, id, ZONE_ACTIONS.update)
```

ui/src/app/zones/views/ZoneDetails/ZoneDetailsForm/ZoneDetailsForm.tsx
```diff
@@ -57,8 +74,8 @@ const ZoneForm = ({ id, closeForm }: Props): JSX.Element | null => {
           );
         }}
         resetOnSave={true}
-        saved={saved}
-        saving={saving}
+        saved={updated}
+        saving={updating}
```
petermakowski, line 78: As above, you wouldn't ever want your component to end up in an inconsistent state (e.g.  both saved and saving at the same time).

Using a single status string/enum would help make illigal states such as this impossible and allow to simplify the codebase.

ui/src/app/store/zone/slice.ts
```diff
@@ -1,22 +1,322 @@
...
+    statuses[actionSuccessful] = false;
+  } else if (status === actionProcessing) {
+    statuses[actionFailed] = false;
+    statuses[actionProcessing] = true;
+    statuses[actionSuccessful] = false;
+  } else {
+    statuses[actionFailed] = false;
+    statuses[actionProcessing] = false;
+    statuses[actionSuccessful] = true;
+  }
```
petermakowski, line 95: Looks like only one of the statuses can be true at any given moment. Why not return a single status string?

petermakowski (commented):
Appreciate you're investigating improvements in this area, nice work. 👌🏻

I have some initial comments around statuses which I outlined below:

ui/src/app/zones/views/ZoneDetails/ZoneDetailsForm/ZoneDetailsForm.tsx
Caleb-Ellis: That's a great point! I much prefer this alternative, thanks

ui/src/app/zones/views/ZoneDetails/ZoneDetailsForm/ZoneDetailsForm.tsx
Caleb-Ellis: Agreed.

ui/src/app/store/zone/slice.ts
Caleb-Ellis: +1

Caleb-Ellis:
@petermakowski Thanks for the suggestions - now the state shape looks something like:
``` ts
{
  zone: {
    errors: [
      {
        action: 'create',
        error: "Some error string",
        modelPK: null
      }
    ],
    genericActions: {
      create: 'failed',
      fetch: 'successful'
    },
    items: [
      // Zones
    ],
    modelActions: {
      delete: {
        failed: [1, 2, 3],
        processing: [4],
        successful: [5]
      },
      update: {
        failed: [1, 2, 3],
        processing: [4],
        successful: [5]
      }
    }
  }
}
```

huwshimi:
@Caleb-Ellis this is really great.

How would this handle models without ids e.g. config? Would all the actions be tracked under `genericActions`?

The other thing that I'm not sure how we'd solve is when a model is performing the same action multiple times at once e.g. you could have two components doing `machine.check_power` (just a random example, maybe a better one would be `machine.create_bond` where each call could have different params) on the same machine at the same time and one could return an error and one could return successfully.

`config.update` is another case where we send multiple of those actions over the websocket at once and I'm not sure how we'd know which ones were successful and which ones failed.

Caleb-Ellis:
> How would this handle models without ids e.g. config? Would all the actions be tracked under `genericActions`?

Yeah I suppose for the sake of consistency that makes sense, "all the actions" being only `fetch` and `update`.

> The other thing that I'm not sure how we'd solve is when a model is performing the same action multiple times at once e.g. you could have two components doing `machine.check_power` (just a random example, maybe a better one would be `machine.create_bond` where each call could have different params) on the same machine at the same time and one could return an error and one could return successfully.

This is essentially a reframing of the original issue in which `pod.update` is called in two different places on the same page, so the same error is shown in two places. This PR won't address that issue, but the next step of adding a form ID (or similar) to the action meta object should. In that case we'd be able to handle each action start/success/error separately, and since the new `errors` shape is an array that gets pushed to, we should be able access errors for each form submission separately.

> `config.update` is another case where we send multiple of those actions over the websocket at once and I'm not sure how we'd know which ones were successful and which ones failed.

I completely forgot about this. I don't think changes in state shape will address that issue, that sounds like we probably need a custom saga to collate the responses. Either way, we have that problem now and this PR neither improves nor worsens it.

Caleb-Ellis:
I think this is finally ready for review! Sorry it's so huge - the actual, functional code changes is already significant, but I think the majority of the diff is updating tests (including every instance of the zone state factory).

Leaving the "Don't merge" label until next cycle.

ui/src/app/base/constants.ts
```diff
@@ -1,3 +1,10 @@
+export const ACTION_STATUS = {
+  failed: "failed",
+  idle: "idle",
+  processing: "processing",
+  successful: "successful",
```
petermakowski, line 5: I'd suggest to replace this with a slightly shorter status strings format that's used in many different libraries (e.g. https://github.com/tannerlinsley/react-query/blob/f2137dc4e4553256c4ebc1891b548fe35efe9231/src/core/types.ts#L250).

The convention below is widely adopted. Slight semantic difference between failed and error, or processing and loading doesn't seem to matter in this case - AFAIK the conceptual categories that we're referring to are the same.

```suggestion
  idle: "idle",
  loading: "loading",
  error: "error",
  success: "success",
```

Caleb-Ellis: Good suggestion, thanks! I guess I was coming up with terms that describe what happened (this thing failed, this thing was successful) but it might be more appropriate to describe the state declaratively instead (this thing is in error state, this thing is in success state).

petermakowski:
I wanted to take another look at this, but you have some failing tests. Could you let me know when it's ready for review 🙏  @Caleb-Ellis

huwshimi (commented):
This is looking really promising.

The main thing that I think that is missing is that right now we use the generic cleanup() action which I think we'll need to replace with a specific cleanup method for each action otherwise it'll clean up other actions that might be in progress. I.e. if we call delete we should call deleteCleanup(), if we call fetch() we should call cleanupFetch() etc which would reset the status (or remove the model ids from the arrays) and clear any errors (fetch would be a special case where it wouldn't reset the status if it was successful).

I guess the things we still need to explore following this will be new common utils for slices and selectors, looking at adapting this to also work with models without ids (e.g. config) and taking a look at associating forms and requests.

ui/src/app/devices/components/DeviceHeaderForms/AddDeviceForm/AddDeviceForm.test.tsx
```diff
@@ -38,8 +41,10 @@ describe("AddDeviceForm", () => {
         loaded: true,
       }),
       zone: zoneStateFactory({
+        genericActions: zoneGenericActionsFactory({
+          [ZONE_ACTIONS.fetch]: ACTION_STATUS.success,
+        }),
         items: [zoneFactory({ id: 0, name: "default" })],
-        loaded: true,
```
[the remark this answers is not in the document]
Caleb-Ellis: I'm pretty sure I didn't check the exact state requirements for each test, just got them working again with the new shape. But yeah that is something we could look into fixing as we go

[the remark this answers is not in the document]
huwshimi: Yeah, just might mean less updates to make.

ui/src/app/store/zone/selectors.ts
```diff
@@ -1,13 +1,108 @@
...
+import { ZONE_ACTIONS, ZONE_MODEL, ZONE_PK } from "./constants";
+import type {
+  ZoneActionNames,
+  ZoneGenericActions,
+  ZoneModelActions,
+  ZonePK,
+  ZoneState,
+} from "./types";
 
-const selectors = generateBaseSelectors<ZoneState, Zone, ZoneMeta.PK>(
```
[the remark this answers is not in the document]
Caleb-Ellis: Absolutely. I think everything I've done here for zones could be refactored into common utils.

ui/src/app/store/zone/slice.ts
```diff
@@ -1,22 +1,291 @@
...
+    statuses[loadingStatus].push(modelPK);
+  } else {
+    statuses[successStatus].push(modelPK);
+  }
+};
 
 const zoneSlice = createSlice({
-  name: ZoneMeta.MODEL,
-  initialState: genericInitialState as ZoneState,
-  reducers: generateCommonReducers<
```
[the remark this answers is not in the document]
Caleb-Ellis: Yeah for sure. I'm a little hesitant about recreating the `generateCommonReducers()` util though because it's a bit opaque - I like the idea of having all the action names + reducers in one file, clearly outlined. Maybe just utils for generating single reducers rather than a collection of reducers?

ui/src/app/store/zone/slice.ts
```diff
@@ -1,22 +1,291 @@
...
-  >(ZoneMeta.MODEL, ZoneMeta.PK),
+  name: ZONE_MODEL,
+  initialState: {
+    errors: [],
+    genericActions: initialGenericActions,
+    items: [],
+    modelActions: initialModelActions,
+  } as ZoneState,
+  reducers: {
+    [cleanupAction]: (state) => {
```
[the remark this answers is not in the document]
Caleb-Ellis: Ah good call. I think we could either have a cleanup for each action (so we could have `action`, `actionStart`, `actionSuccess`, `actionError` and `actionCleanup`) or maybe we could change the behaviour of the current `cleanup` function to be able to choose which actions to cleanup? i.e. `machineActions.cleanup([ZONE_ACTIONS.delete])`

ui/src/app/store/zone/slice.ts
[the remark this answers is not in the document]
huwshimi: The other option might be to provide an array of reducers to include so that is clear which ones apply to that model, something like `generateCommonReducers(["list", "getById", "delete"])`. I remember proposing that back when I was building the utils the first time, but I don't think there was a specific reason why we didn't implement it.

ui/src/app/store/zone/slice.ts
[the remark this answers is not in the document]
huwshimi: I think the latter is a good idea.

ui/src/app/store/zone/reducers.test.ts
```diff
@@ -1,59 +1,442 @@
...
-      items: [],
-      loaded: false,
-      loading: false,
-      saved: false,
-      saving: false,
+it("returns the initial state", () => {
+  expect(reducers(undefined, { type: "" })).toEqual({
+    errors: [],
+    genericActions: {
+      [ZONE_ACTIONS.create]: ACTION_STATUS.idle,
```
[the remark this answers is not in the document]
Caleb-Ellis: I'm not sure I understand the question, but in the machine list you could add a machine and get an error, and fetch machines and get a separate error and have both visible at the same time.

ui/src/app/store/zone/slice.ts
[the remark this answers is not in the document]
Caleb-Ellis: Could do that, but even in that case you wouldn't get the "true" list unless you included `["delete", "deleteStart", "deleteSuccess", "deleteError"]` etc.

Caleb-Ellis:
>In the end these consts end up being used as both values and types, just this way you need to use `typeof ...` or `ValueOf<typeof ...>`.
>It could be worth looking at using [`const enums`](https://www.typescriptlang.org/docs/handbook/enums.html#const-enums) which could potentially be a little friendlier than having to use `ValueOf<typeof ...>` to extract the types.

Yeah I'm not super happy with how using `as const` has come out. I think I'll just go back to using enums like we do for everything else, one less thing to change. It was mainly to test it out for curiosity's sake.

> This could possibly be a more generic key field for models without ids like config and general.

Hmm, I suppose. What about something like `modelRef` or `modelIdentifier`, in which case we could put e.g. config names in there?

ui/src/app/store/zone/reducers.test.ts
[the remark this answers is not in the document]
huwshimi: I was trying to think if we had anywhere that the same `genericAction` action could be called multiple times on the same page. e.g. if we called `model.create` multiple times in a loop we'd need to track the status for each call, even though they're the same endpoint.

Maybe something like[ this](https://github.com/canonical-web-and-design/maas-ui/blob/master/ui/src/app/store/bootresource/slice.ts#L76) if it was called in a loop, each call would be to the same endpoint and there's no id to key off but each call could have a different set of params.

It looks like we have this problem with `modelActions` too, e.g. the [`getLogs`](https://github.com/canonical-web-and-design/maas-ui/blob/master/ui/src/app/store/scriptresult/slice.ts#L208) gets [called in a loop](https://github.com/canonical-web-and-design/maas-ui/blob/master/ui/src/app/machines/views/MachineDetails/MachineTests/MachineTestsDetails/MachineTestsDetails.tsx#L51) with the same node result id but a different script type. So maybe keying of the model id isn't enough.

ui/src/app/base/constants.ts
```diff
@@ -1,3 +1,10 @@
+export const ACTION_STATUS = {
+  error: "error",
+  idle: "idle",
+  loading: "loading",
+  success: "success",
+} as const;
```
huwshimi, line 6: There are some cases where using a mutually exclusive status might not reflect the state of the data in Redux. For example `discovery.list` does not get cached so when we first fetch the list of discoveries we'd end up with `fetch: "success"` then later on if we tried to fetch the list again it would go in to `fetch: "loading"` and if it failed we'd have `fetch: "error"` but really the data still exists in the store and we should display it, we just couldn't update it.

It's possible that if we consider the status just the status of current request and stop using it to decide if data should be displayed then in might be OK (e.g. at the moment we hide content and display a spinner when it's in loading state, but instead we'd hav to check to see if there was data in the store).

We probably need to look into that a bit more to see if it'll cause us problems.

ui/src/app/base/types.ts
```diff
@@ -67,3 +70,23 @@ export type CommonActionFormProps<E = null> = {
...
+export type ModelAction<PK> = {
+  [ACTION_STATUS.error]: PK[];
+  [ACTION_STATUS.loading]: PK[];
+  [ACTION_STATUS.success]: PK[];
+};
+
+export type PayloadActionWithMeta<M, P = null> = PayloadAction<
+  P,
+  string,
+  { modelPK: M }
```
huwshimi, line 85: Some methods do not use the PK of model type e.g. when we [fetch event records](https://github.com/canonical-web-and-design/maas-ui/blob/master/ui/src/app/store/event/slice.ts#L20) it uses a node id, not an event record id so we couldn't store that request in `modelActions` otherwise the object would be keyed by a mix of node and event record ids (which could also potentially be conflicting).

ui/src/app/store/zone/slice.ts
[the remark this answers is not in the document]
huwshimi: Yeah though I'm not sure how much value there is in explicitly seeing those...

ui/src/app/base/sagas/websockets.ts
```diff
@@ -481,14 +481,14 @@ export function* handleMessage(
         // can be 0 when requesting a model with an id of 0.
         if (error || error === 0) {
           yield* put({
-            meta: { item },
+            meta: { item, modelPK: action.meta.modelPK },
```
huwshimi, line 484: It's possible we don't need `item` here anymore if we're explicitly setting the id.

huwshimi:
> > In the end these consts end up being used as both values and types, just this way you need to use `typeof ...` or `ValueOf<typeof ...>`.
> > It could be worth looking at using [`const enums`](https://www.typescriptlang.org/docs/handbook/enums.html#const-enums) which could potentially be a little friendlier than having to use `ValueOf<typeof ...>` to extract the types.
> 
> Yeah I'm not super happy with how using `as const` has come out. I think I'll just go back to using enums like we do for everything else, one less thing to change. It was mainly to test it out for curiosity's sake.

Sure, sounds good.

> 
> > This could possibly be a more generic key field for models without ids like config and general.
> 
> Hmm, I suppose. What about something like `modelRef` or `modelIdentifier`, in which case we could put e.g. config names in there?

I guess in some cases it might not be related to a model at all (e.g. it could be a script type). I'm not too fussed exactly what it gets called really as long as it's obvious what its use is.

ui/src/app/base/constants.ts
petermakowski: Thanks for pointing this out.

What I had in mind when I suggested an enum status was to model the state like a finite-state machine. What this means here - we might just be missing a status for the refetch error state? 

```tsx
export const ACTION_STATUS = {
  idle: "idle",
  loading: "loading",
  loadingError: "loadingError",
  refetchError: "refetchError",
  success: "success",
} as const;
```

Alternatively, we could add a common field for storing the number of retries that'd indicate if we're dealing with an initial load or refetch error. I'm a fan of how `xState` deals with this problem. Pasting an example, I hope it helps illustrate what I had in mind.

```tsx
import { createMachine, assign } from 'xstate';

interface Context {
  retries: number;
}

const fetchMachine = createMachine<Context>({
  id: 'fetch',
  initial: 'idle',
  context: {
    retries: 0
  },
  states: {
    idle: {
      on: {
        FETCH: 'loading'
      }
    },
    loading: {
      on: {
        RESOLVE: 'success',
        REJECT: 'failure'
      }
    },
    success: {
      type: 'final'
    },
    failure: {
      on: {
        RETRY: {
          target: 'loading',
          actions: assign({
            retries: (context, event) => context.retries + 1
          })
        }
      }
    }
  }
});
```
https://stately.ai/viz

ui/src/app/base/sagas/websockets.ts
Caleb-Ellis: Updated to use `identifier`

ui/src/app/store/zone/constants.ts
Caleb-Ellis: Reverted to using the `ZoneMeta` enum

ui/src/app/store/zone/reducers.test.ts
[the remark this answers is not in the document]
Caleb-Ellis: This would be addressed in the follow up that tracks each individual form submission, though I suppose "form submission" might be a bit too narrow, it's more like "API call".

ui/src/app/base/sagas/websockets.ts
Caleb-Ellis: Leaving it in for backwards compatibility

Caleb-Ellis:
Superseded by https://github.com/canonical-web-and-design/maas-ui/pull/4091

PR #3504, opened 2022-01-12
refactor(ui): subnets header, add fabric, add space

petermakowski:
## Done

migrate subnets header, add fabric, add space to React
- add aria-label to FormikField (there was no label associated with an input element)
- add cypress tests for adding a fabric (1 using legacy, 1 new UI)
- reported an issue with ContextualMenu and incorrect tab order https://github.com/canonical-web-and-design/react-components/issues/687

### Slight UX changes
I made a suggestion for a few changes that's been approved by design in the [planning doc](https://docs.google.com/document/d/1Vlih_9RyfNKKwLeBXS6tnRnNZTTR1BtXG_MHHC-Bi1s/edit#heading=h.wqx5o7fcpb6i)
- Add heading to the form (closer to https://vanillaframework.io/docs/layouts/application#panels-example)
- Add auto-focus on first form input for better accessibility

### QA steps

- go to `MAAS/r/networks?by=fabric` to verify the new UI

## Screenshots

image

image

Pasted Graphic 7

### DIFF
--- a/ui/src/app/subnets/views/FormActions/FormActions.tsx
+++ b/ui/src/app/subnets/views/FormActions/FormActions.tsx
@@ -15,25 +15,19 @@ const FormComponents: Record<
 };
 
 export interface FormActionProps {
-  activeForm: SubnetForm | null;
+  activeForm: SubnetForm;
   setActiveForm: React.Dispatch<React.SetStateAction<SubnetForm | null>>;
 }
 
 const FormActions = ({ activeForm, setActiveForm }: FormActionProps) => {
   const FormComponent = activeForm ? FormComponents[activeForm] : () => null;
 
-  const Form = () =>
-    activeForm ? (
-      <section className="p-strip is-shallow u-no-padding--top">
-        <div className="row">
-          <h2 className="p-heading--5">Add {activeForm}</h2>
-          <FormComponent
-            activeForm={activeForm}
-            setActiveForm={setActiveForm}
-          />
-        </div>
-      </section>
-    ) : null;
+  const Form = () => (
+    <>
+      <h2 className="p-heading--5">Add {activeForm}</h2>
+      <FormComponent activeForm={activeForm} setActiveForm={setActiveForm} />
+    </>
+  );
 
   return <Form />;
 };
--- a/ui/src/app/subnets/views/FormActions/FormActions.tsx
+++ b/ui/src/app/subnets/views/FormActions/FormActions.tsx
@@ -1,4 +1,5 @@
 import AddFabric from "./components/AddFabric";
+import AddSpace from "./components/AddSpace";
 
 import { SubnetForms } from "app/subnets/enum";
 import type { SubnetForm } from "app/subnets/types";
@@ -9,7 +10,7 @@ const FormComponents: Record<
 > = {
   [SubnetForms.Fabric]: AddFabric,
   [SubnetForms.VLAN]: () => null,
-  [SubnetForms.Space]: () => null,
+  [SubnetForms.Space]: AddSpace,
   [SubnetForms.Subnet]: () => null,
 };
 
--- a/ui/src/app/subnets/views/FormActions/FormActions.tsx
+++ b/ui/src/app/subnets/views/FormActions/FormActions.tsx
@@ -25,14 +25,9 @@ const FormActions = ({
 }: FormActionProps): JSX.Element => {
   const FormComponent = activeForm ? FormComponents[activeForm] : () => null;
 
-  const Form = () => (
-    <>
-      <h2 className="p-heading--5">Add {activeForm}</h2>
-      <FormComponent activeForm={activeForm} setActiveForm={setActiveForm} />
-    </>
+  return (
+    <FormComponent activeForm={activeForm} setActiveForm={setActiveForm} />
   );
-
-  return <Form />;
 };
 
 export default FormActions;
--- a/ui/src/app/subnets/views/FormActions/FormActions.tsx
+++ b/ui/src/app/subnets/views/FormActions/FormActions.tsx
@@ -1,0 +1,40 @@
+import { useDispatch } from "react-redux";
+
+import AddFabric from "./components/AddFabric";
+import type { SubnetForm } from "app/subnets/types";
+import { SubnetForms } from "app/subnets/enum";
+
+const FormComponents: Record<
+  SubnetForm,
+  ({ activeForm, setActiveForm }: FormActionProps) => JSX.Element | null
+> = {
+  [SubnetForms.Fabric]: AddFabric,
+  [SubnetForms.VLAN]: () => null,
+  [SubnetForms.Space]: () => null,
+  [SubnetForms.Subnet]: () => null,
+};
+
+export interface FormActionProps {
+  activeForm?: SubnetForm;
+  setActiveForm: React.Dispatch<React.SetStateAction<SubnetForm | undefined>>;
+}
+
+const FormActions = ({ activeForm, setActiveForm }: FormActionProps) => {
+  const FormComponent = activeForm ? FormComponents[activeForm] : () => null;
+
+  const Form = () =>
+    activeForm ? (
+      <section className="p-strip is-shallow">
+        <div className="row">
+          <FormComponent
+            activeForm={activeForm}
+            setActiveForm={setActiveForm}
+          />
+        </div>
+      </section>
+    ) : null;
+
+  return <Form />;
+};
+
+export default FormActions;
--- a/ui/src/app/subnets/views/FormActions/FormActions.tsx
+++ b/ui/src/app/subnets/views/FormActions/FormActions.tsx
@@ -23,7 +23,7 @@ const FormActions = ({ activeForm, setActiveForm }: FormActionProps) => {
 
   const Form = () =>
     activeForm ? (
-      <section className="p-strip is-shallow">
+      <section className="p-strip is-shallow u-no-padding--top">
         <div className="row">
           <h2 className="p-heading--5">Add {activeForm}</h2>
           <FormComponent
--- a/ui/src/app/subnets/views/FormActions/FormActions.tsx
+++ b/ui/src/app/subnets/views/FormActions/FormActions.tsx
@@ -15,8 +15,8 @@ const FormComponents: Record<
 };
 
 export interface FormActionProps {
-  activeForm?: SubnetForm;
-  setActiveForm: React.Dispatch<React.SetStateAction<SubnetForm | undefined>>;
+  activeForm: SubnetForm | null;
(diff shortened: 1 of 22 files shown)


ui/src/app/subnets/views/FormActions/components/AddFabric.tsx
```diff
@@ -0,0 +1,54 @@
...
+        action: "Add fabric",
+        category: "Subnets form actions",
+        label: "Add fabric",
+      }}
+      submitLabel={`Add ${activeForm}`}
+      onSubmit={({ name }: { name: string }) => {
+        dispatch(fabricActions.create({ name }));
+      }}
+      onCancel={() => setActiveForm(undefined)}
+      resetOnSave
```
Caleb-Ellis, line 36: The pattern we generally use for header forms is to close them when successfully submitted. For that we have an `onSuccess` prop that runs if `saved` changes from `false` to `true`, so this could be replaced with:
```
onSuccess={() => setActiveForm(null)}
```
Some other forms will include a secondary submit button with the label "Save and add another" (e.g. add machine) for cases where you might want to submit the form multiple times.

ui/src/app/subnets/views/FormActions/components/AddSpace.test.tsx
```diff
@@ -0,0 +1,40 @@
...
+
+  await waitFor(() =>
+    expect(store.getActions()).toEqual([
+      {
+        meta: { method: "create", model: "space" },
+        payload: { params: { name } },
+        type: "space/create",
+      },
+    ])
+  );
```
Caleb-Ellis, line 39: As above

ui/src/app/subnets/views/FormActions/components/AddSpace.tsx
```diff
@@ -0,0 +1,54 @@
...
+      buttonsBordered
+      allowAllEmpty
+      initialValues={{ name: "" }}
+      onSaveAnalytics={{
+        action: "Add space",
+        category: "Subnets form actions",
+        label: "Add space",
+      }}
+      submitLabel={`Add ${activeForm}`}
+      onSubmit={({ name }: { name: string }) => {
```
Caleb-Ellis, line 32: As above

ui/src/app/subnets/views/FormActions/components/AddSpace.tsx
```diff
@@ -0,0 +1,54 @@
...
+      }}
+      onCancel={() => setActiveForm(undefined)}
+      resetOnSave
+      saving={isSaving}
+      saved={isSaved}
+      errors={errors}
+    >
+      <FormikField
+        takeFocus
+        stacked
```
Caleb-Ellis, line 43: As above

ui/src/app/subnets/views/SubnetsList.tsx
```diff
@@ -1,13 +1,47 @@
...
+import Section from "app/base/components/Section";
+import SectionHeader from "app/base/components/SectionHeader";
+import { SubnetForms } from "app/subnets/enum";
+import type { SubnetForm } from "app/subnets/types";
+import FormActions from "app/subnets/views/FormActions";
+
+const SubnetsList = (): React.ReactElement => {
+  const [activeForm, setActiveForm] = React.useState<SubnetForm | undefined>(
+    undefined
+  );
```
Caleb-Ellis, line 14: We conventionally use `null` when we want to define a variable as having "no value", i.e. we consider `null` to mean the reference exists but is (currently) empty, and `undefined` to mean the reference does not exist.

ui/src/app/subnets/views/SubnetsList.tsx
```diff
@@ -1,13 +1,47 @@
...
+
+  return (
+    <Section
+      header={
+        <>
+          <SectionHeader
```
Caleb-Ellis, line 20: `SectionHeader` has a `headerContent` prop that is used to render things like forms in the header. It comes with a bit of styling, so using it should mean you can remove the top-level `section` and `div` from `FormActions`.

``` ts
headerContent={
  activeForm ? (
    <FormActions
      activeForm={activeForm}
      setActiveForm={setActiveForm}
    />
  ) : null
}
```

Since the existence check for `activeForm` happens here you could also remove it from `FormActions`.

ui/src/app/subnets/views/FormActions/FormActions.tsx
```diff
@@ -0,0 +1,41 @@
...
+  [SubnetForms.Space]: AddSpace,
+  [SubnetForms.Subnet]: () => null,
+};
+
+export interface FormActionProps {
+  activeForm?: SubnetForm;
+  setActiveForm: React.Dispatch<React.SetStateAction<SubnetForm | undefined>>;
+}
+
+const FormActions = ({ activeForm, setActiveForm }: FormActionProps) => {
```
Caleb-Ellis, line 22: Just missing return types on these function components.
``` ts
const FormActions = ({
  activeForm,
  setActiveForm,
}: FormActionProps): JSX.Element => {
```

ui/src/app/subnets/views/FormActions/components/AddFabric.tsx
petermakowski: Gotcha.

This will actually help mitigate an issue that I noticed:

[FormikFormButtons](https://github.com/canonical-web-and-design/maas-ui/blob/master/ui/src/app/base/components/FormikFormButtons/FormikFormButtons.tsx) disabled attribute is not in sync with its visual indicators for loading/success state.

This makes for a slightly confusing U, but most importantly makes it possible to make multiple requests unintentionally (e.g. user thinks nothing has happened and needs to try again).

Let me know if I used the component incorrectly, but even if I did, this shoul be looked at - I think that we should make such inconsistent state impossible.

You can see what I mean for yourself if you spin up maas-ui from a commit before I made this change: https://github.com/canonical-web-and-design/maas-ui/pull/3504/commits/a5c6d6385645323c63ea8804343c1cb1b30f61ce and press the Add Fabric submit button multiple times.

ui/src/app/subnets/views/FormActions/components/AddFabric.tsx
```diff
@@ -0,0 +1,54 @@
...
+      saving={isSaving}
+      saved={isSaved}
+      errors={errors}
+    >
+      <FormikField
+        takeFocus
+        stacked
```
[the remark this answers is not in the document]
petermakowski: Will do. Is this documented anywhere/should we disallow this prop in the FormikField component in this case?

ui/src/app/subnets/views/FormActions/components/AddSpace.test.tsx
```diff
@@ -0,0 +1,40 @@
...
+import userEvent from "@testing-library/user-event";
+import { Provider } from "react-redux";
+import { MemoryRouter } from "react-router-dom";
+import configureStore from "redux-mock-store";
+
+import AddSpace from "./AddSpace";
+
+import { rootState as rootStateFactory } from "testing/factories";
+
+test("correctly dispatches fabric create action on form submit", async () => {
```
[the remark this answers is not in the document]
petermakowski: 🙈

ui/src/app/subnets/views/FormActions/FormActions.tsx
petermakowski: Cool, I generally prefer TS to infer the return type, but can add a return type explicitly.

Shall I add an eslint rule to enforce this (or at least warn about it)?
https://github.com/typescript-eslint/typescript-eslint/blob/main/packages/eslint-plugin/docs/rules/explicit-function-return-type.md

Caleb-Ellis: I thought it was enforced already? Hmm, I wonder if one of the recent dependency updates included an eslint plugin that removed the rule. 😕

Anyway yeah it should be enforced to be consistent with our tsconfig - at the moment the build is failing due to these missing return types.

Caleb-Ellis: Right the rule was actually this one:
https://github.com/typescript-eslint/typescript-eslint/blob/main/packages/eslint-plugin/docs/rules/explicit-module-boundary-types.md

ui/src/app/subnets/views/SubnetsList.tsx
petermakowski: Thanks, didn't realise. I'll use your suggestion.

One thing to note though - I'm not sure it's a good idea to replace the h1 of the page you're on with the active form when the dropdown-activated form is opened. We're still within the context of the same page.

This happens on other pages that use this pattern, e.g. `MAAS/r/controllers`

image

image

### How I did it on the Subnets page:
image

Caleb-Ellis: I’m more than happy to go with your suggestion, but you should discuss it with @amylily1011 (and maybe @lyubomir-popov?) who initially proposed the pattern of having a contextual top level header

amylily1011: We sort of use the contextual form title instead of the header title because it is within the context of the same page and it is redundant to show the tab name and the tab title when we open the form. I think we should keep the current design (form title on the form header) for now. This is just so we can do an audit on all pages during MAAS IA work to prepare for the next version. 

We also went with the current pattern because we are preparing to transfer form content into a panel, which has a common pattern in Vanilla already. This way, you won't have to keep changing back and forth 😄 .

amylily1011: Thanks for pointing that out @petermakowski, this is a recurring problem that we faced for that past few cycles. The problem is people shouldn't see this slide down panel as a page. They should see is as a form and the H1 header provides the interpretation that it's a page, which is not our intention. And you pointing out that it confuses you that it's in a different page confirms that this creates a misconception. 

For this reason, our big goal is to change these forms into panels, which should apply to all products at Canonical.

The H1 header should not reflect the page, it should reflect the form title, because in MAAS the some forms can be triggered from different pages.

ui/src/app/subnets/views/FormActions/components/AddFabric.tsx
petermakowski: > If you mean that the button _should_ be disabled when the form is saving/saved but isn't

Yeah, that's what I meant.

ui/src/app/subnets/views/SubnetsList.tsx
amylily1011: So in theory the title of the page you're referencing should say "Add subnet", instead of just "Subnet".

Caleb-Ellis (approved):
Thanks for making the changes. One last thing - I think the form fields could live in a `col-6` but otherwise LGTM.

petermakowski:
>  I think the form fields could live in a col-6 but otherwise LGTM.

Thanks! I'll look into this in a next PR.

PR #3510, opened 2022-01-13
refactor: subnets add VLAN

petermakowski:
## Done

migrate Subnets / add VLAN to React
- add SpaceSelect component
- wrap forms in .col-6
- add cypress test

**design changes:**
- change input fields order (set VID input before Name input as it’s required)
- add help text to VID (Numeric value between 1 and 4094)

### Before
Pasted Graphic

### After
image

image

### MAAS deployment

- [Bolla](/HACKING.md#bolla)

### Running the branch

You can run this branch by:

### QA steps

- go to `/MAAS/r/networks?by=fabric` and verify adding a VLAN works correctly.

### DIFF
--- a/ui/src/app/base/components/FabricSelect/FabricSelect.tsx
+++ b/ui/src/app/base/components/FabricSelect/FabricSelect.tsx
@@ -1,6 +1,5 @@
 import { useEffect } from "react";
 
-import { useFormikContext } from "formik";
 import { useDispatch, useSelector } from "react-redux";
 
 import DynamicSelect from "app/base/components/DynamicSelect";
@@ -22,17 +21,10 @@ export const FabricSelect = ({
   const dispatch = useDispatch();
   const fabrics = useSelector(fabricSelectors.all);
   const fabricsLoaded = useSelector(fabricSelectors.loaded);
-  const { setFieldValue } = useFormikContext();
 
   useEffect(() => {
     if (!fabricsLoaded) dispatch(fabricActions.fetch());
   }, [dispatch, fabricsLoaded]);
-
-  useEffect(() => {
-    if (fabricsLoaded) {
-      setFieldValue(name, fabrics[0]?.id?.toString());
-    }
-  }, [name, fabricsLoaded, fabrics, setFieldValue]);
 
   return (
     <DynamicSelect
--- a/ui/src/app/base/components/FabricSelect/FabricSelect.tsx
+++ b/ui/src/app/base/components/FabricSelect/FabricSelect.tsx
@@ -1,49 +1,51 @@
 import { useEffect } from "react";
 
-import { Spinner } from "@canonical/react-components";
+import { useFormikContext } from "formik";
 import { useDispatch, useSelector } from "react-redux";
 
 import DynamicSelect from "app/base/components/DynamicSelect";
 import type { Props as FormikFieldProps } from "app/base/components/FormikField/FormikField";
 import { actions as fabricActions } from "app/store/fabric";
 import fabricSelectors from "app/store/fabric/selectors";
-import type { Fabric } from "app/store/fabric/types";
 
 type Props = {
-  defaultOption?: { label: string; value: string } | null;
+  defaultOption?: { label: string; value: string; disabled?: boolean } | null;
 } & FormikFieldProps;
 
 export const FabricSelect = ({
-  defaultOption = { label: "Select fabric", value: "" },
+  defaultOption = { label: "Select fabric", value: "", disabled: true },
   name,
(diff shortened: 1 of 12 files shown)


ui/src/app/base/components/FabricSelect/FabricSelect.tsx
```diff
@@ -1,49 +1,51 @@
...
-  }));
-
-  if (defaultOption) {
-    fabricOptions.unshift(defaultOption);
-  }
+  useEffect(() => {
+    if (fabricsLoaded) {
+      setFieldValue(name, fabrics[0]?.id);
+    }
+  }, [name, fabricsLoaded, fabrics, setFieldValue]);
```
petermakowski, line 35: Seems like rendering a spinner until the fabrics has loaded was a workaround for the Formik Field not having a value if there are no options on initial render.

I'm setting a field value explicitly to the first item on load.

ui/src/app/base/components/FabricSelect/FabricSelect.tsx
```diff
@@ -1,49 +1,51 @@
...
-  }));
-
-  if (defaultOption) {
-    fabricOptions.unshift(defaultOption);
-  }
+  useEffect(() => {
+    if (fabricsLoaded) {
+      setFieldValue(name, fabrics[0]?.id?.toString());
+    }
+  }, [name, fabricsLoaded, fabrics, setFieldValue]);
```
petermakowski, line 35: Seems like rendering a spinner until the fabrics has loaded was a workaround for the Formik Field not having a value if there are no options on initial render.

I'm setting a field value explicitly to the first item on load.

amylily1011:
LGTM!

ui/src/app/base/components/FabricSelect/FabricSelect.tsx
Caleb-Ellis: This isn't quite accomplishing what you want it to. Any time the `fabrics` array changes the value is hard-set to the first in the array.

e.g open two tabs on the subnets page. On the first tab, open the Add VLAN form and select a fabric that isn't the first one and don't submit the form. On the second tab add a new fabric, and you should see the first tab's fabric selection is now the first one.

I think in general we shouldn't be using `setFieldValue` in a `useEffect` but rather as the result of some other explicit action, e.g. I change this select and then some other value changes too. Maybe this would work better if instead we just disable the field if there are no options?

petermakowski: > Maybe this would work better if instead we just disable the field if there are no options?

That's what I intended - look below at `disabled={!fabricsLoaded || disabled}`

petermakowski:
@Caleb-Ellis could you please re-review. I added a test to ensure the component maintains selected option after new options are added.

ui/src/app/subnets/views/FormActions/components/AddVlan.tsx
```diff
@@ -0,0 +1,126 @@
...
+import spaceSelectors from "app/store/space/selectors";
+import { actions as vlanActions } from "app/store/vlan";
+import vlanSelectors from "app/store/vlan/selectors";
+import { toFormikNumber } from "app/utils";
+
+type AddVlanValues = {
+  vid?: string;
+  name?: string;
+  fabric?: string;
+  space?: string;
```
Caleb-Ellis, line 25: These must actually all be defined since `""` is still a value

ui/src/app/subnets/views/FormActions/components/AddVlan.tsx
```diff
@@ -0,0 +1,126 @@
...
+  useEffect(() => {
+    if (!fabricsLoaded) dispatch(fabricActions.fetch());
+    if (!spacesLoaded) dispatch(spaceActions.fetch());
+  }, [dispatch, fabricsLoaded, spacesLoaded]);
+
+  return (
+    <FormikForm<AddVlanValues>
+      validationSchema={vlanSchema}
+      buttonsBordered={false}
+      allowAllEmpty
```
Caleb-Ellis, line 62: This should be removed since a few fields are required. You typically only use this prop for forms that contain no fields, or when all the fields are optional (it's used to override a default formik behaviour IIRC)

Caleb-Ellis (approved):
Looks good, just some very minor comments :+1:

ui/src/app/subnets/views/FormActions/components/AddVlan.tsx
petermakowski: True, defined these before I've set all defaults to `""`

PR #3522, opened 2022-01-17
refactor(legacy): add subnet

petermakowski:
## Done

- refactor(legacy): add subnet

## Screenshots
image

### MAAS deployment

- [Bolla](/HACKING.md#bolla)

### Running the branch

You can run this branch by:

### QA steps

- Go to `MAAS/r/networks?by=fabric` and add a new subnet
- verify all options are stored correctly

## Fixes

Fixes: # .

## Launchpad issue

## Backports


ui/src/app/base/components/FormikFormContent/FormikFormContent.tsx
```diff
@@ -158,7 +158,7 @@ const FormikFormContent = <V, E = null>({
           {nonFieldError}
         </Notification>
       )}
-      {children}
+      {typeof children === "function" ? children({ values }) : children}
```
petermakowski, line 161: This allows using render props (just like regular Formik component does)

ui/src/app/base/components/VLANSelect/VLANSelect.tsx
```diff
@@ -15,26 +18,44 @@ type Props = {
   defaultOption?: { label: string; value: string } | null;
   fabric?: VLAN["fabric"];
   includeDefaultVlan?: boolean;
+  showSpinnerOnLoad?: boolean;
```
petermakowski, line 21: adding this prop for compatibility where this pattern is being used (tried to refactor, but this started taking too long)

ui/src/app/base/components/VLANSelect/VLANSelect.tsx
```diff
@@ -15,26 +18,44 @@ type Props = {
...
   const dispatch = useDispatch();
   let vlanList: VLAN[] = useSelector(vlanSelectors.all);
   const vlansLoaded = useSelector(vlanSelectors.loaded);
+  const selectedFabric = useSelector((state: RootState) =>
+    fabricSelectors.getById(state, fabric)
+  );
+  const { setFieldValue } = useFormikContext();
+
+  useEffect(() => {
+    if (setDefaultValueFromFabric) {
```
petermakowski, line 46: automatically set the value to the default vlan when a new fabric is selected. Hidden behind a boolean to allow other implementations (with explicit setFieldValue on change).

ui/src/app/base/components/VLANSelect/VLANSelect.tsx
```diff
@@ -15,26 +18,44 @@ type Props = {
...
+      const vlan = selectedFabric?.default_vlan_id;
+      if (vlan) {
```
Caleb-Ellis, line 48: Usually I try to avoid simple truthy checks like this when dealing with IDs, since some model IDs are indexed at 0. For VLANs this isn't the case (I think it starts at 5001 to completely avoid overlapping with the `vid`) but it's good to keep in mind.

ui/src/app/base/components/FormikFormContent/FormikFormContent.tsx
Caleb-Ellis: If we want full compatibility with using render props we might as well include the entire formik context instead of just `values`?

ui/src/app/subnets/views/FormActions/components/AddSubnet.tsx
```diff
@@ -0,0 +1,149 @@
...
+              />
+            </Col>
+          </Row>
+          <Row>
+            <Col size={6}>
+              <FabricSelect name="fabric" defaultOption={null} required />
+            </Col>
+            <Col size={6}>
+              <VLANSelect
+                name="vlan"
```
Caleb-Ellis, line 112: Might as well disable these selects too while the form is saving

Caleb-Ellis (approved):
LGTM! Just some minor comments for consideration

ui/src/app/subnets/views/FormActions/components/AddSubnet.tsx
```diff
@@ -0,0 +1,149 @@
...
+          })
+        );
+      }}
+      onCancel={() => setActiveForm(null)}
+      onSuccess={() => setActiveForm(null)}
+      saving={isSaving}
+      saved={isSaved}
+      errors={errors}
+    >
+      {({ values }) => (
```
huwshimi, line 81: We generally try to avoid this way of working with Formik as you're effectively defining a component here, but without the benefits of a separate component (for example this will result in a new function being defined for every render of the parent component).

Where we need to access formik values or other formik functions we create a new `...Fields` component .e.g. https://github.com/canonical-web-and-design/maas-ui/blob/master/ui/src/app/machines/components/MachineHeaderForms/AddMachine/AddMachineForm/AddMachineForm.tsx#L200

Then we access the values using `useFormikContext`: https://github.com/canonical-web-and-design/maas-ui/blob/master/ui/src/app/machines/components/MachineHeaderForms/AddMachine/AddMachineFormFields/AddMachineFormFields.tsx#L27

ui/src/app/base/components/FormikFormContent/FormikFormContent.tsx
huwshimi: I've left a comment below about using `useFormikContext` instead of an inner function.

ui/src/app/base/components/VLANSelect/VLANSelect.tsx
huwshimi: I recently added a util you could use here: `isId` that supports checking for `0` so we don't get false negatives: https://github.com/canonical-web-and-design/maas-ui/blob/master/ui/src/app/utils/isId.ts

ui/src/app/subnets/views/FormActions/components/AddSubnet.tsx
petermakowski: Thanks! Yeah, I used `useFormikContext` in a few places already. I can see the potential downsides of render props approach but also how it can be very convenient and simplify the implementation.

For now I will follow your suggestion for consistency.

PR #3526, opened 2022-01-18
ci: run each test suite in parallel

huwshimi:
## Done

- Run the tests suites in parallel.


huwshimi:
Looks like the update to node 16 is the cause of the ui tests taking nearly twice as long to complete.

Issue #3533, opened 2022-01-19
Can't open machine take action forms


huwshimi:
- Go to the machine list.
- Tick some machines.
- Open the take action menu and click an action.
- The form doesn't open and the machines get deselected.

PR #3535, opened 2022-01-19
fix(ui): fix route rerenders

huwshimi:
## Done

- Fix rerenders on all state changes due to using the [Route component prop](https://v5.reactrouter.com/web/api/Route/component). The relevant part of the docs being: `That means if you provide an inline function to the component prop, you would create a new component every render. This results in the existing component unmounting and the new component mounting instead of just updating the existing component.`

### MAAS deployment

- [Bolla](/HACKING.md#bolla)

### Running the branch

You can run this branch by:

### QA steps

- Go to the machine list.
- Tick some machines.
- Use the take action menu to open a form in the header. It should appear.

## Fixes

Fixes: #3533.

### DIFF
--- a/ui/src/app/Routes.tsx
+++ b/ui/src/app/Routes.tsx
@@ -35,11 +35,11 @@ const Routes = (): JSX.Element => (
     <Route
       exact
       path={baseURLs.index}
-      component={() => <Redirect to={machineURLs.machines.index} />}
+      render={() => <Redirect to={machineURLs.machines.index} />}
     />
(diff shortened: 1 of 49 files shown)


Caleb-Ellis (approved):
I wonder what the purpose of `component` is, then? 😕

huwshimi:
> I wonder what the purpose of `component` is, then? 😕

No idea! I wonder how long we've had performance issues relating to our use of `component`.

Issue #3550, opened 2022-01-25
Dedupe timestamped types


huwshimi:
We should create a base TimestampedModel type to dedupe the created/updated params to match the backend: https://github.com/maas/maas/blob/master/src/maasserver/models/timestampedmodel.py

PR #3551, opened 2022-01-25
refactor(ui): add timestamped model

huwshimi:
## Done

- Move created and updated types to a shared type.

## Fixes

Fixes: #3550.


ui/src/app/store/types/model.ts
```diff
@@ -2,6 +2,13 @@ export type Model = {
   id: number;
 };
 
+export type TimestampFields = {
+  created: string;
+  updated: string;
+};
+
+export type TimestampedModel = Model & TimestampFields;
```
petermakowski, line 10: Personally I think this _might_ be complicating things unnecessarily. I might be clearer if the model was left as is and we treated time stamp fields separately without creating a separate type that's intersection of the two.

The naming makes sense. Maybe it's just a matter of getting used to it. 🤷‍♂️

petermakowski (approved):
LGTM! 👍

PR #3566, opened 2022-02-01
feat(ui): add helper components for rendering network model links

Caleb-Ellis:
## Done

- Built helper components for each of fabric, space, subnet and VLANs that simply, given an ID:
  - returns a spinner if that model type is loading
  - handles when a model with that ID does not exist
  - returns a link to the React details page of that model instance

### MAAS deployment

- [Bolla](/HACKING.md#bolla)

### Running the branch

You can run this branch by:

### QA steps

- Go to the React details page of a VLAN and check that the fabric and space links still render as before 

## Fixes

Fixes canonical-web-and-design/app-tribe#717

### DIFF
--- a/ui/src/app/base/components/FabricLink/FabricLink.tsx
+++ b/ui/src/app/base/components/FabricLink/FabricLink.tsx
@@ -1,0 +1,41 @@
+import { Spinner } from "@canonical/react-components";
+import { useSelector } from "react-redux";
+import { Link } from "react-router-dom";
+
+import fabricSelectors from "app/store/fabric/selectors";
+import type { Fabric, FabricMeta } from "app/store/fabric/types";
+import { getFabricDisplay } from "app/store/fabric/utils";
+import type { RootState } from "app/store/root/types";
+import subnetsURLs from "app/subnets/urls";
+
+type Props = {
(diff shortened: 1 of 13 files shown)


ui/src/app/base/components/FabricLink/FabricLink.tsx
```diff
@@ -0,0 +1,41 @@
...
+    // TODO: Put aria-label directly on Spinner component when issue is fixed.
+    // https://github.com/canonical-web-and-design/react-components/issues/651
+    return (
+      <span aria-label="Loading fabrics">
+        <Spinner />
+      </span>
+    );
+  }
+  if (!fabric) {
+    return <span aria-label="Fabric does not exist">{fabricDisplay}</span>;
```
huwshimi, line 32: Is it a bit odd to have an aria-label but no visual content (same for all these components)?

huwshimi (approved):
One question.

ui/src/app/base/components/FabricLink/FabricLink.tsx
Caleb-Ellis: Yeah I suppose these could just return null instead

PR #3569, opened 2022-02-01
feat(ui): build Subnets in this VLAN table

Caleb-Ellis:
## Done

- Built Subnets on this VLAN table in VLAN details page
  - I've updated it a bit by removing the "Space" column since it's a VLAN-specific piece of data (i.e. every subnet in the table will be in the same space, so its inclusion is redundant). In its place I've added a few columns of extra data.

### MAAS deployment

- [Bolla](/HACKING.md#bolla)

### Running the branch

You can run this branch by:

### QA steps

- Go to the React details page of a VLAN and check that there is a table for subnets in that VLAN
- Check that the data is correct as per the legacy details pages for those subnets
- Check that sorting works correctly

## Fixes

Fixes canonical-web-and-design/app-tribe#671

### DIFF
--- a/ui/src/app/subnets/views/VLANDetails/VLANSubnets/VLANSubnets.test.tsx
+++ b/ui/src/app/subnets/views/VLANDetails/VLANSubnets/VLANSubnets.test.tsx
@@ -1,0 +1,63 @@
+import { render, screen, within } from "@testing-library/react";
+import { Provider } from "react-redux";
+import { MemoryRouter } from "react-router-dom";
+import configureStore from "redux-mock-store";
+
+import VLANSubnets from "./VLANSubnets";
+
+import subnetsURLs from "app/subnets/urls";
+import {
+  rootState as rootStateFactory,
+  subnet as subnetFactory,
+  subnetState as subnetStateFactory,
+  subnetStatistics as subnetStatisticsFactory,
+  vlan as vlanFactory,
+  vlanState as vlanStateFactory,
+} from "testing/factories";
+
+const mockStore = configureStore();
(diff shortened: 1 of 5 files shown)


ui/src/app/subnets/views/VLANDetails/VLANSubnets/VLANSubnets.test.tsx
```diff
@@ -0,0 +1,61 @@
...
+  subnet as subnetFactory,
+  subnetState as subnetStateFactory,
+  subnetStatistics as subnetStatisticsFactory,
+  vlan as vlanFactory,
+  vlanState as vlanStateFactory,
+} from "testing/factories";
+
+const mockStore = configureStore();
+
+it("renders correct details", () => {
```
huwshimi, line 20: It feels like this doesn't really test what kind of UI we're showing to the user. I'm not totally sure what the solution is, we don't need to test the internals of MainTable, but these tests would pass even if it was displaying a couple of `<p>`s on the page.

huwshimi (approved):
Just a comment about revisiting the tests.

ui/src/app/subnets/views/VLANDetails/VLANSubnets/VLANSubnets.test.tsx
Caleb-Ellis: I could increase the specificity to assert the details are indeed inside a table, instead of just floating around somewhere on the page. TBH I'm still trying to work my way around how we're "supposed" to be writing tests with RTL.

PR #3580, opened 2022-02-02
feat(ui): build VLAN DHCP status section

Caleb-Ellis:
## Done

- Built DHCP status section in VLAN details. I've updated the design a little bit so that "High availability" is shown as a boolean instead of only showing when it's true, and also some of the model names are now links.

### MAAS deployment

- [Bolla](/HACKING.md#bolla)

### Running the branch

You can run this branch by:

### QA steps

- Go to the VLAN details page of a VLAN with/without DHCP (MAAS, external, relayed) and check that the text matches the legacy version. Note that a lot of edge cases will be hard to test on bolla.

## Fixes

Fixes canonical-web-and-design/app-tribe#668

## Screenshot
Screenshot 2022-02-02 at 18-54-06 Default VLAN details bolla MAAS

### DIFF
--- a/ui/src/app/base/components/ControllerLink/ControllerLink.test.tsx
+++ b/ui/src/app/base/components/ControllerLink/ControllerLink.test.tsx
@@ -1,18 +1,70 @@
 import { generateLegacyURL } from "@maas-ui/maas-ui-shared";
 import { render, screen } from "@testing-library/react";
+import { Provider } from "react-redux";
+import { MemoryRouter } from "react-router-dom";
+import configureStore from "redux-mock-store";
 
 import ControllerLink from "./ControllerLink";
 
 import baseURLs from "app/base/urls";
-import { controller as controllerFactory, modelRef } from "testing/factories";
+import {
+  controller as controllerFactory,
+  controllerState as controllerStateFactory,
+  modelRef as modelRefFactory,
+  rootState as rootStateFactory,
+} from "testing/factories";
 
-const controller = controllerFactory({
-  hostname: "bolla",
-  system_id: "1234",
-  domain: modelRef({ name: "maas" }),
+const mockStore = configureStore();
+
+it("handles when controllers are loading", () => {
+  const state = rootStateFactory({
+    controller: controllerStateFactory({ items: [], loading: true }),
+  });
+  const store = mockStore(state);
+  render(
+    <Provider store={store}>
+      <MemoryRouter>
+        <ControllerLink systemId="abc123" />
+      </MemoryRouter>
+    </Provider>
+  );
+
+  expect(screen.getByLabelText("Loading controllers")).toBeInTheDocument();
 });
-it("displays a correct link", () => {
-  render(<ControllerLink {...controller} />);
+
+it("handles when a controller does not exist", () => {
+  const state = rootStateFactory({
+    controller: controllerStateFactory({ items: [], loading: false }),
+  });
+  const store = mockStore(state);
+  const { container } = render(
+    <Provider store={store}>
+      <MemoryRouter>
+        <ControllerLink systemId="abc123" />
+      </MemoryRouter>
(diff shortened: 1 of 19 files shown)


ui/src/app/base/components/ControllerLink/ControllerLink.test.tsx
```diff
@@ -1,18 +1,70 @@
...
+  const store = mockStore(state);
+  render(
+    <Provider store={store}>
+      <MemoryRouter>
+        <ControllerLink systemId="abc123" />
+      </MemoryRouter>
+    </Provider>
+  );
+
+  expect(screen.getByLabelText("Loading controllers")).toBeInTheDocument();
```
huwshimi, line 32: This tests the label, but not that it's displaying a spinner.

ui/src/app/base/components/ControllerLink/ControllerLink.tsx
```diff
@@ -1,16 +1,38 @@
...
-import type { BaseController } from "app/store/controller/types/base";
+import controllerSelectors from "app/store/controller/selectors";
+import type { Controller, ControllerMeta } from "app/store/controller/types";
+import type { RootState } from "app/store/root/types";
+
+type Props = {
+  systemId?: Controller[ControllerMeta.PK] | null;
+};
+
+const ControllerLink = ({ systemId }: Props): JSX.Element | null => {
```
huwshimi, line 14: Should this also fetch the controllers to make sure they exist?

ui/src/app/subnets/views/VLANDetails/DHCPStatus/DHCPStatus.tsx
```diff
@@ -1,5 +1,117 @@
...
+        <Link to={subnetsURLs.vlan.index({ id: vlan.relay_vlan })}>
+          {getFullVLANName(vlan.relay_vlan, vlans, fabrics)}
+        </Link>
+      </span>
+    );
+  }
+  return "Disabled";
+};
+
+const DHCPStatus = ({ id }: Props): JSX.Element | null => {
```
huwshimi, line 44: Should this also fetch the vlans, fabrics and subnets? I wonder if that would be better to do in this component as someone could easily remove it from the parent component and this would break.

huwshimi (approved):
A couple of small comments.

ui/src/app/base/components/ControllerLink/ControllerLink.tsx
Caleb-Ellis: Yeah, since this exists in base components it should probably just handle everything

ui/src/app/subnets/views/VLANDetails/DHCPStatus/DHCPStatus.tsx
Caleb-Ellis: It could, but I'm of two minds about including all the `fetch()` actions even if a parent component already does that. Taken to its logical extreme we should be running `fetch()` or `get()` functions for every component that retrieves something from state (e.g. in table cell components), which to me seems cumbersome. And in some cases where we're passing an `id` prop down 4 or 5 levels, are we meant to include `fetch()` at every step? But yeah, it does mean the component can be moved anywhere and "just work". I dunno, what do you think?

huwshimi: Yeah I think it's a balancing act. `fetch()` is cheap because it's cached. In this case I was particularly thinking about the fact you added `dispatch(fabricActions.fetch());` in the parent component but its not data is not used by that component so I can imagine someone in the future thinking it's no longer needed and remove it.

Caleb-Ellis: That's fair. I'll add them in.

PR #3583, opened 2022-02-03
feat(ui): add iprange slice

huwshimi:
## Done

- Add a slice to handle reducers, actions, types etc. for the iprange model.

## Fixes

Fixes: canonical-web-and-design/app-tribe#674.

### DIFF
--- a/ui/src/app/store/iprange/types/base.ts
+++ b/ui/src/app/store/iprange/types/base.ts
@@ -10,10 +10,10 @@ export type IPRange = TimestampedModel & {
   comment: string;
   end_ip: string;
   start_ip: string;
-  subnet: Subnet[SubnetMeta.PK];
+  subnet: Subnet[SubnetMeta.PK] | null;
   type: IPRangeType;
-  user?: string;
-  vlan?: VLAN[VLANMeta.PK];
+  user: string;
+  vlan: VLAN[VLANMeta.PK] | null;
 };
 
 export type IPRangeState = GenericState<IPRange, APIError>;
--- a/ui/src/app/store/iprange/types/base.ts
+++ b/ui/src/app/store/iprange/types/base.ts
@@ -1,0 +1,19 @@
+import type { IPRangeType } from "./enum";
+
+import type { APIError } from "app/base/types";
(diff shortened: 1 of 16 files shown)


ui/src/app/store/iprange/types/base.ts
```diff
@@ -0,0 +1,19 @@
...
+import type { GenericState } from "app/store/types/state";
+import type { VLAN, VLANMeta } from "app/store/vlan/types";
+
+export type IPRange = TimestampedModel & {
+  comment: string;
+  end_ip: string;
+  start_ip: string;
+  subnet: Subnet[SubnetMeta.PK];
+  type: IPRangeType;
+  user?: string;
```
Caleb-Ellis, line 15: This is always defined, just as an empty string if there is no user.
`data["user"] = "" if obj.user is None else obj.user.`

ui/src/app/store/iprange/types/base.ts
```diff
@@ -0,0 +1,19 @@
...
+  vlan?: VLAN[VLANMeta.PK];
```
Caleb-Ellis, line 16: Similar here, looks like it should be `vlan: VLAN[VLANMeta.PK] | null`
`data["vlan"] = None if obj.subnet is None else obj.subnet.vlan_id`

ui/src/app/store/iprange/types/base.ts
```diff
@@ -0,0 +1,19 @@
...
+import type { Subnet, SubnetMeta } from "app/store/subnet/types";
+import type { TimestampedModel } from "app/store/types/model";
+import type { GenericState } from "app/store/types/state";
+import type { VLAN, VLANMeta } from "app/store/vlan/types";
+
+export type IPRange = TimestampedModel & {
+  comment: string;
+  end_ip: string;
+  start_ip: string;
+  subnet: Subnet[SubnetMeta.PK];
```
Caleb-Ellis, line 13: Looks like this can also be `null`

Caleb-Ellis (approved):
LGTM, just some minor type changes

PR #3600, opened 2022-02-08
Build subnet summary edit form

steverydz:
## Done

- Itemised list of what was changed by this PR.

### MAAS deployment

- [Bolla](/HACKING.md#bolla)

### Running the branch

You can run this branch by:

### QA steps

- Steps for QA.

## Fixes

Fixes: # .

## Launchpad issue

## Backports

### DIFF
--- a/ui/src/app/subnets/views/SubnetDetails/SubnetSummary/SubnetSummaryForm/SubnetSummaryForm.tsx
+++ b/ui/src/app/subnets/views/SubnetDetails/SubnetSummary/SubnetSummaryForm/SubnetSummaryForm.tsx
@@ -11,7 +11,7 @@ import { actions as fabricActions } from "app/store/fabric";
 import type { RootState } from "app/store/root/types";
 import { actions as subnetActions } from "app/store/subnet";
 import subnetSelectors from "app/store/subnet/selectors";
-import type { Subnet } from "app/store/subnet/types";
+import type { Subnet, SubnetMeta } from "app/store/subnet/types";
 import { actions as vlanActions } from "app/store/vlan";
 import vlanSelectors from "app/store/vlan/selectors";
 
@@ -25,29 +25,37 @@ const subnetSummaryFormSchema = Yup.object().shape({
   active_discovery: Yup.boolean(),
   allow_proxy: Yup.boolean(),
   allow_dns: Yup.boolean(),
-  vlan: Yup.number(),
+  vlan: Yup.number().required("VLAN is required"),
 });
 
-const SubnetSummaryForm = ({
-  subnet,
-  handleDismiss,
-}: {
-  subnet: Subnet;
+type Props = {
   handleDismiss: () => void;
-}): JSX.Element => {
+  id: Subnet[SubnetMeta.PK];
+};
+
+const SubnetSummaryForm = ({
+  handleDismiss,
+  id,
+}: Props): JSX.Element | null => {
+  const dispatch = useDispatch();
   const subnetErrors = useSelector(subnetSelectors.errors);
   const saving = useSelector(subnetSelectors.saving);
   const saved = useSelector(subnetSelectors.saved);
-  const dispatch = useDispatch();
-
+  const subnet = useSelector((state: RootState) =>
+    subnetSelectors.getById(state, id)
+  );
   const vlan = useSelector((state: RootState) =>
     vlanSelectors.getById(state, subnet?.vlan)
   );
 
   useEffect(() => {
-    dispatch(vlanActions.fetch());
     dispatch(fabricActions.fetch());
+    dispatch(vlanActions.fetch());
   }, [dispatch]);
+
+  if (!subnet || !vlan) {
+    return null;
+  }
 
   return (
     <FormikForm<SubnetSummaryFormValues>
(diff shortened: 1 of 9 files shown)


steverydz:
@Caleb-Ellis I've still not been able to make the fabric save when it's changed. I'm using the `values` object rather than the vlan and subsequently fabric from the subnet directly.

The only thing I can think of is that on submit the form should also dispatch `vlanActions.update()` using the new fabric ID. I have tried this but am not able to get past TypeScript which keeps giving me errors like: `Property does not exist on type 'number'` when I pass in the fabric ID.

I've also tried using the `update` function as it is here in the `EditVLAN` view: https://github.com/canonical-web-and-design/maas-ui/blob/master/ui/src/app/subnets/views/VLANDetails/EditVLAN/EditVLAN.tsx#L96

This gives me the following:

```
Argument of type '{ id: number; } | { id: number; created: string; updated: string; description: string; dhcp_on: boolean; external_dhcp: string | null; fabric: number; mtu: number; name: string; primary_rack: string | null; ... 4 more ...; vid: number; } | { ...; }' is not assignable to parameter of type 'UpdateParams'.
  Type '{ id: number; }' is not assignable to type 'UpdateParams'.
    Property 'vid' is missing in type '{ id: number; }' but required in type 'CreateParams'.
```

Caleb-Ellis:
@steverydz I've pushed up a commit that fixes the functionality of the form but did not add any tests. You got *very* close to making this work, but there were a couple of little (admittedly obtuse) things that were missing. I've left comments on the commit and tagged you.

ui/src/app/subnets/views/SubnetDetails/SubnetSummary/SubnetSummaryForm/SubnetSummaryForm.tsx
```diff
@@ -25,80 +25,76 @@ const subnetSummaryFormSchema = Yup.object().shape({
...
   allow_proxy: Yup.boolean(),
   allow_dns: Yup.boolean(),
-  vlan: Yup.number(),
+  vlan: Yup.number().required("VLAN is required"),
 });
 
+type Props = {
+  handleDismiss: () => void;
+  id: Subnet[SubnetMeta.PK];
+};
```
Caleb-Ellis, line 34: @steverydz Typically we define a `Props` type separate from the component - just think it's a bit neater, plus we can export it if we need to.

ui/src/app/subnets/views/SubnetDetails/SubnetSummary/SubnetSummaryForm/SubnetSummaryForm.tsx
```diff
@@ -25,80 +25,76 @@ const subnetSummaryFormSchema = Yup.object().shape({
...
 
   useEffect(() => {
-    dispatch(vlanActions.fetch());
     dispatch(fabricActions.fetch());
+    dispatch(vlanActions.fetch());
   }, [dispatch]);
 
+  if (!subnet || !vlan) {
+    return null;
+  }
```
Caleb-Ellis, line 58: Instead of using a bunch of optional chaining in the form, I think it's cleaner to just exit out quick if we don't have the data we need. With this the TS compiler knows that beyond this conditional `subnet` and `vlan` definitely exist.

A better option would be to show a spinner but I took a shortcut here.

ui/src/app/subnets/views/SubnetDetails/SubnetSummary/SubnetSummaryForm/SubnetSummaryForm.tsx
```diff
@@ -25,80 +25,76 @@ const subnetSummaryFormSchema = Yup.object().shape({
...
-        description,
-        managed,
-        active_discovery,
-        allow_proxy,
-        allow_dns,
-        vlan,
-        fabric,
-      }) => {
+      onSubmit={(values) => {
+        dispatch(subnetActions.cleanup());
```
Caleb-Ellis, line 84: This is just to clear errors from a previous submission (if they exist)

ui/src/app/subnets/views/SubnetDetails/SubnetSummary/SubnetSummaryForm/SubnetSummaryFormFields/SubnetSummaryFormFields.tsx
```diff
@@ -73,23 +46,12 @@ const SubnetSummaryFormFields = ({
...
-        {values.fabric && (
-          <FabricSelect
-            name="Fabric"
-            defaultOption={{
-              label: getFabricDisplay(fabric) || "",
-              value: values?.fabric.toString() || "",
-            }}
-          />
-        )}
+        <FabricSelect name="fabric" />
```
Caleb-Ellis, line 49: The main problem here initially was that the `name` prop was "Fabric" instead of "fabric". The `name` prop has to correspond to a form value name exactly. So this is one reason why the fabric never updated the VLAN options.

We could probably handle this a bit better with TS generics so that it's impossible to use an invalid `name`, but for now we just have to remember to get it right.

ui/src/app/subnets/views/SubnetDetails/SubnetSummary/SubnetSummaryForm/SubnetSummaryFormFields/SubnetSummaryFormFields.tsx
```diff
@@ -73,23 +46,12 @@ const SubnetSummaryFormFields = ({
...
         <VLANSelect
+          fabric={Number(values.fabric)}
           name="vlan"
+          setDefaultValueFromFabric
           showSpinnerOnLoad
-          fabric={vlan?.fabric}
-          defaultOption={{
-            label: getVLANDisplay(vlan) || "",
-            value: values?.vlan.toString(),
-          }}
```
Caleb-Ellis: This was a bit trickier to figure out. Basically, there's some internal logic in `VLANSelect` to automatically render a list of VLANs based on a given fabric. The way it does that is via the `fabric` prop, which expects a fabric ID (type === `number`). However, in the `FabricSelect` component it sets that ID to a string, so there is actually some inconsistency between what we've typed the form values to be and what they actually are. I'm not certain why the IDs are coerced to be strings (@petermakowski might be able to answer that), but for the form values to be the right type, `fabric` and `vlan` should both be `string`. And since that's the case, we have to convert it *back* to `number` when we pass it to the `VLANSelect` component, because otherwise you'll never find a fabric with e.g `id === "1"`.

I'm happy to jump on a call when you get in if I'm not being very clear. :+1:

Caleb-Ellis (approved):

PR #3610, opened 2022-02-10
generate unique ids when mocking nanoid

huwshimi:
## Done

- generate unique ids when mocking nanoid


ui/src/setupTests.ts
```diff
@@ -6,12 +6,4 @@ import { enableFetchMocks } from "jest-fetch-mock";
 
 Enzyme.configure({ adapter: new Adapter() });
 
-jest.mock("@reduxjs/toolkit", () => {
-  const toolkit = jest.requireActual("@reduxjs/toolkit");
-  return {
-    ...toolkit,
-    nanoid: jest.fn(() => "Uakgb_J5m9g-0JDMbcJqLJ"),
```
huwshimi, line 13: This was previously returning `undefined` as it was being reset between tests.

ui/src/app/base/hooks/base.ts
```diff
@@ -2,7 +2,7 @@ import { useCallback, useEffect, useRef, useState } from "react";
 
 import { NotificationSeverity } from "@canonical/react-components";
 import type { NotificationProps } from "@canonical/react-components";
-import { nanoid } from "nanoid/non-secure";
+import { nanoid } from "@reduxjs/toolkit";
```
huwshimi, line 5: @petermakowski `nanoid` exported from redux toolkit is the nonsecure version: https://redux-toolkit.js.org/api/other-exports#nanoid.

huwshimi: I meant to also say that by switching it back to the reduxjs/toolkit version it gives us one less thing to mock.

Caleb-Ellis (approved):

ui/src/app/base/hooks/base.ts
petermakowski: This was done on purpose, hence the comment:
```jsx
 * @returns non-secure random ID string
```
https://github.com/canonical-web-and-design/maas-ui/blob/master/ui/src/app/base/hooks/base.ts#L164

non-secure version is faster and we don't really need enhanced security when generating IDs for the DOM elements.

PR #3616, opened 2022-02-11
feat(ui): display a subnet column for VLAN reserved ranges

huwshimi:
## Done

- Display a subnet column when the reserved ranges table is displayed for a VLAN.

### MAAS deployment

- [Bolla](/HACKING.md#bolla)

### Running the branch

You can run this branch by:

### QA steps

- Go to a subnet details page and scroll down to the reserved ranges table.
- There should be no subnet column on the left hand side.
- Go to a VLAN details page and scroll down to the reserved ranges table.
- There should be a subnet column on the left hand side.

## Fixes

Fixes: canonical-web-and-design/app-tribe#723

## Screenshots

VLAN details:
Screen Shot 2022-02-11 at 1 39 18 pm
Subnet details:
Screen Shot 2022-02-11 at 1 39 29 pm

### DIFF
--- a/ui/src/app/subnets/components/ReservedRangeForm/ReservedRangeForm.tsx
+++ b/ui/src/app/subnets/components/ReservedRangeForm/ReservedRangeForm.tsx
@@ -31,7 +31,7 @@ export enum Labels {
   CreateRange = "Create reserved range",
   EditRange = "Edit reserved range",
   EndIp = "End IP address",
-  Comment = "Comment",
+  Comment = "Purpose",
   StartIp = "Start IP address",
 }
 
@@ -142,7 +142,7 @@ const ReservedRangeForm = ({
               disabled={showDynamicComment}
               label={Labels.Comment}
               name="comment"
-              placeholder="IP range comment (optional)"
+              placeholder="IP range purpose (optional)"
               type="text"
             />
           </Col>
--- a/ui/src/app/subnets/components/ReservedRanges/ReservedRanges.test.tsx
+++ b/ui/src/app/subnets/components/ReservedRanges/ReservedRanges.test.tsx
@@ -364,3 +364,38 @@ it("can display an add form", async () => {
     screen.getByRole("form", { name: ReservedRangeFormLabels.CreateRange })
   ).toBeInTheDocument();
 });
+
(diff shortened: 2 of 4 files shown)


ui/src/app/subnets/components/ReservedRangeForm/ReservedRangeForm.tsx
```diff
@@ -142,7 +142,7 @@ const ReservedRangeForm = ({
               disabled={showDynamicComment}
               label={Labels.Comment}
               name="comment"
-              placeholder="IP range comment (optional)"
+              placeholder="IP range purpose (optional)"
```
huwshimi, line 145: Drive-by revert to the angular version.

@petermakowski sorry I might not have been very clear last night. I think it's hard in these cases to know whether this sort of inconsistency has crept in over time or if the column name and form label are deliberately different.

In this case "comment" is used in the CLI when creating ranges:
```
maas $PROFILE ipranges create type=dynamic \
    start_ip=$IP_DYNAMIC_RANGE_LOW end_ip=$IP_DYNAMIC_RANGE_HIGH \
    comment='This is a reserved dynamic range'
```
So the column could be labelled the same for consistency when users are switching between the two, but a designer could have discovered that when users are entering the comment in the web form that they're not sure what kind of comment to leave.

So this is the dilemma we face, we can ask our current designers what they think, but it's good to give them as much context as possible as they might not have the same insight as the original designer.

Caleb-Ellis (approved):
LGTM, but I would be hesitant to suggest we use a nebulous previous designer to inform our design decisions today. Without any info on why the legacy designs are the way they are I think we might end up being a little bit overly cautious.

Having said that, I think we should at least make the table header and form label consistent. And with nothing else to go on, at least "Comment" is consistent with the CLI (despite being an entirely unhelpful descriptor).

huwshimi:
> LGTM, but I would be hesitant to suggest we use a nebulous previous designer to inform our design decisions today. Without any info on why the legacy designs are the way they are I think we might end up being a little bit overly cautious.

On the other hand we're nebulous developers who might not have the context of why something was designed in a certain way (presuming this was a conscious choice and not an inconsistency that crept in).

PR #3617, opened 2022-02-11
fix(ui): minor improvements to subnet Used IP table

### LINKED ISSUE #1286: IPv6 leases UI is unreadable (Subnet details)
On my ultrawide display here (2560x1080) with a window configured to take about half of that width, I'm getting all IPv6 addresses cropped in a way that only shows me the network part and not the address part, making the list completely useless :)

https://bugs.launchpad.net/maas/+bug/1881803

Maybe we can use the double row pattern we use in other places to make room for super long IPs. At the very least a tooltip on hover would help.

Caleb-Ellis:
This branch turns the IP node text into links and also handles
extra cases for IP usage.

## Done

- Updated subnet "Used IP addresses" table to include:
  - Rendering the node as link, if it exists
  - Extra cases for IP usage 
  - Make "IP address" column text wrap if it gets long

### MAAS deployment

- [Bolla](/HACKING.md#bolla)

### Running the branch

You can run this branch by:

### QA steps

- Go to the details page of a subnet with used IP addresses that are being used for a node
- Check that the "Node" column is rendered as a link
- Open the console and edit the "IP address" text to be super long and check that it wraps to the next line instead of truncating

## Fixes

Fixes #1286


ui/src/app/store/subnet/types/enum.ts
```diff
@@ -6,14 +6,6 @@ export enum IPAddressType {
...
 }
 
-export enum IPAddressTypeLabel {
-  AUTO = "Automatic",
-  STICKY = "Sticky",
-  USER_RESERVED = "User reserved",
-  DHCP = "DHCP",
-  DISCOVERED = "Discovered",
-}
-
```
Caleb-Ellis, line 16: @andesol Instead of using a second enum for getting the display text, I've used a util instead. This is a bit of a preference thing, but enums are pretty awkward to use since you can't use computed values for their keys. i.e. it would be good if we could do:
``` ts
export enum IPAddressTypeLabel {
  [IPAddressType.AUTO] = "Automatic",
  [IPAddressType.STICKY] = "Sticky",
  ...
}
```

But since we can't, I generally prefer to use [const assertions](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-3-4.html#const-assertions) or in this case just a separate util.

ui/src/app/subnets/views/SubnetDetails/UsedIPs/SubnetUsedIPs.tsx
```diff
@@ -1,151 +0,0 @@
...
-  Node = "Node",
-  Interface = "interface",
-  Usage = "Usage",
-  Owner = "Owner",
-  LastSeen = "Last Seen",
-}
-
-const getType = (ip: SubnetIP) => {
-  return IPAddressTypeLabel[
-    IPAddressType[ip.alloc_type] as keyof typeof IPAddressTypeLabel
```
Caleb-Ellis, line 32: @andesol as mentioned above, using an enum for the labels means we have to explicitly cast the type here, which I almost always try to avoid.

PR #3634, opened 2022-02-22
feat(ui): add hidden columns dropdown to machines table

petermakowski:
## Done

- enable hidden columns dropdown to machines table (implemented in https://github.com/canonical-web-and-design/maas-ui/pull/3641)
- hidden columns are persisted to local storage
 
## Screenshots
Kapture 2022-02-22 at 11 53 06

## Background
It has been raised by our users that customising the machines table view by hiding certain columns would be very useful.  I decided to create a proof of concept PR as it's a very easy change and will have immediate positive impact.

> most of the time I don’t want to see how many cores, RAM, disks, storage, arch a machine has, because I know the hardware by its name, zone or pool. 
[MAAS UI prototype post](https://discourse.maas.io/t/maas-ui-the-new-maas-layout-and-react-sprint-migration/4705/5?u=petermakowski)

### MAAS deployment

- [Bolla](/HACKING.md#bolla)

### Running the branch

You can run this branch by:

### QA steps

- Steps for QA.

## Fixes

Fixes: # .

## Launchpad issue

## Backports

### DIFF
--- a/integration/cypress/integration/machines/list.spec.ts
+++ b/integration/cypress/integration/machines/list.spec.ts
@@ -18,7 +18,7 @@ context("Machine listing", () => {
     );
   });
 
-  it.skip("can hide machine table columns", () => {
+  it("can hide machine table columns", () => {
     cy.findAllByRole("columnheader").should("have.length", 8);
 
     cy.findAllByRole("button", { name: "Hidden columns" }).click();
--- a/ui/src/app/machines/views/MachineList/MachineList.tsx
+++ b/ui/src/app/machines/views/MachineList/MachineList.tsx
@@ -44,6 +44,19 @@ const MachineList = ({
     "hiddenGroups",
     []
   );
+  const [hiddenColumns, setHiddenColumns] = useStorageState<string[]>(
+    localStorage,
+    "hiddenColumns",
+    []
+  );
+
+  const toggleHiddenColumn = (column: string): void => {
+    if (hiddenColumns.includes(column)) {
+      setHiddenColumns(hiddenColumns.filter((c) => c !== column));
+    } else {
+      setHiddenColumns([...hiddenColumns, column]);
+    }
+  };
 
   useEffect(() => {
     dispatch(tagActions.fetch());
@@ -70,6 +83,8 @@ const MachineList = ({
         </Notification>
       ) : null}
       <MachineListControls
+        hiddenColumns={hiddenColumns}
+        toggleHiddenColumn={toggleHiddenColumn}
         filter={searchFilter}
         grouping={grouping}
         setFilter={setSearchFilter}
@@ -77,6 +92,7 @@ const MachineList = ({
         setHiddenGroups={setHiddenGroups}
       />
       <MachineListTable
+        hiddenColumns={hiddenColumns}
         filter={searchFilter}
         grouping={grouping}
         hiddenGroups={hiddenGroups}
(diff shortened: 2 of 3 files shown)


petermakowski:
Would appreciate feedback on this when you find some time @amylily1011 / @rayito 🙏

amylily1011:
Love it, love it, love it! :shipit:   😂

rayito:
It is indeed a very useful feature. UX-wise, however, we might need to think a bit more about it. The first thing that comes to my mind is that it's easy to miss that some columns are hidden. The dropdown says "Hidden columns (#)" but that might not be enough, we can't assume users are going to check that control (in fact, most likely they won't even look at that control section on top of the table).

So maybe we need some extra visual feedback.

petermakowski:
> it's easy to miss that some columns are hidden

That's a good point. On the other hand, nothing's hidden by default. A user needs to discover and interact with the dropdown first in order to hide anything. Once they did that it seems unlikely they'll miss it.

It's also worth mentioning that the "Grouping" select works in a similar fashion. The only visual indicator of a currently selected option is different text in the select button:

image
image

Caleb-Ellis:
This is great @petermakowski! I hope users will find it helpful. :)

I think the general functionality is good but there are a few visual quirks I think should be addressed before we consider merging. Note that not all of these issues were necessarily introduced by this PR, but rather are just more visible.

#### Filter dropdown


The filter dropdown has reduced horizontal space, which to me looks a bit *too* tight. It's worse on certain breakpoints. We might need to use a custom grid here as opposed to the stock Vanilla columns.

#### Hidden columns dropdown


I think it'd be nicer if we used formatted language here instead of the internal key strings, e.g. "Owner & Tags" instead of "owner"

#### Group count


The group count has been moved to the right of the cell, and overlaps a bit with the toggle button.

#### FQDN column


If only the FQDN column is left, the IP address row moves to the right.

#### Last column


It seems the last column is always right-aligned, even when it shouldn't be. This might be related to the above.

#### Row heights


When every column is visible I think every row is guaranteed to have 2 rows of data. With some columns hidden this isn't the case, leading to different row heights which to me looks a bit awkward.

petermakowski:
Thanks @Caleb-Ellis appreciate you took a good look at this, all comments were spot on. Made all changes that you suggested. Screenshot after applying fixes below.

**Filter dropdown**
Filters
Filters

**FQDN column**
10 1 238 4 (PXE) (+1)

**Last column**
**Group count**
Pasted Graphic 17

**Row heights**
10 1 238 4 (PXE) (+1)

petermakowski:
**Hidden columns dropdown**
image

amylily1011:
Can we ship this as a demo and not in production? Because it would be nice to discuss this in the Engineering sprint. (Sorry, for misleading earlier when I said ship it, I didn't mean to add this to MAAS 3.2 because it's not part of the deliverables, but I would like to bring this in the FF discussion. 😃 ) 
For the production version, we will need to redo the design to fit in with the new MAAS layout anyway.

petermakowski:
> Can we ship this as a demo and not in production? Because it would be nice to discuss this in the Engineering sprint. (Sorry, for misleading earlier when I said ship it, I didn't mean to add this to MAAS 3.2 because it's not part of the deliverables, but I would like to bring this in the FF discussion. 😃 ) For the production version, we will need to redo the design to fit in with the new MAAS layout anyway.

We're trying to avoid the complexity of introducing extra variations in general.

For that reason I think it's best to merge the underlying changes from this PR allowing for hiding columns (so de facto no change for the user), and keeping the extra "Hidden columns" dropdown option on a separate branch/in a separate PR until we decide how to proceed.

amylily1011:
Let's look at this when we start working on MAAS new layout.

cristinadresch:
@rayito @amylily1011 @petermakowski to review the modular table spec and ensure this solution is added there

PR #3635, opened 2022-02-23
Set up sitespeed.io CI job.

huwshimi:
## Done

- Add a GitHub action to run sitespeed on merge.

### QA steps

- Check that the action artefact has the sitespeed results. Here's a passing run (I temporarily changed the action to run on this PR): https://github.com/canonical-web-and-design/maas-ui/actions/runs/2567167138

## Fixes

Fixes: https://github.com/canonical-web-and-design/app-tribe/issues/1076.

### DIFF
--- a/.github/workflows/sitespeed.yml
+++ b/.github/workflows/sitespeed.yml
@@ -1,0 +1,47 @@
+name: sitespeed.io
+on:
+  pull_request:
+    types: [closed]
+jobs:
+  run-sitespeed:
+    if: github.event.pull_request.merged == true
+    name: Run sitespeed.io
+    runs-on: ubuntu-20.04
+    env:
+      MAAS_DOMAIN: localhost
+    steps:
+      - uses: actions/checkout@main
+      - name: Install MAAS
+        run: |
+          sudo systemctl enable snapd
+          sudo snap install maas-test-db --channel=latest/edge
+          sudo snap install maas --channel=latest/edge
+      - name: Fetch database dump
+        uses: wei/wget@v1
+        with:
+          args: -O maasdb.dump https://github.com/canonical-web-and-design/maas-ui-testing/raw/main/db/maasdb-20.04-master-1000.dump
+      - name: Set up MAAS with database dump
+        run: |
+          sudo sed -i "s/dynamic_shared_memory_type = posix/dynamic_shared_memory_type = sysv/" /var/snap/maas-test-db/common/postgres/data/postgresql.conf
+          sudo snap restart maas-test-db
+          sudo mv maasdb.dump /var/snap/maas-test-db/common/maasdb.dump
+          sudo snap run --shell maas-test-db.psql -c 'db-dump restore /var/snap/maas-test-db/common/maasdb.dump maassampledata'
+          sudo maas init region+rack --maas-url=http://${{env.MAAS_DOMAIN}}:5240/MAAS --database-uri maas-test-db:///
+          sudo sed -i "s/database_name: maasdb/database_name: maassampledata/" /var/snap/maas/current/regiond.conf
+          sudo snap restart maas
+      - name: Create MAAS admin
+        run: sudo maas createadmin --username=admin --password=test --email=fake@example.org
+      - name: Wait for MAAS
+        uses: nev7n/wait_for_response@v1
+        with:
+          url: "http://${{env.MAAS_DOMAIN}}:5240/MAAS/r"
+          responseCode: 200
+          timeout: 200000
+          interval: 500
+      - name: Run sitespeed.io tests
+        run: yarn sitespeed --browsertime.domain=${{env.MAAS_DOMAIN}}
+      - name: Upload results
+        uses: actions/upload-artifact@v2
+        with:
+          name: sitespeed.io-results
+          path: sitespeed.io/results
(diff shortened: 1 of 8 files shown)


huwshimi:
On hold until we can use sampledata to set up machines.

Caleb-Ellis:
I think this is unblocked now. Instead of generating new sample data every time we run CI, the core team have a few Jenkins jobs for creating a db dump of 10, 100 and 1000 sample machines ([link](http://maas-ci.internal:8080/view/maas-sampledata-dumper-master/)), which should be able to be downloaded and imported into a local PostgreSQL server.

With `maas-test-db`, we should be able to do:
``` bash
sudo cp maasdb.dump /var/snap/maas-test-db/common
sudo snap run --shell maas-test-db.psql -c 'db-dump restore $SNAP_COMMON/maasdb.dump maassampledata'
```

huwshimi:
@Caleb-Ellis @petermakowski I've got this is working with the sample data, however the database dump with 1000 machines is about 8mb. Where do you think we should store that? Maybe on the asset server?

huwshimi:
This is ready for review now!

petermakowski:
> This is ready for review now!

The detailed summary in the report indicates there were no JS, CSS or font files loaded. Either something is not right or I don't understand how it's supposed to work.

image

.github/workflows/sitespeed.yml
```diff
@@ -0,0 +1,47 @@
+name: sitespeed.io
+on:
+  pull_request:
+    types: [closed]
```
petermakowski, line 4: Could we run this action on opened PRs and identify performance issues before changes are merged?

.github/workflows/sitespeed.yml
```diff
@@ -0,0 +1,47 @@
...
+      - uses: actions/checkout@main
+      - name: Install MAAS
+        run: |
+          sudo systemctl enable snapd
+          sudo snap install maas-test-db --channel=latest/edge
+          sudo snap install maas --channel=latest/edge
+      - name: Fetch database dump
+        uses: wei/wget@v1
+        with:
+          args: -O maasdb.dump https://github.com/canonical-web-and-design/maas-ui-testing/raw/main/db/maasdb-20.04-master-1000.dump
```
petermakowski, line 22: Does this dump need to be updated manually or do we/could we automate this?

docs/HACKING.md
```diff
@@ -600,3 +601,24 @@ _Note: this only limits the tests in a single file. You will also need to make s
...
+PRs are merged.
+
+Sitespeed can also be run manually, though the tests expect a MAAS with a
+specific dataset. For best results a [local MAAS](#local-deployments) can be set
+up using [sample data](#sample-data).
+
+To run against a MAAS deployment you can use:
+
+```shell
+yarn sitespeed --browsertime.domain=[maas.ip.or.hostname]
```
petermakowski, line 617: Would be great if we could just read the hostname and port values from the `MAAS_URL` env variable, we wouldn't need this extensive readme and differentiation between local and remote MAAS deployments.

sitespeed.io/login.js
```diff
@@ -0,0 +1,24 @@
...
+    name: "skipsetupintro",
+    value: "true",
+  });
+  await commands.cdp.send("Network.setCookie", {
+    domain: context.options.domain,
+    name: "skipintro",
+    value: "true",
+  });
+  await commands.navigate(constructURL(context, "/dashboard"));
+  await commands.wait.bySelector("input[name='username']", 5000);
```
petermakowski, line 15: This is nitpicking, but I wonder why you decided not to abstract this to a variable as it's used in 2 places, but did in the other file where a timeout is only used in one place.

sitespeed.io/scripts/machines.js
```diff
@@ -0,0 +1,10 @@
+const { constructURL } = require("../utils");
+
+const TIMEOUT = 120000;
```
petermakowski, line 3: This is only used in 1 place. Was your intention to make this clearer and improve the readability?

sitespeed.io/scripts/machines.js
```diff
@@ -0,0 +1,10 @@
...
+
+module.exports = async (context, commands) => {
+  await commands.measure.start("Machine list");
+  await commands.navigate(constructURL(context, "/machines"));
+  await commands.wait.byXpath("//*[text()='1000 Machines']", TIMEOUT);
+  return commands.measure.stop();
```
petermakowski, line 9: If I understand correctly, we're waiting an arbitrary amount of time before stopping the measurement which does not guarantee we're measuring the entire page load cycle (until it's fully loaded and interactive). 2 minutes is probably sufficient, but could you think of any other ways of checking the page has fully loaded?

petermakowski (commented):
This looks good overall, I just have a few questions/comments. Would be also good to solve conflicts and verify the report is correct before merge (comment above).

huwshimi (commented):
I figured out the caching issue. By default it clears the cache between tests, [but not when using a preScript](https://www.sitespeed.io/documentation/sitespeed.io/best-practice/#clear-browser-cache-between-runs) (which we use for logging in). I've updated the test to clear the cache before the test and also added a second test after caching.

It's looking better now:

Screen Shot 2022-07-01 at 1 51 49 pm

sitespeed.io/scripts/machines.js
petermakowski: Oooh, I missed that "small" detail. This should be perfectly fine then.

petermakowski (approved):

PR #3637, opened 2022-02-23
fix(ui): fix KVM config form disabled submit button after initial submit

### LINKED ISSUE #2673: Can't change VMhost overcommit back without a reload
Bug originally filed by ack at https://bugs.launchpad.net/bugs/1928792

From a VMhost (LXD or virsh) settings, if I change the CPU overcommit slider from 1 to something else, save, and then try to change it back, the save button gets disabled as soon as 1 is selected as value, so I can't save it again.

Reloading the page fixes the issue.

This seems to only happen if you change just the overcommit.

Note that the issue also happens with the memory slider, but for some reason it seems harder to reproduce (not all time times as the CPU one)

Caleb-Ellis:
## Done

- Allow changing KVM values to initial after first submit
- Rewrite KVM config tests to use RTL

### MAAS deployment

- [Bolla](/HACKING.md#bolla)

### Running the branch

You can run this branch by:

### QA steps

- Go to the config page of a KVM and check that the submit button is disabled
- Change a value (like CPU overcommit) and check that the submit button becomes enabled
- Submit the form
- After success, change the value back to the initial value and check that the submit button is still enabled

## Fixes

Fixes #2673

### DIFF
--- a/ui/src/app/kvm/components/KVMConfigurationCard/KVMConfigurationCard.test.tsx
+++ b/ui/src/app/kvm/components/KVMConfigurationCard/KVMConfigurationCard.test.tsx
@@ -55,24 +55,22 @@ it("can handle updating a lxd KVM", async () => {
     </Provider>
   );
 
-  await waitFor(() => {
-    fireEvent.change(screen.getByRole("combobox", { name: "Zone" }), {
-      target: { value: "3" },
-    });
-    fireEvent.change(screen.getByRole("combobox", { name: "Resource pool" }), {
-      target: { value: "2" },
-    });
-    fireEvent.change(screen.getByRole("slider", { name: "CPU overcommit" }), {
-      target: { value: "5" },
-    });
-    fireEvent.change(
-      screen.getByRole("slider", { name: "Memory overcommit" }),
-      { target: { value: "7" } }
-    );
+  fireEvent.change(screen.getByRole("combobox", { name: "Zone" }), {
+    target: { value: "3" },
+  });
+  fireEvent.change(screen.getByRole("combobox", { name: "Resource pool" }), {
+    target: { value: "2" },
+  });
+  fireEvent.change(screen.getByRole("slider", { name: "CPU overcommit" }), {
+    target: { value: "5" },
+  });
+  fireEvent.change(screen.getByRole("slider", { name: "Memory overcommit" }), {
+    target: { value: "7" },
   });
   await waitFor(() => {
-    fireEvent.click(screen.getByRole("button", { name: "Save changes" }));
+    expect(screen.getByRole("button", { name: "Save changes" })).toBeEnabled();
   });
+  fireEvent.click(screen.getByRole("button", { name: "Save changes" }));
 
   const expectedAction = podActions.update({
     cpu_over_commit_ratio: 5,
@@ -84,9 +82,11 @@ it("can handle updating a lxd KVM", async () => {
     type: PodType.LXD,
     zone: 3,
   });
-  expect(
-    store.getActions().find((action) => action.type === expectedAction.type)
-  ).toStrictEqual(expectedAction);
+  await waitFor(() => {
+    expect(
+      store.getActions().find((action) => action.type === expectedAction.type)
+    ).toStrictEqual(expectedAction);
+  });
 });
 
 it("can handle updating a virsh KVM", async () => {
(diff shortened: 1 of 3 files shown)


ui/src/app/kvm/components/KVMConfigurationCard/KVMConfigurationCard.test.tsx
```diff
@@ -1,126 +1,202 @@
...
+    <Provider store={store}>
+      <MemoryRouter
+        initialEntries={[{ pathname: "/kvm/1/edit", key: "testKey" }]}
+      >
+        <KVMConfigurationCard pod={pod} />
+      </MemoryRouter>
+    </Provider>
+  );
+
+  await waitFor(() => {
```
petermakowski, line 109: These `waitFor` seem arbitrary. It's usually be better to be explicit (and avoid focusing implementation details) as much as possible.

If you're waiting for the button to be enabled after interacting with the form inputs, it might be better to use something along the lines of:

```jsx
await waitFor(() => expect(screen.getByRole("button", { name: "Save changes" }).toBeEnabled())
```

Will also make this test easier to read and help other devs understand the expected behaviour.

ui/src/app/kvm/components/KVMConfigurationCard/KVMConfigurationCard.test.tsx
```diff
@@ -1,126 +1,202 @@
...
-          initialEntries={[{ pathname: "/kvm/1/edit", key: "testKey" }]}
-        >
-          <KVMConfigurationCard pod={pod} />
-        </MemoryRouter>
-      </Provider>
-    );
+  // Submit should be enabled.
+  expect(
+    screen.getByRole("button", { name: "Save changes" })
+  ).not.toBeDisabled();
```
petermakowski, line 177: Similar to the comment above, `waitFor` wrapper should probably be used around this expect instead of around fireEvent.

ui/src/app/kvm/components/KVMConfigurationCard/KVMConfigurationCard.test.tsx
```diff
@@ -1,126 +1,202 @@
...
+  await waitFor(() => {
+    fireEvent.change(screen.getByRole("slider", { name: "CPU overcommit" }), {
+      target: { value: pod.cpu_over_commit_ratio.toString() },
     });
   });
+
+  // Submit should still be enabled.
+  expect(
+    screen.getByRole("button", { name: "Save changes" })
+  ).not.toBeDisabled();
```
petermakowski, line 201: As above.

petermakowski (approved):
Appreciate refactoring to @testing-library 👍 
A few minor comments below, but other than that looks good!

ui/src/app/kvm/components/KVMConfigurationCard/KVMConfigurationCard.tsx
```diff
@@ -42,6 +47,7 @@ type Props = {
 };
 
 const KVMConfigurationCard = ({ pod, zoneDisabled }: Props): JSX.Element => {
+  const [hasUpdated, setHasUpdated] = useState(false);
```
petermakowski, line 50: **QA note**
This results in different behaviour to the same interactions across time.

Before first submission saving an unchanged (equal to what's saved on the back-end) value is disallowed, but allowed on every subsequent change.

We should probably be updating the initial value on after submit to keep the same behaviour across interactions. I'm okay to leave this as is If it happens to be too complex to fix, but would be good to change.

ui/src/app/kvm/components/KVMConfigurationCard/KVMConfigurationCard.test.tsx
Caleb-Ellis: Good point, thanks!

ui/src/app/kvm/components/KVMConfigurationCard/KVMConfigurationCard.tsx
Caleb-Ellis: Yeah I ended up taking the "easy" route to avoid the complexity of having to track when the underlying `pod` object updates. Figured the behaviour was inconsistent but at least it wasn't buggy. But I actually forgot that `Formik` provides an `enableReinitialize` prop that solves this problem entirely, so that whenever `initialValues` changes the form is reset. Whoops!

Caleb-Ellis:
@petermakowski alright I've updated the PR to use `enableReinitialize` so the behaviour should be consistent now.

ui/src/app/kvm/components/KVMConfigurationCard/KVMConfigurationCard.tsx
```diff
@@ -52,6 +57,7 @@ const KVMConfigurationCard = ({ pod, zoneDisabled }: Props): JSX.Element => {
     <FormCard highlighted={false} sidebar={false} title="KVM configuration">
       <FormikForm<KVMConfigurationValues>
         cleanup={cleanup}
+        enableReinitialize
```
petermakowski, line 60: 🎉

petermakowski (approved):
Much better. Thanks!

ui/src/app/kvm/components/KVMConfigurationCard/KVMConfigurationCard.test.tsx
petermakowski: Opened a PR to add eslint-plugin-testing-library https://github.com/canonical-web-and-design/maas-ui/pull/3644 which will give a warning for such a use of `waitFor`

PR #3647, opened 2022-02-25
refactor(ui): upgrade to Vanilla 3

huwshimi:
## Done

- Update Vanilla and react-components to Vanilla 3.

### MAAS deployment

- [Bolla](/HACKING.md#bolla)

### Running the branch

You can run this branch by:

### QA steps

- Run the full client with `yarn start`.
- Look around the site at different browser sizes and make sure nothing is broken (that wasn't already broken).
- Note: [this issue](https://github.com/canonical-web-and-design/react-components/issues/716) means that the mixed checkbox state is not appearing in the machine list.


legacy/package.json
```diff
@@ -86,7 +86,7 @@
     "js-cookie": "3.0.1",
     "ng-tags-input": "3.2.0",
     "single-spa-angularjs": "4.3.1",
-    "vanilla-framework": "2.37.1"
+    "vanilla-framework": "3.0.0-rc.1"
```
petermakowski, line 89: We should use the latest stable instead of a release candidate, but this can be done in another PR.

huwshimi: Oh no, thanks. It only affected the controller details and the initial loading screen, so I've upgraded to the release versions for both and it looks fine to me.

huwshimi:
@petermakowski thanks for the review! I've upgraded to the release version for legacy/root and it looks OK to me, thanks for catching that. If you're happy feel free to land this so we can upgrade to the latest react-components.

legacy/package.json
petermakowski: Ok great, even better. Thanks!

PR #3689, opened 2022-03-18
feat(ui): display manual and automatic tags in the tag form

huwshimi:
## Done

- Add the existing manual and automatic tags on the selected machines in the tag form in the machine list.

### MAAS deployment

- [Bolla](/HACKING.md#bolla)

### Running the branch

You can run this branch by:

### QA steps

- Go to the machine list and select some machines. I've created an automatic tag so if you select all machines, or one of the machines attached to the tag (http://bolla.internal:5240/MAAS/r/tag/74) you'll see automatic tags in the form.
- Use the take action menu to open the tag form.
- You should see the existing tags on the machines you selected.

## Fixes

Fixes: canonical-web-and-design/app-tribe#685.

## Screenshots

Screen Shot 2022-03-18 at 1 04 29 pm

### DIFF
--- a/ui/src/app/store/machine/utils/common.ts
+++ b/ui/src/app/store/machine/utils/common.ts
@@ -1,4 +1,5 @@
 import type { Machine, MachineDetails } from "app/store/machine/types";
+import type { Tag, TagMeta } from "app/store/tag/types";
 
 /**
  * Whether a machine has a Machine or MachineDetails type.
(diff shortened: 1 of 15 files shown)


ui/src/app/store/machine/utils/common.ts
```diff
@@ -9,3 +10,24 @@ export const isMachineDetails = (
...
+export const getTagCountsForMachines = (machines: Machine[]): TagIdCountMap => {
+  const ids = machines.reduce<Tag[TagMeta.PK][]>(
+    (tagIds, machine) => tagIds.concat(machine.tags),
+    []
+  );
+  const tagCounts = new Map();
+  ids.forEach((id) => {
+    if (!tagCounts.has(id)) {
+      tagCounts.set(id, ids.filter((tagId) => tagId === id).length);
+    }
```
Caleb-Ellis, line 30: It seems a bit inefficient to filter the entire tag ID list on each new entry. Would a simple conditional work here?

``` ts
tagCounts.set(id, (tagCounts.get(id) || 0) + 1);
```

Caleb-Ellis (approved):
LGTM! Just one minor comment

ui/src/app/store/machine/utils/common.ts
Caleb-Ellis: Then again I dunno if continuously updating a map entry is a more expensive operation, so maybe the point is moot

huwshimi: I just ran both versions using the profiler and with my dataset on my machine the `+ 1` version was on average 0.1ms slower, but we're talking an imperceptibly small amount of difference. It's possible a different data set could return different results, but I guess we can revisit this if we're noticing performance issues with large amounts of data.

PR #3696, opened 2022-03-21
feat(ui): delete tags

huwshimi:
## Done

- Delete tags from the tag list or tag details.

### MAAS deployment

- [Bolla](/HACKING.md#bolla)

### Running the branch

You can run this branch by:

### QA steps

- Visit the tag list.
- Click delete on one of the tag rows.
- The delete form should appear in the header.
- There should be warnings if it has kernel options or if it is applied to any nodes.
- Confirm and the tag should be removed.
- Go to the details page for a tag.
- Click delete and the form should appear.

## Fixes

Fixes: canonical-web-and-design/app-tribe#701. Fixes: canonical-web-and-design/app-tribe#700.

## Screenshot

Screen Shot 2022-03-22 at 10 15 40 am

### DIFF
--- a/ui/src/app/tags/components/TagsHeader/DeleteTagForm/DeleteTagForm.test.tsx
+++ b/ui/src/app/tags/components/TagsHeader/DeleteTagForm/DeleteTagForm.test.tsx
@@ -16,6 +16,7 @@ import {
 const mockStore = configureStore();
 
 let state: RootState;
+let scrollToSpy: jest.Mock;
 
 beforeEach(() => {
   state = rootStateFactory({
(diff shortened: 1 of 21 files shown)


Caleb-Ellis (approved):
LGTM! Just some minor things that could be handled in a follow-up:
- The design for the warning only shows the yellow strip if there is more than one message.
- When deleting a tag with kernel options that's assigned to 1 machine, it says "There are 1 machine" instead of "There is 1 machine"
- We should use the scroll to top hook for the tag header forms too, otherwise it seems like nothing happens when you click the delete button on a tag far down the page.

huwshimi:
> LGTM! Just some minor things that could be handled in a follow-up:
> 
> * The design for the warning only shows the yellow strip if there is more than one message.

I think that's just a design inconsistency, the grouped message design was added later.

> * When deleting a tag with kernel options that's assigned to 1 machine, it says "There are 1 machine" instead of "There is 1 machine"

Oops, thanks.

> * We should use the scroll to top hook for the tag header forms too, otherwise it seems like nothing happens when you click the delete button on a tag far down the page.

Thanks, I'll add that.

PR #3739, opened 2022-03-31
feat(ui): add tags to machines

huwshimi:
## Done

- Display added tags in the machine list tag form.

### MAAS deployment

- [Bolla](/HACKING.md#bolla)

### Running the branch

You can run this branch by:

### QA steps

- Visit the machine list and select some machines and open the tag form.
- Click on the text input and choose some tag.
- The chosen tags should appear as green items in the change list.
- Clicking Dismiss should remove them.
- Submit the form and then reopen the tag form and the newly added tags should appear in the currently assigned tags list (blue).

## Fixes

Fixes: canonical-web-and-design/app-tribe#688.

## Screenshot

Screen Shot 2022-03-31 at 12 57 34 pm

### DIFF
--- a/ui/src/app/base/components/TagSelector/TagSelector.tsx
+++ b/ui/src/app/base/components/TagSelector/TagSelector.tsx
@@ -16,6 +16,7 @@ export type Props = {
   allowNewTags?: boolean;
   disabled?: boolean;
   error?: string;
+  externalSelectedTags?: Tag[];
   help?: string;
   initialSelected?: Tag[];
   label?: string | null;
@@ -25,6 +26,7 @@ export type Props = {
   generateDropdownEntry?: (tag: Tag, highlightedName: ReactNode) => ReactNode;
   header?: ReactNode;
   showSelectedTags?: boolean;
+  useExternalTags?: boolean;
   tags: Tag[];
   disabledTags?: Tag[];
 };
@@ -174,6 +176,7 @@ export const TagSelector = ({
   allowNewTags = false,
   disabled,
   error,
+  externalSelectedTags,
   generateDropdownEntry,
   help,
   initialSelected = [],
@@ -184,13 +187,18 @@ export const TagSelector = ({
   header,
   showSelectedTags = true,
   tags = [],
+  useExternalTags = false,
   disabledTags = [],
   ...props
 }: Props): JSX.Element => {
   const wrapperRef = useRef<HTMLSpanElement | null>(null);
 
   const [dropdownOpen, setDropdownOpen] = useState(false);
-  const [selectedTags, setSelectedTags] = useState(initialSelected);
+  const [internalSelectedTags, setInternalSelectedTags] =
+    useState(initialSelected);
+  const selectedTags = useExternalTags
+    ? externalSelectedTags || initialSelected
+    : internalSelectedTags;
   const [filter, setFilter] = useState("");
   const hasSelectedTags = showSelectedTags && selectedTags.length > 0;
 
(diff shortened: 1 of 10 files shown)


ui/src/app/machines/components/MachineHeaderForms/ActionFormWrapper/TagForm/hooks.ts
```diff
@@ -0,0 +1,29 @@
...
+  const selectedTags = useSelector((state: RootState) =>
+    tagSelectors.getByIDs(
+      state,
+      values.tags.reduce<Tag[TagMeta.PK][]>((tagList, id) => {
+        const idNumber = toFormikNumber(id);
+        if (idNumber) {
+          tagList.push(idNumber);
+        }
+        return tagList;
+      }, [])
```
Caleb-Ellis, line 25: This could probably be split out into its own variable, seems a bit muddled in the useSelector hook

ui/src/app/machines/components/MachineHeaderForms/ActionFormWrapper/TagForm/TagFormFields/TagFormFields.tsx
```diff
@@ -19,11 +20,13 @@ type Props = {
 };
 
 export const TagFormFields = ({ machines }: Props): JSX.Element => {
+  const addedTags = useSelectedTags();
   const tags = useSelector(tagSelectors.getManual);
   return (
     <Row>
       <Col size={6} className="col-start-large-4">
         <TagField
+          externalSelectedTags={addedTags}
```
Caleb-Ellis, line 29: This is very confusing language. `addedTags = useSelectedTags()` `externalSelectedTags={addedTags}`

Could it all be `selectedTags`?

ui/src/app/base/components/TagSelector/TagSelector.tsx
```diff
@@ -184,21 +187,28 @@ export const TagSelector = ({
   header,
   showSelectedTags = true,
   tags = [],
+  useExternalTags = false,
```
Caleb-Ellis, line 190: Could this just be determined by `externalSelectedTags.length > 0`?

Caleb-Ellis (approved):
Everything works fine, but I think once this feature is complete we should really go over our current tag components and make them easier to use and understand. Now that `TagSelector` has an `externalSelectedTags` prop I'm wondering if it's handling too many concerns at once, and whether we might want to split the component up or streamline it somehow.

ui/src/app/machines/components/MachineHeaderForms/ActionFormWrapper/TagForm/hooks.ts
huwshimi: Good idea, I've also added a comment about why this is necessary.

ui/src/app/machines/components/MachineHeaderForms/ActionFormWrapper/TagForm/TagFormFields/TagFormFields.tsx
huwshimi: Yeah, that makes a lot more sense (I copied and pasted and didn't think about the names).

ui/src/app/base/components/TagSelector/TagSelector.tsx
huwshimi: No, because initially the `externalSelectedTags` array could be empty (which would result in setting both internal and external tag state). Though we could just check for a truthy value (with `[]` being truthy). So if the array exists then we use that state. I'll update that.

PR #3748, opened 2022-04-01
feat(ui): build read-only machine power config section

Caleb-Ellis:
## Done

- Built read-only machine power config section
  - This replaces the old functionality of it always being a form, with form fields being enabled depending on whether the user was in edit mode.
  - Added new components for the read-only section, with common utils being extracted from PowerTypeFields
- Removed "disabled" functionality for PowerTypeFields, since now this can't be done anywhere in the UI

### MAAS deployment

- [Bolla](/HACKING.md#bolla)

### Running the branch

You can run this branch by:

### QA steps

- Go to the config tab of a machine that is not in a pod
- Check that the power parameters render as a list and not as form fields
- Click "Edit" and check that the section is now a form
- Change the power type to LXD and choose to generate a certificate
- Submit the form and check that it closes after success, and the generated certificate is displayed with its metadata and a download button
- Go to the config tab of a LXD machine that is in a pod
- Check that there are only node-scoped fields displayed
- Click "Edit" and check that you do not get the option to change the certificate

## Fixes

Fixes canonical-web-and-design/app-tribe#745

## Screenshot
Peek 2022-04-01 13-58

### DIFF
--- a/ui/src/app/machines/views/MachineDetails/MachineConfiguration/PowerForm/PowerFormFields/PowerFormFields.tsx
+++ b/ui/src/app/machines/views/MachineDetails/MachineConfiguration/PowerForm/PowerFormFields/PowerFormFields.tsx
@@ -1,67 +1,31 @@
-import { Col, Notification, Row } from "@canonical/react-components";
-import { useFormikContext } from "formik";
-import { useSelector } from "react-redux";
+import { Col, Row } from "@canonical/react-components";
 
 import type { PowerFormValues } from "../PowerForm";
 
 import PowerTypeFields from "app/base/components/PowerTypeFields";
-import { useIsRackControllerConnected } from "app/base/hooks";
 import { PowerTypeNames } from "app/store/general/constants";
-import { powerTypes as powerTypesSelectors } from "app/store/general/selectors";
-import { getPowerTypeFromName } from "app/store/general/utils";
 import type { MachineDetails } from "app/store/machine/types";
 import { getMachineFieldScopes } from "app/store/machine/utils";
 
 type Props = {
-  editing: boolean;
   machine: MachineDetails;
 };
 
-const PowerFormFields = ({ editing, machine }: Props): JSX.Element => {
-  const powerTypes = useSelector(powerTypesSelectors.get);
-  const { values } = useFormikContext<PowerFormValues>();
-  const isRackControllerConnected = useIsRackControllerConnected();
-  const powerType = getPowerTypeFromName(powerTypes, values.powerType);
-  const machineInPod = Boolean(machine.pod);
+const PowerFormFields = ({ machine }: Props): JSX.Element => {
+  const isMachineInPod = Boolean(machine.pod);
   const fieldScopes = getMachineFieldScopes(machine);
 
   return (
     <Row>
       <Col size={6}>
-        {!isRackControllerConnected && (
-          <Notification data-testid="no-rack-controller" severity="negative">
-            Power configuration is currently disabled because no rack controller
-            is currently connected to the region.
-          </Notification>
-        )}
-        {isRackControllerConnected && !values.powerType && (
-          <Notification data-testid="no-power-type" severity="negative">
-            This node does not have a power type set and MAAS will be unable to
-            control it. Update the power information below.
-          </Notification>
-        )}
-        {values.powerType === "manual" && (
-          <Notification data-testid="manual-power-type" severity="caution">
-            Power control for this power type will need to be handled manually.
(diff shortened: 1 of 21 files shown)


Caleb-Ellis:
Waiting on #3749 to land first

ui/src/app/machines/views/MachineDetails/MachineConfiguration/PowerForm/PowerParameters/PowerParameterDefinition/PowerParameterDefinition.test.tsx
```diff
@@ -0,0 +1,56 @@
...
+    field_type: PowerFieldType.MULTIPLE_CHOICE,
+  });
+  render(
+    <PowerParameterDefinition
+      field={field}
+      powerParameter={["choice1", "choice2"]}
+    />
+  );
+
+  expect(screen.getByText(/Choice 1, Choice 2/)).toBeInTheDocument();
```
petermakowski, line 46: I wonder whether we should use a [Definition Term](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/dl) element in the underlying [Definition](https://github.com/canonical-web-and-design/maas-ui/blob/master/ui/src/app/base/components/Definition/Definition.tsx#L21) component.
I decided against it when I was working on the Definition component as a Definition Term has to be a part of a [Description List](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/dl) and it felt like an overkill in a lot of cases where we use it for just a single item.

But in this case it would make for better semantics (and a11y). It looks like a classic example for when Definition List could be used? I wonder what are your thoughts on this.

> Description lists are useful for displaying metadata as a list of key-value pairs.
https://developer.mozilla.org/en-US/docs/Web/HTML/Element/dl#metadata

ui/src/app/machines/views/MachineDetails/MachineConfiguration/PowerForm/PowerFormFields/PowerFormFields.tsx
```diff
@@ -1,67 +1,31 @@
...
 };
 
-const PowerFormFields = ({ editing, machine }: Props): JSX.Element => {
-  const powerTypes = useSelector(powerTypesSelectors.get);
-  const { values } = useFormikContext<PowerFormValues>();
-  const isRackControllerConnected = useIsRackControllerConnected();
-  const powerType = getPowerTypeFromName(powerTypes, values.powerType);
-  const machineInPod = Boolean(machine.pod);
+const PowerFormFields = ({ machine }: Props): JSX.Element => {
+  const machineInPod = !!machine.pod;
```
petermakowski, line 15: I wonder what made you decide to change the syntax from `Boolean` to double negation `!!`?

Also, do you think it'd be worth renaming this to include an `is` prefix, e.g. `isMachineInPod`.
That'd be consistent with `canEditCertificate`, `initialShouldGenerateCert` which both include a word indicating a boolean value.

petermakowski (approved):
LGTM

ui/src/app/machines/views/MachineDetails/MachineConfiguration/PowerForm/PowerParameters/PowerParameterDefinition/PowerParameterDefinition.test.tsx
Caleb-Ellis: Yeah that sounds like a good idea. Maybe we could upstream this to Vanilla itself? I feel like a component that displays key-value pairs (where the value could be a list) should be standardised. Regardless I've added an issue (#3751) to track it.

ui/src/app/machines/views/MachineDetails/MachineConfiguration/PowerForm/PowerFormFields/PowerFormFields.tsx
Caleb-Ellis: Good point! Updating now.

I actually prefer using `Boolean()` since it's clearer, but looks like it was a casualty in this rewrite where I reverted to using `!!` as shorthand, oops.

ui/src/app/machines/views/MachineDetails/MachineConfiguration/PowerForm/PowerParameters/PowerParameterDefinition/PowerParameterDefinition.test.tsx
petermakowski: Awesome, thanks @Caleb-Ellis

PR #3790, opened 2022-04-11
fix(ui): unset IP mode when removing subnet

huwshimi:
## Done

- Fix unsetting an interface's IP mode when the subnet is set to unconfigured.

### MAAS deployment

- [Bolla](/HACKING.md#bolla)

### Running the branch

You can run this branch by:

### QA steps

- Go to the network tab for a machine where the network is editable and has a subnet and a physical interface that is set to "Auto assign" (or update an interface to this state).
- Open the edit form for the interface.
- Open the Subnet select and change it to "Unconfigured".
- Save the form and the IP column should change to "Unconfigured".

## Fixes

Fixes: #3687.

### DIFF
--- a/ui/src/app/machines/views/MachineDetails/MachineNetwork/NetworkFields/NetworkFields.test.tsx
+++ b/ui/src/app/machines/views/MachineDetails/MachineNetwork/NetworkFields/NetworkFields.test.tsx
@@ -1,13 +1,25 @@
-import { mount } from "enzyme";
-import type { ReactWrapper } from "enzyme";
+import {
+  fireEvent,
+  render,
+  screen,
+  waitFor,
+  within,
+} from "@testing-library/react";
+import userEvent from "@testing-library/user-event";
 import { Formik } from "formik";
-import { act } from "react-dom/test-utils";
 import { Provider } from "react-redux";
 import { MemoryRouter } from "react-router-dom";
 import configureStore from "redux-mock-store";
 
-import NetworkFields from "./NetworkFields";
+import NetworkFields, {
+  Label as NetworkFieldsLabel,
+  networkFieldsInitialValues,
+} from "./NetworkFields";
 
+import { Label as FabricSelectLabel } from "app/base/components/FabricSelect/FabricSelect";
+import { Label as LinkModeSelectLabel } from "app/base/components/LinkModeSelect/LinkModeSelect";
+import { Label as SubnetSelectLabel } from "app/base/components/SubnetSelect/SubnetSelect";
+import { Label as VLANSelectLabel } from "app/base/components/VLANSelect/VLANSelect";
 import type { RootState } from "app/store/root/types";
 import { NetworkInterfaceTypes, NetworkLinkMode } from "app/store/types/enum";
 import {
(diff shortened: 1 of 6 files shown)


ui/src/app/machines/views/MachineDetails/MachineNetwork/NetworkFields/NetworkFields.test.tsx
```diff
@@ -27,31 +39,16 @@ import {
...
   let state: RootState;
 
   beforeEach(() => {
+    const vlan = vlanFactory({ fabric: 1, vid: 1 });
     state = rootStateFactory({
       fabric: fabricStateFactory({
         items: [
-          fabricFactory({ id: 1, default_vlan_id: 1 }),
-          fabricFactory({ default_vlan_id: 1 }),
+          fabricFactory({ id: 1, default_vlan_id: vlan.id }),
```
huwshimi, line 50: For some reason when I migrated this file to RTL it introduced a leak between the tests. I tracked it down to setting up the store in two concurrent tests but I couldn't figure out why the leak was occurring (the leak was happening even if the only thing in the "it(...)" block was creating the state).

I'd love to hear any suggestions about why that leak might be occurring. In the meantime, using unique ids model ids appears to make the tests pass.

ui/src/app/machines/views/MachineDetails/MachineNetwork/NetworkFields/NetworkFields.test.tsx
```diff
@@ -307,77 +414,158 @@ describe("NetworkFields", () => {
...
+      {
+        target: { value: "" },
+      }
+    );
+    expect(
+      screen.queryByRole("combobox", { name: LinkModeSelectLabel.Select })
+    ).not.toBeInTheDocument();
+    fireEvent.submit(screen.getByRole("form", { name: "test form" }));
+    await waitFor(() =>
+      expect(onSubmit.mock.calls[0][0].mode).toBe(NetworkLinkMode.LINK_UP)
```
huwshimi, line 475: Apologies for the noise (I regret doing the RTL migration in this PR). This test is the only change in this file that is not a migration.

ui/src/app/machines/views/MachineDetails/MachineNetwork/NetworkFields/NetworkFields.tsx
```diff
@@ -70,13 +74,17 @@ const NetworkFields = ({
...
+    hasSubnet?: boolean
+  ) => {
     // Reset all fields after this one.
     const position = fieldOrder.indexOf(name);
     for (let i = position + 1; i < fieldOrder.length; i++) {
       let value = "";
       if (fieldOrder[i] === "mode") {
-        value = editing ? NetworkLinkMode.AUTO : NetworkLinkMode.LINK_UP;
+        value =
+          editing && hasSubnet ? NetworkLinkMode.AUTO : NetworkLinkMode.LINK_UP;
```
huwshimi, line 90: `LINK_UP` is ["Unconfigured"](https://github.com/canonical-web-and-design/maas-ui/blob/master/ui/src/app/store/utils/node/networking.ts#L472)

ui/src/app/machines/views/MachineDetails/MachineNetwork/NetworkFields/NetworkFields.test.tsx
Caleb-Ellis: What do you mean by "a leak between the tests"? If you mean that state in one test affects state in another, it could be the use of `sequence` for model ids. It increments by one every time it's invoked so maybe there can be overlaps when working with low id numbers? Maybe we should just be using `random`?

Caleb-Ellis (approved):

huwshimi:
> What do you mean by "a leak between the tests"? If you mean that state in one test affects state in another, it could be the use of sequence for model ids. It increments by one every time it's invoked so maybe there can be overlaps when working with low id numbers? Maybe we should just be using random?

I'm not sure. It was like something was happening asynchronously and not getting cleaned up between the tests, but setting up state isn't async...

I tried all sorts of combinations of `waitFor` and `act` and even tried manually calling the RTL `cleanup` between tests.

I don't think it could have been conflicting model ids because the ids were set to fixed values in the tests.

PR #3809, opened 2022-04-13
[periodic hardware sync] verify back-end integration

petermakowski:
## Done

- verify hardware sync back-end integration
- refactor timeSpan with date-fns
- add HardwareSyncFields type for better type safety, more info below.

#### HardwareSyncFields type
```tsx
type HardwareSyncFields =
  | {
      enable_hw_sync: false;
    }
  | {
      enable_hw_sync: true;
      last_sync: string;
      next_sync: string;
      is_sync_healthy: boolean;
      sync_interval: Seconds;
    };
```

Now the code below will throw an error (and rightly so, as `is_sync_healthy` won't be added to the machine details object if the hardware sync is not enabled.
```javascript
export const getHasSyncFailed = (machine?: Machine | null): boolean => {
  if (!isMachineDetails(machine)) {
    return false;
  }
  return machine.is_sync_healthy === false;
};
```

This can be fixed by checking whether `machine.enable_hw_sync` is true and the rest of keys will exist on the machine details object.
```javascript
export const getHasSyncFailed = (machine?: Machine | null): boolean => {
  if (!isMachineDetails(machine) || !machine.enable_hw_sync) {
    return false;
  }
  return machine.is_sync_healthy === false;
};
```

Note: this isn't as strict in tests due to how `cooky-cutter` type merging works; e.g. code below does not report type errors.
```javascript
machineDetailsFactory({
  enable_hw_sync: false,
  is_sync_healthy: true,
}
```

### MAAS deployment

- [Bolla](/HACKING.md#bolla)

### Running the branch

You can run this branch by:

### QA steps
- go to a machine that can be deployed (e.g. `MAAS/r/machine/c7mn8b/summary`) and to deploy form via "take action" -> "deploy"
- verify that the checkbox looks like before - and displays the interval in `minutes`

## Fixes
Fixes: https://github.com/canonical-web-and-design/app-tribe/issues/788

## Launchpad issue

## Backports

### DIFF
--- a/ui/src/app/store/machine/types/base.ts
+++ b/ui/src/app/store/machine/types/base.ts
@@ -67,7 +67,8 @@ export type BaseMachine = BaseNode & {
 // used in the machine details pages. This type contains all possible properties
 // of a machine model.
 export type MachineDetails = BaseMachine &
-  TimestampFields & {
+  TimestampFields &
+  HardwareSyncFields & {
     bios_boot_method: string;
     bmc: number;
     boot_disk: Disk | null;
@@ -82,7 +83,6 @@ export type MachineDetails = BaseMachine &
     devices: NodeDeviceRef[];
     dhcp_on: boolean;
     disks: Disk[];
-    enable_hw_sync: boolean;
     error: string;
     events: NodeEvent[];
     grouped_storages: GroupedStorage[];
@@ -92,14 +92,11 @@ export type MachineDetails = BaseMachine &
     installation_status: number;
     interface_test_status: TestStatus;
     interfaces: NetworkInterface[];
-    is_sync_healthy?: boolean;
-    last_sync: string;
     license_key: string;
     memory_test_status: TestStatus;
     metadata: NodeMetadata;
     min_hwe_kernel: string;
     network_test_status: TestStatus;
-    next_sync: string;
     node_type: number;
     numa_nodes: NodeNumaNode[];
     on_network: boolean;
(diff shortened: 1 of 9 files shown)


ui/src/app/store/machine/types/base.ts
```diff
@@ -111,11 +108,22 @@ export type MachineDetails = BaseMachine &
...
+  | {
+      enable_hw_sync: false;
+    }
+  | {
+      enable_hw_sync: true;
+      last_sync: string;
+      next_sync: string;
+      is_sync_healthy: boolean;
+      sync_interval: Seconds;
+    };
```
Caleb-Ellis, line 125: Interesting. This isn't how the websocket API *normally* works - we usually get back everything, with `null` returned for the values that are not applicable.

On one hand it makes sense to only get back the data that's relevant, and with this granularity we can be more confident in the possible value combinations for a machine. On the other hand this opens up the `Machine` type to take many different shapes, where each new conditional like this multiplies the possible `Machine` shapes by two. I think that could quickly become problematic, or at least be a hassle having to coerce the type in multiple ways in multiple places. I'm leaning toward asking if we could update the API to return `null`, if only for simplicity and consistency's sake.

What do you think @petermakowski @huwshimi?

Caleb-Ellis (approved):
LGTM with one comment for consideration

ui/src/app/store/machine/types/base.ts
Caleb-Ellis: Actually I've changed my mind, having to null check each of these values even when you *know* they're not null sounds like more of a hassle. Forget everything I just said hahaha

Issue #2748, opened 2021-06-09
Remove FormikForm component and replace with Formik


Caleb-Ellis:
https://github.com/canonical-web-and-design/maas-ui/pull/2741 introduced some changes so that formik's base functionality and our custom functionality was better separated. This has resulted in `FormikForm` being an unnecessary abstraction layer.

Caleb-Ellis:
> After further discussion between @Caleb-Ellis and myself this doesn't seem like the right approach due to the high chance of bugs it introduces. I'll close this and we can revisit if we come up with a better solution.

PR #3812, opened 2022-04-14
refactor(ui): remove FormikForm component

huwshimi:
## Done

- Remove the FormikForm content and use Formik instead.

### MAAS deployment

- [Bolla](/HACKING.md#bolla)

### Running the branch

You can run this branch by:

### QA steps

- All forms should be the same, it's probably worth trying a few to make sure.

## Fixes

Fixes: #2748.


ui/src/app/base/components/ActionForm/ActionForm.tsx
```diff
@@ -96,31 +98,40 @@ const getLabel = (
...
-  FormikFormProps<V, E>,
+  FormikFormContentProps<V, E>,
   "buttonsAlign" | "saved" | "saving" | "savingLabel" | "submitLabel"
 > & {
   actionName: string;
   loaded?: boolean;
+  initialTouched?: FormikConfig<V>["initialTouched"];
+  initialValues: FormikConfig<V>["initialValues"];
   modelName: string;
+  onSubmit: FormikConfig<V>["onSubmit"];
```
huwshimi, line 109: `ActionForm` intercepts `onSubmit` so we can't wrap these forms in `Formik` and because we're already spreading on `FormikFormContent` we need to explicitly capture the `Formk` props.

ui/src/app/base/components/Login/Login.tsx
```diff
@@ -89,34 +90,37 @@ export const Login = (): JSX.Element => {
                   Go to login page
                 </Button>
               ) : (
-                <FormikForm<LoginValues>
-                  errors={error}
+                <Formik
```
huwshimi, line 93: This PR is so huge mostly because of indentation changes. What's going on is that each form gets wrapped in `<Formik>` and the `Formik` props (`initialValues`, `onSubmit` etc.) get moved to `<Formik>`.

ui/src/app/base/components/DhcpForm/DhcpForm.tsx
```diff
@@ -89,9 +93,7 @@ export const DhcpForm = ({
   }
 
   return (
-    <FormikForm<DHCPFormValues>
-      cleanup={dhcpsnippetActions.cleanup}
-      errors={errors}
+    <Formik
```
Caleb-Ellis, line 96: Do you think not including the form values generic here could cause an issue? `Formik` still infers the values from `initialValues` but I'm wondering if there are any instances where that's not enough. Can you think of a way where we could include the generic once, and have it be used for both `Formik` and `FormikFormContent`?

Caleb-Ellis (approved):
LGTM, tested out a bunch of forms throughout the app and everything works fine. Just one comment about using generics

ui/src/app/base/components/DhcpForm/DhcpForm.tsx
huwshimi: I did end up having to set the value type for both components in a few places as it wasn't able to infer the full type (e.g. [here](https://github.com/canonical-web-and-design/maas-ui/pull/3812/files#diff-447826ada27bb574ed7ccc0695504cd55577b1e7ac1a7c8a500d3c0480a9dc6fR98)).

I don't think there's any way of providing the type automatically, I guess we could have a pattern where if you want to share the values type then you use an alias e.g. `type FormValues = AddDeviceValues`.

I suppose the danger with this new way is that errors could creep in if you accidentally used two different generics for the two components.

I guess the way to solve that would be two wrap both components in a new component that passes the generic along, maybe we could call it something like `FormikForm` :).

ui/src/app/base/components/DhcpForm/DhcpForm.tsx
```diff
@@ -40,7 +42,9 @@ type Props = {
   analyticsCategory: string;
   id?: DHCPSnippet["id"];
   onSave?: () => void;
-} & Partial<FormikFormProps<DHCPFormValues>>;
+} & Partial<
+  FormikFormContentProps<DHCPFormValues, FormikErrors<DHCPFormValues>>
```
petermakowski, line 46: I wonder if this repetition is necessary. Have you tried setting `FormikErrors<V>` as default for the second type parameter in `FormikFormContentProps`?

petermakowski, line 46: I wonder if this repetition is necessary. Have you tried setting `FormikErrors<V>` as default for the second type parameter in `FormikFormContentProps`?

petermakowski (approved):
Nice. Getting rid of the custom `FormikForm` definitely makes things less confusing.

huwshimi:
After further discussion between @Caleb-Ellis and myself this doesn't seem like the right approach due to the high chance of bugs it introduces. I'll close this and we can revisit if we come up with a better solution.

PR #3836, opened 2022-04-21
fix(ui): update TLS sections based on latest changes

Caleb-Ellis:
## Done

- Removed download button from TLS enabled page.
- Hide TLS section in user preferences if TLS is not enabled.
- Change lower limit for TLS expiry notification from 0 days to 1 day.

### MAAS deployment

- [Bolla](/HACKING.md#bolla)

### Running the branch

You can run this branch by:

### QA steps

- For a MAAS with TLS disabled:
  - Check that there is no "TLS fingerprint" section in User preferences > API authentication
- For a MAAS with TLS enabled:
  - Check that there is no certificate download button in Settings > Configuration > Security
  - Check that the minimum notification interval is 1 day instead of 0 days

## Fixes

Fixes canonical-web-and-design/app-tribe#839

### DIFF
--- a/ui/src/app/settings/views/Configuration/Security/TLSEnabled/TLSEnabled.tsx
+++ b/ui/src/app/settings/views/Configuration/Security/TLSEnabled/TLSEnabled.tsx
@@ -1,10 +1,9 @@
-import { Icon } from "@canonical/react-components";
+import { Icon, Textarea } from "@canonical/react-components";
 import { useDispatch, useSelector } from "react-redux";
 import * as Yup from "yup";
 
 import TLSEnabledFields from "./TLSEnabledFields";
 
-import CertificateDownload from "app/base/components/CertificateDownload";
 import CertificateMetadata from "app/base/components/CertificateMetadata";
 import FormikForm from "app/base/components/FormikForm";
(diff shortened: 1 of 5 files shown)


ui/src/app/settings/views/Configuration/Security/TLSEnabled/TLSEnabled.tsx
```diff
@@ -20,7 +19,8 @@ export type TLSEnabledValues = {
 export enum Labels {
   NotificationCheckbox = "Notify the certificate is due to expire in...",
   Interval = "Days",
-  IntervalRangeError = "Notification interval must be between 0 and 90 days.",
+  IntervalRangeError = "Notification interval must be between 1 and 90 days.",
```
Caleb-Ellis, line 22: Apparently `Computed values are not permitted in an enum with string valued members.` D:

huwshimi: I think for strings that have computed values it would be OK to not have the string defined in an enum (even though in this case it would use values that are also enums). I think that computing a string using the min/max values here might be less error prone.

huwshimi (approved):

PR #3890, opened 2022-05-04
fix(ui): fix instances of tooltip icons not using buttons

Caleb-Ellis:
## Done

- Used `TooltipButton` component wherever we had an icon show a tooltip on hover
- Slightly updated `TooltipButton` prop shape so that `iconName` is exposed at the top-level

### MAAS deployment

- [Bolla](/HACKING.md#bolla)

### Running the branch

You can run this branch by:

### QA steps

- Navigate around the app and check that every instance of an icon tooltip is attached to a button

## Fixes

Fixes canonical-web-and-design/app-tribe#869
Fixes canonical-web-and-design/app-tribe#894

### DIFF
--- a/ui/src/app/base/components/TooltipButton/TooltipButton.tsx
+++ b/ui/src/app/base/components/TooltipButton/TooltipButton.tsx
@@ -13,12 +13,14 @@ import { breakLines, unindentString } from "app/utils";
 type Props = Omit<TooltipProps, "children"> & {
   buttonProps?: SubComponentProps<ButtonProps>;
   children?: ReactNode;
-  iconProps?: SubComponentProps<IconProps>;
+  iconName?: IconProps["name"];
+  iconProps?: SubComponentProps<Omit<IconProps, "name">>;
 };
 
 const TooltipButton = ({
   buttonProps,
   children,
+  iconName = "information",
   iconProps,
   message,
   ...tooltipProps
@@ -34,13 +36,13 @@ const TooltipButton = ({
     >
       <Button
         appearance="base"
-        className="u-display--block u-no-border u-no-margin"
+        className="u-no-border u-no-margin u-text--default-size"
         hasIcon
         small
         type="button"
         {...buttonProps}
       >
-        {children || <Icon name="information" {...iconProps} />}
+        {children || <Icon name={iconName} {...iconProps} />}
       </Button>
     </Tooltip>
   );
--- a/ui/src/app/base/components/TooltipButton/TooltipButton.tsx
+++ b/ui/src/app/base/components/TooltipButton/TooltipButton.tsx
@@ -10,14 +10,16 @@ import { Button, Icon, Tooltip } from "@canonical/react-components";
 
 import { breakLines, unindentString } from "app/utils";
 
-type Props = Omit<TooltipProps, "children"> & {
-  buttonProps?: SubComponentProps<ButtonProps>;
+type Props = Omit<TooltipProps, "aria-label" | "children"> & {
+  "aria-label"?: ButtonProps["aria-label"];
+  buttonProps?: SubComponentProps<Omit<ButtonProps, "aria-label">>;
   children?: ReactNode;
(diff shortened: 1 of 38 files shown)


petermakowski:
There's now additional padding that's pushing the content further down and makes the icon no longer properly aligned vertically (apparent when displayed in table headings). See an example below. Ideally this shouldn't cause any visual regressions.

Kapture 2022-05-05 at 15 08 26

ui/src/app/machines/views/MachineDetails/MachineSummary/OverviewCard/StatusCard/StatusCard.tsx
```diff
@@ -101,7 +102,12 @@ const StatusCard = ({ machine }: Props): JSX.Element => {
             <hr />
             <p className="u-text--muted">
               Periodic hardware sync enabled{" "}
-              <Tooltip
+              <TooltipButton
+                buttonProps={{
+                  "aria-label": "more about periodic hardware sync",
+                }}
```
petermakowski, line 108: Adding an aria-label will be a _very_ common use case (as we're mostly relying on just icons for these buttons so we need labels in most cases).

I can't imagine why you'd ever want/need to attach aria-label an element other than `button`. I'd suggest to simplify and make it a top-level prop.

```suggestion
                aria-label="more about periodic hardware sync"
```

petermakowski (approved):
LGTM, just a small comment for your consideration.

ui/src/app/machines/views/MachineDetails/MachineSummary/OverviewCard/StatusCard/StatusCard.tsx
Caleb-Ellis: Sure thing, sounds good to me :+1:

Caleb-Ellis:
> There's now additional padding around tooltip icons. It's causing a variety of issues, e.g. pushing the content further down or messing up the vertical alignment (apparent when displayed in table headings).

Ack, can't believe I missed that. I've pushed up a commit that *almost* fixes the issue. Now the content is only pushed down by 1px in table headers but fine everywhere else. This is due to the font-size in table headers being 15px instead of 16px, and all the icons are standardised at 16px. I think this is probably fine, and definitely a better alternative to having slightly fuzzy icons?

ui/src/app/base/components/TooltipButton/TooltipButton.tsx
```diff
@@ -34,13 +38,14 @@ const TooltipButton = ({
...
-        className="u-display--block u-no-border u-no-margin"
+        aria-label={ariaLabel}
+        className="u-no-border u-no-line-height u-no-margin"
         hasIcon
         small
         type="button"
         {...buttonProps}
       >
-        {children || <Icon name="information" {...iconProps} />}
+        {children || <Icon name={iconName} {...iconProps} />}
```
petermakowski, line 48: (This is out of scope of this PR but we should keep this in mind for later)

Automatically sync images


As the click/touch target is extremely small (often 16x16px) we might want to activate the tooltip on hover of not just the icon but the accompanying label as well (so in this case “Automatically sync images”). This raises some questions around the API of this component.

Should we always render the icon if the iconName prop is provided instead?

```suggestion
        {children}
        {iconName ? <Icon name={iconName} {...iconProps} /> : null}
```

If we wanted to keep this behaviour this should be made obvious (e.g. by conditional type that would return an error if you try to combine children and iconName).

petermakowski (approved):

ui/src/app/base/components/TooltipButton/TooltipButton.tsx
petermakowski: https://github.com/canonical-web-and-design/maas-ui/issues/3907

Caleb-Ellis: +1 on being able to render both icon and children, and yeah thanks for filing an issue we can look into that later!

PR #3893, opened 2022-05-04
fix: display consistent dhcp snippet type

petermakowski:
## Done

- display consistent dhcp snippet type

### Before

image

### After
image

### MAAS deployment

- [Bolla](/HACKING.md#bolla)

### Running the branch

You can run this branch by:

### QA steps

- Add a DHCP snippet for each type via settings - `MAAS/r/settings/dhcp` (machine, device, subnet and global) and apply to one of the items on the list
-  verify that a correct type is displayed on the details page (e.g. subnet details) page and in the settings

## Fixes

Fixes: https://github.com/canonical-web-and-design/app-tribe/issues/899

## Launchpad issue

## Backports

### DIFF
--- a/ui/src/app/base/components/DhcpSnippetType/DhcpType.test.tsx
+++ b/ui/src/app/base/components/DhcpSnippetType/DhcpType.test.tsx
@@ -1,0 +1,104 @@
+import { render, screen } from "@testing-library/react";
+import { Provider } from "react-redux";
+import configureStore from "redux-mock-store";
+
+import DhcpType from "./DhcpSnippetType";
+
+import type { RootState } from "app/store/root/types";
+import {
+  controllerState as controllerStateFactory,
+  deviceState as deviceStateFactory,
+  dhcpSnippet as dhcpSnippetFactory,
+  dhcpSnippetState as dhcpSnippetStateFactory,
+  machine as machineFactory,
+  machineState as machineStateFactory,
+  modelRef as modelRefFactory,
+  subnet as subnetFactory,
+  subnetState as subnetStateFactory,
+  rootState as rootStateFactory,
+} from "testing/factories";
+
+const mockStore = configureStore();
+
+let state: RootState;
+
+beforeEach(() => {
+  state = rootStateFactory({
+    controller: controllerStateFactory({
+      loaded: true,
+    }),
+    device: deviceStateFactory({
+      loaded: true,
+    }),
+    dhcpsnippet: dhcpSnippetStateFactory({
+      loaded: true,
+      items: [
+        dhcpSnippetFactory({ id: 1, name: "class", description: "" }),
+        dhcpSnippetFactory({
+          id: 2,
+          name: "lease",
+          subnet: 2,
+          description: "",
+        }),
+        dhcpSnippetFactory({
+          id: 3,
+          name: "boot",
+          node: "xyz",
+          description: "",
+        }),
+      ],
+    }),
+    machine: machineStateFactory({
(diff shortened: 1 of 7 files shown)


ui/src/app/base/components/DhcpSnippetType/DhcpType.test.tsx
```diff
@@ -0,0 +1,104 @@
+import { render, screen } from "@testing-library/react";
+import { Provider } from "react-redux";
+import configureStore from "redux-mock-store";
+
+import DhcpType from "./DhcpSnippetType";
```
Caleb-Ellis, line 5: Is the different filename here an accident?

ui/src/app/settings/hooks.ts
```diff
@@ -47,6 +55,13 @@ export const useDhcpTarget = (
     (!!subnetId && subnetLoaded) ||
     (!!nodeId && controllerLoaded && deviceLoaded && machineLoaded);
 
+  useEffect(() => {
+    dispatch(subnetActions.fetch());
+    dispatch(controllerActions.fetch());
+    dispatch(deviceActions.fetch());
+    dispatch(machineActions.fetch());
+  }, [dispatch]);
```
Caleb-Ellis, line 63: I'm of two minds about this general pattern of always dispatching fetch actions. Most of the time its fine when its handled by a larger parent component, but in a hook like this that's used in table rows these actions are going to be fetched a lot. Sure the websocket saga prevents the fetch actually happening, but it's a lot of unnecessary noise that's especially unhelpful when going over the action logs.

All this to say this is fine to land because it's a pattern used pretty much everywhere, but I'd like it if we came up with an alternative. Maybe a simple hook for each model would be enough...
``` typescript
export const useFetchSubnets = (): void => {
  const subnetsLoaded = useSelector(subnetSelectors.loaded);

  useEffect(() => {
    if (!subnetsLoaded) {
      dispatch(subnetActions.fetch());
    }
  }, [dispatch, subnetsLoaded]);
};
```

Caleb-Ellis (approved):

ui/src/app/base/components/DhcpSnippetType/DhcpType.test.tsx
petermakowski: Good spot - missed this when renaming. Updated.

ui/src/app/settings/hooks.ts
petermakowski: Totally agree with you. Checking if that has been loaded before fetching did cross my mind, but decided no to do it for the reason you mentioned.

> Maybe a simple hook for each model 

I like your thinking. I was thinking about something similar a while ago... What if we provided something like useData/useQuery type-of hook for each model? We could potentially make quite a few improvements as this new layer of abstraction wouldn't be so coupled to the redux store (e.g. including a single status string as an option).

```javascript
const { status, loading, loaded (...), data, error } = useQuery('subnets')
``````

We could even make something that would allow for a combination of models:
```javascript
const [ subnets, devices ] = useQueries(["subnets", "devices"])

subnets.status === "success" ? subnets.data.map(...
```

...or add a single status / allLoaded or similar and not have to duplicate this logic in all components that need data from multiple models like we do now.
```javascript
const { status, slices: [subnets, devices] ] = useQueries(["subnets", "devices"])
```

petermakowski: https://github.com/canonical-web-and-design/maas-ui/issues/3896

PR #3957, opened 2022-05-24
Spike: nested routes with React Router v6

huwshimi:
@petermakowski @Caleb-Ellis let me know what you think about this as a workaround to be able to use React Router v6 with  absolute paths.

The issue we face is that in RR v6 if you have nested `Routes` (which is used in place of `<Switch>`) then any `Link` used inside of those routes become relative to the current `Route`. e.g. the subnets page is shown at `/networks` and a link on that page to `/vlan/5001` would actually end up at `/networks/vlan/5001`. To get to the right location you'd have to link to `../vlan/5001` which isn't great for reusable components and is very fragile.

There's a request to change this here: https://github.com/remix-run/react-router/issues/8035.

We currently use a `Switch` for each section of the app so this branch shows how we could include all the routes at the top level, therefore removing the issue of nested `Routes` and so making it possible to use absolute URLs.

If this seems reasonable then we can go ahead and update the app with this approach.


ui/package.json
```diff
@@ -42,6 +42,7 @@
     "react-dropzone": "12.1.0",
     "react-redux": "7.2.8",
     "react-router-dom": "5.3.1",
+    "react-router-dom-v5-compat": "6.3.0",
```
huwshimi, line 45: This package allows us to migrate parts of the app to React Router v6 without doing it all at once.

ui/src/app/base/hooks/urls.ts
```diff
@@ -1,28 +1,20 @@
...
 
 import { parseNumberId } from "app/utils";
 
-type RouteParams = Record<string, string>;
-
 /**
  * Get the id param from the URL and convert to the right type.
  * @param pk - The model id key.
  */
-export function useGetURLId<P extends RouteParams, K extends keyof P>(
```
huwshimi, line 13: These changes are just to make this hook to work with React Router v6 types.

huwshimi:
This would be slightly more awkward than I was hoping as we'd have to pass state from the wrapping component e.g.

```JSX
<MachineDetails children={
    (setHeaderContent) => <MachineSummary setHeaderContent={setHeaderContent} />
} />
```

ui/src/app/App.tsx
```diff
@@ -51,6 +52,8 @@ export const App = (): JSX.Element => {
   const completedUserIntro = useCompletedUserIntro();
   const debug = process.env.NODE_ENV === "development";
 
+  const appRoutes = useRoutes(routes);
```
petermakowski, line 55: Is there a reason why you used `useRoutes` over regular JSX? The approach you're proposing would also work with maintaining the current JSX style of declaring routes.

petermakowski, line 55: Is there a reason why you used `useRoutes` over regular JSX? The approach you're proposing would also work with maintaining the current JSX style of declaring routes.

petermakowski:
> This would be slightly more awkward than I was hoping as we'd have to pass state from the wrapping component e.g.
> 
> ```js
> <MachineDetails children={
>     (setHeaderContent) => <MachineSummary setHeaderContent={setHeaderContent} />
> } />
> ```

I really like the current approach we use for routes, not sure this workaround is worth the effort - especially in case the issue you linked in the description gets resolved soon.

huwshimi:
> I experimented with this a little, link below in case you're interested: [main...petermakowski:spike-routes-outlet](https://github.com/canonical-web-and-design/maas-ui/compare/main...petermakowski:spike-routes-outlet)

I think that works because the top level Routes file is still using a React Router v5 `<Switch>`. Once you upgrade that you end up with the same situation where the child `Routes` need to be relative (right now they're all absolute).

Using an outlet does look like a clean way to isolate the Details components from the inner components.

I've made another attempt which uses relative paths in the child `Routes` which means there's no need to make changes to the Details components (so no need to wrap each route, or use the outlet).

PR #3959, opened 2022-05-25
refactor(ui): use React Router v6 Link components

huwshimi:
## Done

- Update React Router Links to the v6 components
- Update tests to support the v6 compatibility router.

### MAAS deployment

- [Bolla](/HACKING.md#bolla)

### Running the branch

You can run this branch by:

### QA steps

- Load the React client.
- Make sure you can navigate through the app via the header links and links within the content.

## Fixes

Fixes: #3958.

### DIFF
--- a/ui/src/app/App.test.tsx
+++ b/ui/src/app/App.test.tsx
@@ -2,6 +2,7 @@ import * as reactComponentHooks from "@canonical/react-components/dist/hooks";
 import { mount } from "enzyme";
 import { Provider } from "react-redux";
 import { MemoryRouter } from "react-router-dom";
+import { CompatRouter } from "react-router-dom-v5-compat";
 import configureStore from "redux-mock-store";
 
 import { App } from "./App";
@@ -38,7 +39,9 @@ describe("App", () => {
     const wrapper = mount(
       <Provider store={store}>
         <MemoryRouter initialEntries={[{ pathname: "/settings" }]}>
-          <App />
+          <CompatRouter>
+            <App />
+          </CompatRouter>
         </MemoryRouter>
       </Provider>
     );
@@ -52,7 +55,9 @@ describe("App", () => {
     const wrapper = mount(
       <Provider store={store}>
         <MemoryRouter initialEntries={[{ pathname: "/settings" }]}>
-          <App />
+          <CompatRouter>
+            <App />
+          </CompatRouter>
         </MemoryRouter>
       </Provider>
     );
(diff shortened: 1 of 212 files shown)


ui/src/app/App.test.tsx
```diff
@@ -38,7 +39,9 @@ describe("App", () => {
     const wrapper = mount(
       <Provider store={store}>
         <MemoryRouter initialEntries={[{ pathname: "/settings" }]}>
-          <App />
+          <CompatRouter>
```
huwshimi, line 42: This same change is made in all the test files containing links.

ui/src/app/base/components/DHCPTable/DHCPTable.tsx
```diff
@@ -2,7 +2,7 @@ import { useEffect, useState } from "react";
 
 import { List, MainTable, Spinner } from "@canonical/react-components";
 import { useDispatch, useSelector } from "react-redux";
-import { Link } from "react-router-dom";
+import { Link } from "react-router-dom-v5-compat";
```
huwshimi, line 5: These are the changes to use the v6 Links.

anthonydillon:
Out of curiosity, what is the benefit of v6 over the older version of Link?

huwshimi:
> Out of curiosity, what is the benefit of v6 over the older version of Link?

There's nothing specific that we need for `Link` it's just a step towards upgrading React Router itself to v6 to save a really gigantic PR (as if this PR wasn't gigantic enough!)

huwshimi:
> > > Out of curiosity, what is the benefit of v6 over the older version of Link?
> > 
> > 
> > There's nothing specific that we need for `Link` it's just a step towards upgrading React Router itself to v6 to save a really gigantic PR (as if this PR wasn't gigantic enough!)
> 
> Yeah exactly, such a large diff for a dep change. Something seems overly complex in this system. Just an observation.

Maybe to give the full picture, the reason that this is necessary is that React Router introduced a bunch of breaking changes in v6.

To ease the transition they introduced the compatibility router so that you can use v5 and v6 alongside one another.

This PR introduces that compatibility layer as well as upgrading the Links, however the reason this PR is so large is that the compatibility layer needed to be introduced in all tests that involve routing (which is 99% of the changes in the PR).

Once we've updated everything to React Router v6 we can remove the compatibility router.

huwshimi:
> Oh sorry, I understand that and it's not an opinion of complexity in this project but in React router.

Yeah it's definitely a little unfortunate that the upgrade is so complex. I think they realised that which is why they made the compatibility layer, but it's still a lot of work!

Caleb-Ellis (approved):
LGTM! It's a shame the upgrade is such a hassle but at least this gets us started on an incremental upgrade

PR #3991, opened 2022-06-03
feat(ui): add logs to controllers

huwshimi:
## Done

- Display the logs and events for controllers.
- Refactor logs components to work with controllers.

### MAAS deployment

- [Bolla](/HACKING.md#bolla)

### Running the branch

You can run this branch by:

### QA steps

- Go to the React logs tab for a controller.
- Check that you can see events and installation output and can download the files.

## Fixes

Fixes: canonical-web-and-design/app-tribe#944.
Fixes: canonical-web-and-design/app-tribe#945.

## Screenshots

Screen Shot 2022-06-03 at 10 09 24 am


Caleb-Ellis (approved):
LGTM! I did notice a bug where the Download button gets disabled when you change tabs, but I think that wasn't introduced here and seems to be caused by `get` and `setActive` being called on every route change

huwshimi:
> LGTM! I did notice a bug where the Download button gets disabled when you change tabs, but I think that wasn't introduced here and seems to be caused by `get` and `setActive` being called on every route change

Thanks for filing the issue, we should fix that!

Issue #3994, opened 2022-06-03
Controller details pages calling "get" and "setActive" on every route change


Caleb-Ellis:
Every time you navigate to a different tab in controller details both `get` and `setActive` are called. They should only be called once.

PR #4020, opened 2022-06-06
fix(ui): controller details calling setActive on every route change

petermakowski:
## Done

- fix controller details calling setActive on every route change
  - add integration test to `Controllers.test.tsx` verifying it gets and sets the controller as active only once when navigating within the same controller

### MAAS deployment

- [Bolla](/HACKING.md#bolla)

### Running the branch

You can run this branch by:

### QA steps

- Steps for QA.

## Fixes

Fixes: https://github.com/canonical-web-and-design/maas-ui/issues/3994

## Launchpad issue

## Backports

## Screenshots

### DIFF
--- a/ui/src/app/controllers/views/Controllers.tsx
+++ b/ui/src/app/controllers/views/Controllers.tsx
@@ -28,12 +28,7 @@ const Controllers = (): JSX.Element => {
         controllersURLs.controller.usbDevices(null, true),
         controllersURLs.controller.vlans(null, true),
       ].map((path) => (
-        <Route
-          exact
-          key={path}
-          path={path}
-          render={() => <ControllerDetails />}
-        />
+        <Route exact path={path} render={() => <ControllerDetails />} />
       ))}
       <Route path="*" render={() => <NotFound />} />
     </Switch>
--- a/ui/src/app/controllers/views/Controllers.tsx
+++ b/ui/src/app/controllers/views/Controllers.tsx
@@ -28,7 +28,14 @@ const Controllers = (): JSX.Element => {
         controllersURLs.controller.usbDevices(null, true),
         controllersURLs.controller.vlans(null, true),
       ].map((path) => (
-        <Route exact path={path} render={() => <ControllerDetails />} />
+        <Route
+          // using a single key as a workaround for Controller details pages
+          // calling "get" and "setActive" on every route change
+          key="controller-details"
+          exact
+          path={path}
+          render={() => <ControllerDetails />}
+        />
       ))}
       <Route path="*" render={() => <NotFound />} />
     </Switch>
--- a/ui/src/app/controllers/constants.ts
+++ b/ui/src/app/controllers/constants.ts
@@ -1,3 +1,5 @@
+import type urls from "./urls";
+
 import { NodeActions } from "app/store/types/node";
 
 export const ControllerActionHeaderViews = {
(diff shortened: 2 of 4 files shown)


ui/src/app/controllers/views/Controllers.tsx
```diff
@@ -28,12 +28,7 @@ const Controllers = (): JSX.Element => {
         controllersURLs.controller.usbDevices(null, true),
         controllersURLs.controller.vlans(null, true),
       ].map((path) => (
-        <Route
-          exact
-          key={path}
-          path={path}
-          render={() => <ControllerDetails />}
-        />
+        <Route exact path={path} render={() => <ControllerDetails />} />
```
petermakowski, line 31: Removed the key prop as react-router handles this internally and we were overwriting the key it's using causing route to re-render.

ui/src/app/Routes.tsx
```diff
@@ -57,45 +57,34 @@ const Routes = (): JSX.Element => (
         </ErrorBoundary>
       )}
     />
-    {[
-      controllersURLs.controllers.index,
-      controllersURLs.controller.index(null, true),
-    ].map((path) => (
```
Caleb-Ellis, line 63: @huwshimi wasn't there a reason we switched to mapping over an array? Could it be a "preparing for React Router v6" thing?

huwshimi: Thats right, in v6 `path` no longer supports an array.

ui/src/app/controllers/views/Controllers.tsx
huwshimi: I can't remember for sure, but I wonder if the key was there so that if you change tabs it doesn't maintain the same scroll position. It might be worth checking.

ui/src/app/Routes.tsx
petermakowski: Thanks, will revert to using mapping through the array.

ui/src/app/controllers/views/Controllers.tsx
petermakowski: I can't imagine why this would break it, the order of the array never changes so even if a default array index was used as a key, the render still should work as expected. I checked it anyway and haven't noticed anything strange.

Caleb-Ellis: but honestly why did RRv6 have to go and break everything :(

huwshimi: Hmm... it does also appear to break the scrolling for me when it doesn't have the key:

Screen Recording 2022-06-07 at 6 01 10 pm

petermakowski: ok thanks, at least I now know what you meant exactly with the scrolling. I'll double check.

petermakowski: @huwshimi  It's not caused by this change, looks like we forgot to add the `useScrollToTop` hook to `Controllers`

Caleb-Ellis: I think what's happening is when using dynamic `key`s we're overriding `Route`'s intended behaviour, causing the component to unmount and remount on every route change. This unmount/remount triggers `get` and `setActive` calls, but it also sends the scroll position back to the top. So I don't think the current implementation breaks scrolling so much as fixes an unintentional scroll to top. We should probably be more explicit about when we want route changes to automatically scroll to the top, i.e. using `useScrollOnRender()` or something like that

petermakowski: > We should probably be more explicit about when we want route changes to automatically scroll to the top, i.e. using `useScrollOnRender()` or something like that

Can you think of a use case where the current implementation via `useScrollToTop` (scroll to top on pathname change) would not be sufficient?

If you think of each new pathname like a new page this pretty much replicates regular browser's behaviour (with static HTML pages).

Caleb-Ellis: Nope! `useScrollToTop` works fine

Caleb-Ellis (approved):
LGTM, shame about the workaround!

PR #4029, opened 2022-06-08
fix(cypress): fix Cypress utils imports

huwshimi:
## Done

- Fix the util imports in the Cypress tests.

## QA

n/a

## Fixes

Fixes: #4028.
Fixes: #4030.


cypress/integration/with-users/legacy/auth.spec.ts
```diff
@@ -1,14 +0,0 @@
-import { generateMAASURL } from "../../utils";
-
-context("Legacy authentication", () => {
```
huwshimi, line 3: The legacy client no longer needs to redirect to the React client to log in. Login tests are covered in login.spec.js.

Caleb-Ellis (approved):

Caleb-Ellis:
I keep forgetting that the cypress tests don't run using the development build, so the green tick from the previous PR didn't really mean anything 😞

huwshimi:
> I keep forgetting that the cypress tests don't run using the development build, so the green tick from the previous PR didn't really mean anything 😞

Yeah and I keep forgetting to push my branches to upstream so they run the tests!

Issue #4027, opened 2022-06-07
Cannot save controller/device configuration if it has an automatic tag


Caleb-Ellis:
The API won't allow the `update` method to be called if it includes an automatic tag. We need to filter those tags out of the tag select component as well as from the dispatch action.

PR #4041, opened 2022-06-08
fix: cannot save controller or device configuration with automatic tags

petermakowski:
## Done

- fix: cannot save controller or device configuration with automatic tags

### MAAS deployment

- [Bolla](/HACKING.md#bolla)

### Running the branch

You can run this branch by:

### QA steps
- Go to tags and add a new automatic tag (with a description)
- Go to device or controller configuration and verify that this tag does not show on the list of possible choices

## Fixes

Fixes: https://github.com/canonical-web-and-design/maas-ui/issues/4027

## Launchpad issue

## Backports

## Screenshots

### DIFF
--- a/ui/src/app/base/components/NodeConfigurationFields/NodeConfigurationFields.test.tsx
+++ b/ui/src/app/base/components/NodeConfigurationFields/NodeConfigurationFields.test.tsx
@@ -76,6 +76,39 @@ it("can open a create tag form", async () => {
   );
 });
 
+it("does not display automatic tags on the list", async () => {
+  const manualTag = tagFactory({ id: 1, name: "tag1" });
+  const automaticTag = tagFactory({
+    id: 4,
+    name: "automatic-tag",
+    definition: `//node[@class="system"]/vendor = "QEMU"`,
+  });
+  state.tag.items = [manualTag, automaticTag];
+  const store = mockStore(state);
+  render(
+    <Provider store={store}>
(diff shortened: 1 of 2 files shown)


Caleb-Ellis:
Is this still meant to be a draft PR?

petermakowski:
> Is this still meant to be a draft PR?

Ready for review now!

Caleb-Ellis (changes requested):
The auto tags have been filtered out of the select, but they're still being sent in the API request so I get the error back `Cannot add tag to node because it has a definition`. At a minimum the auto tags need to be filtered out of the API request, but maybe we should also disable auto tags from the select so they can't be removed? e.g.:

petermakowski:
> The auto tags have been filtered out of the select, but they're still being sent in the API request so I get the error back `Cannot add tag to node because it has a definition`. At a minimum the auto tags need to be filtered out of the API request

If you don't send the automatic tags through the API request, they are removed from the controller - this seems to be an issue with the API and I filed a bug: https://bugs.launchpad.net/maas/+bug/1978118

> maybe we should also disable auto tags from the select so they can't be removed? e.g.:
> 
> 

Good point, I added the disabled attribute to automatic tag buttons.

Caleb-Ellis (approved):
Ah ok - in that case this looks good on our end and hopefully sometime soon the API will be fixed.

PR #3922, opened 2022-05-09
refactor: sort jsx props alphabetically

petermakowski:
## Done

-  sort jsx props alphabetically via `react/jsx-sort-props`

### MAAS deployment

To run this branch you will need access to one of the following MAAS deployments:

- [Karura](/HACKING.md#karura)
- [Bolla](/HACKING.md#bolla)
- [A development MAAS](/HACKING.md#development-deployment)
- [A local snap MAAS](/HACKING.md#snap-deployment) (this will not usually have machines)

### Running the branch

You can run this branch by:

- Serving with [dotrun](/HACKING.md#maas-ui-development-setup)
- [Building in a development MAAS](/HACKING.md#running-maas-ui-from-a-development-maas)

### QA steps

- Steps for QA.

## Fixes

Fixes: https://github.com/canonical-web-and-design/maas-ui/issues/3314

## Launchpad issue

Related Launchpad maas issue in the form `lp#number`.

## Backports

In general, please propose fixes against *master* rather than release branches (e.g. 2.7), unless the fix is only applicable for that specific release. Please apply backport labels to the PR (e.g. "Backport 2.7") for the appropriate releases to target.

Only bug and security fixes should be backported, new features should only land in master.

PR #4089, opened 2022-06-14
Fix eslint on save

huwshimi:
## Done

- Fix eslint on save.

### QA steps

- Open vscode. Make a change that the linter complains about.
- Save the file.
- The eslint error should get fixed (if it knows how to).

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -1,3 +1,6 @@
 {
-  "cSpell.words": ["lifecycles", "unmount"]
+  "cSpell.words": ["lifecycles", "unmount"],
+  "editor.codeActionsOnSave": {
+    "source.fixAll.eslint": true
+  }
 }


.vscode/settings.json
```diff
@@ -1,3 +1,6 @@
 {
-  "cSpell.words": ["lifecycles", "unmount"]
+  "cSpell.words": ["lifecycles", "unmount"],
+  "editor.codeActionsOnSave": {
+    "source.fixAll.eslint": true
```
huwshimi, line 4: Should we enable this do you think?

Caleb-Ellis: I'm personally not a fan. Formatting on save always catches me by surprise - sometimes I just want to save a file so I can close it, then I can come back later to fix it up and have it be exactly as I left it, warts and all. Plus I've baked it into my muscle memory to format before saving "for real" anyway.

Though I'm not gonna be around to be very slightly annoyed by it so don't listen to me at all hahaha

huwshimi: I guess I was thinking about https://github.com/canonical-web-and-design/maas-ui/pull/3922 and the import order linting which is much easier for the computer to do. Right now I've got a shortcut mapped to fix eslint errors and I can happily keep running that.

petermakowski: I always had this enabled in my user settings anyway. Let's get this merged!

petermakowski (approved):

.vscode/settings.json
huwshimi: > I always had this enabled in my user settings anyway. Let's get this merged!

I looked, but I couldn't find a way to do that! What do I need to do? That seems a bit better than having to turn this on for everyone.

petermakowski: Adding this option to global vscode user preferences (`Preferences: Open Settings (JSON)`  via the command palette) seemed to do the trick for me.

huwshimi: Thanks! I didn't know about that.

PR #4091, opened 2022-06-14
refactor: update zone state shape in Redux

### LINKED ISSUE #3055: Explore ways to improve errors in Redux state
In some cases we have two separate forms on a page that both dispatch the same action when submitted, e.g. KVM config has two forms that call `podActions.update()`. If an error is returned it will show up in both forms. We should find a way to link an error with the form that submitted it, or alternatively, track in form state whether it has been submitted or not and show an error based on that.

Caleb-Ellis:
## Done
- This branch updates the state shape for the zone model in order to better handle actions and errors. It will serve as the blueprint for updating the rest of the models to the same shape.
- Supersedes https://github.com/canonical-web-and-design/maas-ui/pull/3503

### MAAS deployment

- [Bolla](/HACKING.md#bolla)

### Running the branch

You can run this branch by:

### QA steps

- Open Redux DevTools
- For the following actions, check that they work and that you like how the changes in state occur:
  - Go to the zone list and check that zones are fetched
  - Check that creating a zone works
  - Try to create another zone with the same name and check that an error shows
  - Go to a zone details page
  - Check that updating the zone works
  - Try updating the zone name to something that already exists and check that an error shows
  - Check that you can delete a zone and get redirected to the list

## Fixes

Fixes #3055

### DIFF
--- a/src/app/store/zone/selectors.ts
+++ b/src/app/store/zone/selectors.ts
@@ -1,13 +1,110 @@
-import { generateBaseSelectors } from "app/store/utils";
-import { ZoneMeta } from "app/store/zone/types";
-import type { Zone, ZoneState } from "app/store/zone/types";
+import { createSelector } from "@reduxjs/toolkit";
 
-const searchFunction = (zone: Zone, term: string) => zone.name.includes(term);
+import { ZONE_ACTIONS } from "./constants";
+import type {
+  ZoneActionNames,
+  ZoneGenericActions,
+  ZoneModelActions,
+  ZonePK,
+  ZoneState,
+} from "./types";
+import { ZoneMeta } from "./types";
 
-const selectors = generateBaseSelectors<ZoneState, Zone, ZoneMeta.PK>(
-  ZoneMeta.MODEL,
-  ZoneMeta.PK,
-  searchFunction
+import { ACTION_STATUS } from "app/base/constants";
+import type { RootState } from "app/store/root/types";
+import { isId } from "app/utils";
+
+const all = (state: RootState): ZoneState["items"] =>
+  state[ZoneMeta.MODEL].items;
+
+const errors = (state: RootState): ZoneState["errors"] =>
+  state[ZoneMeta.MODEL].errors;
+
+const genericActions = (state: RootState): ZoneState["genericActions"] =>
+  state[ZoneMeta.MODEL].genericActions;
+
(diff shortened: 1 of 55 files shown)


src/app/store/zone/slice.ts
```diff
@@ -1,22 +1,249 @@
...
-    UpdateParams
-  >(ZoneMeta.MODEL, ZoneMeta.PK),
+  initialState: {
+    errors: [],
+    genericActions: initialGenericActions,
+    items: [],
+    modelActions: initialModelActions,
+  } as ZoneState,
+  reducers: {
+    [cleanup]: (state, action: PayloadAction<ZoneActionNames[]>) => {
```
Caleb-Ellis, line 89: @huwshimi I've updated the cleanup action as you suggested in the previous PR, so now you pass in a list of actions and it clears errors and statuses for those actions only. I thought about having the default be to clear everything, but maybe that promotes encourages behaviour?

src/app/store/zone/selectors.ts
```diff
@@ -1,13 +1,110 @@
...
+import type {
+  ZoneActionNames,
+  ZoneGenericActions,
+  ZoneModelActions,
+  ZonePK,
+  ZoneState,
+} from "./types";
+import { ZoneMeta } from "./types";
 
-const selectors = generateBaseSelectors<ZoneState, Zone, ZoneMeta.PK>(
```
huwshimi, line 7: When we land this PR we should create a followup issue to build the new generate util.

src/app/store/zone/slice.ts
```diff
@@ -1,22 +1,249 @@
...
+    statuses[loading].push(zonePK);
+  } else {
+    statuses[success].push(zonePK);
+  }
+};
 
 const zoneSlice = createSlice({
   name: ZoneMeta.MODEL,
-  initialState: genericInitialState as ZoneState,
-  reducers: generateCommonReducers<
```
huwshimi, line 14: We should file an issue about following up to recreate this util as well.

src/app/store/zone/slice.ts
huwshimi: Yeah think this less likely to lead to unintentional bugs.

huwshimi (approved):
This is looking great! I think we probably better get this landed before it's too late.

PR #4098, opened 2022-06-15
Spike 2: nested routes

huwshimi:
This is a much lower impact change to support nested routes than my previous attempt: https://github.com/canonical-web-and-design/maas-ui/pull/3957.

The main change here is to use a function to transform absolute URLs into paths that are relative to the child `Routes`.


src/app/Routes.tsx
```diff
@@ -64,101 +65,103 @@ const Routes = (): JSX.Element => (
...
       />
     ))}
-    {[devicesURLs.devices.index, devicesURLs.device.index(null, true)].map(
-      (path) => (
-        <Route
-          key={path}
-          path={path}
-          render={() => (
-            <ErrorBoundary>
-              <Devices />
```
huwshimi, line 81: Instead of wrapping the list and details components in the same component we'd need to move all top level paths (e.g. `/devices/`, `/device/`) into this file. These top level components (e.g. `Devices`) don't currently contain any state, they just wrap the routes.

src/app/Routes.tsx
```diff
@@ -64,101 +65,103 @@ const Routes = (): JSX.Element => (
...
+        <ErrorBoundary>
+          <DeviceList />
+        </ErrorBoundary>
+      }
+    />
+    <Route
+      path={`${devicesURLs.device.index(null, true)}/*`}
+      element={
+        <ErrorBoundary>
+          <DeviceDetails />
```
huwshimi, line 87: These components contain often shared state between the child routes and with this approach (unlike this one: https://github.com/canonical-web-and-design/maas-ui/pull/3957) we can keep this structure.

src/app/devices/views/DeviceDetails/DeviceDetails.tsx
```diff
@@ -66,27 +68,24 @@ const DeviceDetails = (): JSX.Element => {
       }
     >
       {device && (
-        <Switch>
+        <Routes>
           <Route
-            exact
-            path={deviceURLs.device.summary(null, true)}
-            render={() => <DeviceSummary systemId={id} />}
+            path={getRelativeRoute(deviceURLs.device, "summary")}
```
huwshimi, line 73: The relative path gets extracted for each child route.

src/app/devices/views/Devices.tsx
```diff
@@ -1,30 +0,0 @@
...
-
-import DeviceDetails from "./DeviceDetails";
-import DeviceList from "./DeviceList";
-
-import NotFound from "app/base/views/NotFound";
-import devicesURLs from "app/devices/urls";
-
-const Devices = (): JSX.Element => {
-  return (
-    <Switch>
```
huwshimi, line 11: These intermediary components are no longer required as the routes would be moved to the top level `Routes` file.

src/app/devices/views/Devices.tsx
```diff
@@ -1,30 +0,0 @@
...
-      <Route
-        exact
-        path={devicesURLs.devices.index}
-        render={() => <DeviceList />}
-      />
-      {[
-        devicesURLs.device.configuration(null, true),
-        devicesURLs.device.index(null, true),
-        devicesURLs.device.network(null, true),
-        devicesURLs.device.summary(null, true),
```
huwshimi, line 21: We'd no longer need to define all the subsections like this as they're captured by the top level `/devices/*` path.

src/app/utils/getRelativeRoute.ts
```diff
@@ -0,0 +1,40 @@
...
+};
+
+/**
+ * Returns a relative route using the index of the routes object.
+ * @param routes - A collection of urls containing an index and the route you
+ * want to return.
+ * @param route - A route key on the routes object.
+ * @returns A relative route.
+ */
+export const getRelativeRoute = <A extends ArgPathArgs>(
```
huwshimi, line 33: I actually wrote this relative path extractor for an early attempt at migrating to v6.

src/app/Routes.tsx
```diff
@@ -64,101 +65,103 @@ const Routes = (): JSX.Element => (
...
+    <Route
+      path={devicesURLs.devices.index}
+      element={
+        <ErrorBoundary>
+          <DeviceList />
+        </ErrorBoundary>
+      }
+    />
+    <Route
+      path={`${devicesURLs.device.index(null, true)}/*`}
```
huwshimi, line 84: Only the index needs to be defined at the top level, all the child routes are handled inside the details component as before.

huwshimi:
@petermakowski @Caleb-Ellis @ndv99 what do you think of this as a way to upgrade to React Router v6?

Caleb-Ellis:
@huwshimi this seems like a much cleaner approach. It's still in the "spirit" of React Router v6 but adapted to suit our particular structure. Plus I'm a big fan of not having to define a route in multiple places!

petermakowski (commented):
Thanks @huwshimi - this looks really good.

src/app/utils/getRelativeRoute.ts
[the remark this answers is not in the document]
huwshimi: This looks like it's going to be necessary to work with KVM urls which are nested several levels deep (or at least a lot easier to type).

huwshimi:
This has a PR now: https://github.com/canonical-web-and-design/maas-ui/pull/4143

PR #4100, opened 2022-06-15
refactor: replace instances of fireEvent with userEvent and enforce with linting

### LINKED ISSUE #3934: Enforce using userEvent library instead of fireEvent for simulating user interactions
`userEvent` adds related event calls from browsers to make tests more realistic than its counterpart `fireEvent`, which is a low-level api.

This makes for a more user-centric, and closer-to-truth testing of user interactions.

Ideally this should be fixed before we begin React migration so that new tests follow this rule.

- [ ] add eslint rule https://github.com/testing-library/eslint-plugin-testing-library/blob/main/docs/rules/prefer-user-event.md
- [ ] replace existing `fireEvent` with `userEvent`

Caleb-Ellis:
## Done

- Replaced most instances of `fireEvent` with `userEvent`
- Added a rule to our eslint config to only allow `fireEvent` for change methods (currently it's needed for changing sliders)

### MAAS deployment

- [Bolla](/HACKING.md#bolla)

### Running the branch

You can run this branch by:

### QA steps

- N/A

## Fixes

Fixes #3934

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -127,6 +127,14 @@ module.exports = {
       rules: {
         "testing-library/prefer-find-by": "off",
         "testing-library/prefer-explicit-assert": "error",
+        "testing-library/prefer-user-event": [
+          "error",
+          {
+            // Remove once sliders can be updated with user-event
+            // https://github.com/testing-library/user-event/issues/871
+            allowedMethods: ["change"],
+          },
+        ],
       },
     },
     {
--- a/src/app/base/components/NodeConfigurationFields/NodeConfigurationFields.test.tsx
+++ b/src/app/base/components/NodeConfigurationFields/NodeConfigurationFields.test.tsx
@@ -1,4 +1,4 @@
-import { fireEvent, render, screen, waitFor } from "@testing-library/react";
+import { render, screen, waitFor } from "@testing-library/react";
 import userEvent from "@testing-library/user-event";
 import { Formik } from "formik";
 import { Provider } from "react-redux";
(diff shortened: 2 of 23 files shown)


Caleb-Ellis:
@huwshimi @petermakowski I cannot for the life of me figure out how to get this test to work with `userEvent`:
https://github.com/Caleb-Ellis/maas-ui/blob/use-user-event/src/app/machines/views/MachineDetails/MachineConfiguration/PowerForm/PowerForm.test.tsx#L158

The change event just never seems to be triggered with `userEvent.selectOptions`. The power form stuff is pretty spaghetti-y in order to get validation working for all the different power types, so it could be some iffy logic in the code. Anyway let me know if you can see an obvious fix here.

huwshimi:
> @huwshimi @petermakowski I cannot for the life of me figure out how to get this test to work with `userEvent`: https://github.com/Caleb-Ellis/maas-ui/blob/use-user-event/src/app/machines/views/MachineDetails/MachineConfiguration/PowerForm/PowerForm.test.tsx#L158
> 
> The change event just never seems to be triggered with `userEvent.selectOptions`. The power form stuff is pretty spaghetti-y in order to get validation working for all the different power types, so it could be some iffy logic in the code. Anyway let me know if you can see an obvious fix here.

It looks like the select is disabled because the machine is inside a pod. If you remove the pod attribute from the machine it should enable the field which should allow userEvent to work.

Caleb-Ellis:
@huwshimi you're a genius, thank you! `userEvent` is already proving itself more reliable than accessing the `onChange` handler directly :P

Should probably remove `pod` from the test factory too, seems like that's more of an exception and shouldn't be the default

huwshimi (approved):

PR #4102, opened 2022-06-16
fix: use correct API request for commissioning without tests

### LINKED ISSUE #3955: smartctl-validate test runs even when explicitly removed from commissioning step
Bug originally filed by setuid at https://bugs.launchpad.net/bugs/1964024

As titled, when commissioning machines, if you explicitly remove (uncheck) the smartctl-validate unit test, it runs anyway, regardless of whether it's enabled or not. 

This presents a larger problem of triggering race conditions with concurrent commissioning, where multiple smartctl-validate tests report their success/failure back to regiond, and cause race conditions, leading to commissioning failures. 

Reducing the number nodes requested to commission in parallel reduces the impact of these failures, but it's still non-zero. 

When tests are disabled, they should not run at all, including smartctl-validate.

Caleb-Ellis:
## Done

- Fixed commissioning a machine without running tests
- Updated types so that only script names are used when running commissioning/testing

### MAAS deployment

- [Bolla](/HACKING.md#bolla)

### Running the branch

You can run this branch by:

### QA steps

- Go to the Tests tab of a machine that can be tested
- Open the Take action menu and click Commission
- Clear the testing script field so none are selected and submit the form
- Check that the tests tab does not update with a new test, and that the commissioning action was dispatched with `payload.params.extra.testing_scripts === ["none"]` instead of `[]`

## Fixes

Fixes #3955


src/app/machines/components/MachineHeaderForms/ActionFormWrapper/CommissionForm/CommissionForm.tsx
```diff
@@ -132,15 +129,18 @@ export const CommissionForm = ({
...
         if (updateFirmware) {
           commissioningScriptsParam.push("update_firmware");
         }
         if (configureHBA) {
           commissioningScriptsParam.push("configure_hba");
         }
-        const testingScriptsParam = testingScripts.map((script) => script.id);
+        const testingScriptsParam = testingScripts.length
+          ? testingScripts.map((script) => script.name)
+          : ["none"];
```
huwshimi, line 143: It feels like this special string should be an enum.

src/app/store/machine/types/actions.ts
```diff
@@ -43,13 +43,13 @@ export type CloneParams = BaseNodeActionParams & {
...
 export type CommissionParams = BaseNodeActionParams & {
-  commissioning_scripts?: (Script[ScriptMeta.PK] | Script["name"])[];
+  commissioning_scripts?: Script["name"][];
   enable_ssh?: boolean;
   script_input?: ScriptInputParam;
   skip_bmc_config?: boolean;
   skip_networking?: boolean;
   skip_storage?: boolean;
-  testing_scripts?: (Script[ScriptMeta.PK] | Script["name"])[];
+  testing_scripts?: Script["name"][] | ["none"];
```
huwshimi, line 52: The enum could be used here too.

huwshimi (approved):
Just one suggestion.

PR #4153, opened 2022-06-23
Server side filtering API interrogation

Caleb-Ellis:
I've set up a sandbox at http://0.0.0.0:8400/MAAS/r/ssf-test to test that the API is capable of doing what we need it to.

Acceptance criteria from the [spec](https://docs.google.com/document/d/1B8CQXRxzapUq8dbSILzCZdrNzU49brVS8K5KQAZeQiw/edit#):
- [ ] The user can filter machines by using free text search, which matches any of:
  - [ ] FQDN
  - [ ] MAC address
  - [ ] IP addresses
  - [ ] Power type
  - [ ] Status text
  - [ ] OS
  - [ ] Owner name
  - [ ] Tags
  - [ ] Pool name
  - [ ] Description
  - [ ] Zone name
  - [ ] Boot VLAN name
  - [ ] Boot VLAN’s fabric name
  - [ ] Architecture
- [ ] The user can filter machines by partially matching machine values (e.g. hostname contains “test”)
- [ ] The user can filter machines by exact matching machine values (e.g. hostname is “test-machine”)
- [ ] The user can filter machines by negatively matching machine values (e.g. power state is not “error”, status does not contain “Failed”)
- [ ] The user can filter machines by multiple machine values (e.g. hostname contains “test” and status is “Testing”)
- [ ] The user can filter machines by multiple matches of a machine value (e.g. status is “New” or status is “Ready”)
- [ ] The user can filter machines by a list of filter options.
- [ ] The list of machines is not added to or removed from after a filter has been performed, except in some specific circumstances:
  - [ ] When a machine is deleted it gets removed from the list.
- [ ] The data of the machines in the list is updated in real time.
- [ ] The user can see a count of the machines that match a certain filter criteria.
- [ ] The user can sort the list of machines, ascending or descending, according to a machine parameter.
- [ ] The list of machines can be paginated.
- [ ] The user can see a count of the total number of machines in their MAAS within 10 seconds of page load.
- [ ] The user can choose to group the machines in the machine list by owner, pool, power state, status or zone.
- [ ] The user can choose not to have the machines in the machine list grouped.
- [ ] The user can see a count of how many machines are in each group.
- [ ] The user can collapse and expand groups.


src/app/store/machine/slice.ts
```diff
@@ -1682,6 +1696,300 @@ const machineSlice = createSlice({
...
+          meta: { identifier },
+          payload,
+        } = action;
+        const list = state.lists[identifier];
+
+        if (!list) {
+          return;
+        }
+
+        // Convert ungrouped response into null group
```
Caleb-Ellis, line 1888: The API payload has a different shape depending on whether `group_key` was included in the request, so it's normalised in the reducer.

PR #4162, opened 2022-06-27
Migrate zones components to RTL part 2

ndv99:
## Done

Migrated the following to RTL:
- app/zones/views/ZoneDetails/ZoneDetails.test.tsx
- app/zones/views/ZoneDetails/ZoneDetailsForm/ZoneDetailsForm.test.tsx
- app/zones/views/ZoneDetails/ZoneDetailsHeader/DeleteConfirm/DeleteConfirm.test.tsx
- app/zones/views/ZoneDetails/ZoneDetailsHeader/ZoneDetailsHeader.test.tsx
- app/zones/views/ZonesList/ZonesList.test.tsx


src/app/zones/views/ZoneDetails/ZoneDetails.test.tsx
```diff
@@ -58,14 +57,9 @@ describe("ZoneDetails", () => {
...
-    expect(
-      wrapper
-        .findWhere(
-          (node) =>
-            node.type() === "button" && node.text() === Labels.EditButton
-        )
-        .exists()
-    ).toBe(true);
+
+    const editButtons = screen.getAllByRole("button", { name: "Edit" });
```
huwshimi, line 61: It's good practice to use the string from `Labels.EditButton` to make it easier to maintain these tests (i.e. we can update the label without breaking tests) as well as being more reliable when doing a test for `.not.toBeInTheDocument();`

src/app/zones/views/ZoneDetails/ZoneDetails.test.tsx
```diff
@@ -91,13 +85,8 @@ describe("ZoneDetails", () => {
...
-      wrapper
-        .findWhere(
-          (node) =>
-            node.type() === "button" && node.text() === Labels.EditButton
-        )
-        .exists()
-    ).toBe(false);
+
+    const editButtons = screen.queryAllByRole("button", { name: "Edit" });
+    editButtons.forEach((button) => expect(button).not.toBeInTheDocument());
```
huwshimi, line 90: Here the test would pass if we changed the button label in the component but didn't update it here.

src/app/zones/views/ZoneDetails/ZoneDetails.tsx
```diff
@@ -40,7 +40,10 @@ const ZoneDetails = (): JSX.Element => {
   }
 
   return (
-    <Section header={<ZoneDetailsHeader id={zoneID} />}>
+    <Section
+      aria-label="Zone Details"
```
huwshimi, line 44: In my example I've updated these to use enums for these labels.

src/app/zones/views/Zones.test.tsx
```diff
@@ -1,34 +1,67 @@
...
+} from "testing/factories";
 
 const mockStore = configureStore();
 
 describe("Zones", () => {
+  let initialState: RootState;
+
+  beforeEach(() => {
+    initialState = rootStateFactory({
+      user: userStateFactory({
```
huwshimi, line 27: The user is not necessary for these tests.

src/app/zones/views/Zones.test.tsx
```diff
@@ -1,34 +1,67 @@
...
-      path: "/not/a/path",
+      component: "Zone not found",
+      path: zonesURLs.details({ id: 2 }),
     },
   ].forEach(({ component, path }) => {
     it(`Displays: ${component} at: ${path}`, () => {
-      const store = mockStore(rootStateFactory());
-      const wrapper = mount(
+      const store = mockStore(initialState);
+      render(
```
huwshimi, line 64: The main reason this isn't working is that for these tests to work with React Testing Library the components need to be contained inside the matching route pattern e.g. for a zone it would be `/zone/:id` this is so the path params, in this case `:id` can be captured.

Inside this test you could add the `<Routes>` and `<Route>` components, but a better way would be to use the `renderWithBrowserRouter` function and pass the correct `routePattern` for each path. That could look something like this:

```TypeScript
[
  {
    label: ZoneListLabel.List,
    path: zonesURLs.index,
    pattern: zonesURLs.index,
  },
  {
    label: ZoneDetailsLabel.Details,
    path: zonesURLs.details({ id: 1 }),
    pattern: zonesURLs.details(null, true),
  },
  {
    label: NotFoundLabel.Title,
    path: `${zonesURLs.index}/not/a/path`,
    pattern: `${zonesURLs.index}/*`,
  },
].forEach(({ label, path, pattern }) => {
  it(`Displays: ${label} at: ${path}`, () => {
    renderWithBrowserRouter(<Zones />, {
      wrapperProps: { routePattern: pattern, state },
      route: path,
    });
    expect(screen.getByLabelText(label)).toBeInTheDocument();
  });
});
```

I've pushed a branch as a full example of how I got this test to pass:
https://github.com/canonical-web-and-design/maas-ui/commit/41066b2c4c501856c2ef3905037f0ce5f3d97325

You could also look at the Tags component as a reference: https://github.com/canonical-web-and-design/maas-ui/blob/main/src/app/tags/views/Tags.test.tsx#L41.

huwshimi (commented):
Hey @ndv99 the routing is new to all of us so it's a bit tricky if you haven't worked with it before. I've left a few comments, hope they make sense. Do let me know if not!

src/app/zones/views/ZoneDetails/ZoneDetailsForm/ZoneDetailsForm.test.tsx
```diff
@@ -43,13 +42,13 @@ describe("ZoneDetailsForm", () => {
       </Provider>
     );
 
-    wrapper.find("button[data-testid='cancel-action']").simulate("click");
+    await userEvent.click(screen.getByTestId("cancel-action"));
```
petermakowski, line 45: Could you use a text label instead of a testid?

src/app/zones/views/ZoneDetails/ZoneDetailsHeader/DeleteConfirm/DeleteConfirm.test.tsx
```diff
@@ -44,15 +45,18 @@ describe("DeleteConfirm", () => {
       </Provider>
     );
 
-    wrapper.find("button[data-testid='delete-az']").simulate("click");
+    // wrapper.find("button[data-testid='delete-az']").simulate("click");
+    // expect(onConfirm).toHaveBeenCalled();
+
+    await userEvent.click(screen.getByTestId("delete-az"));
```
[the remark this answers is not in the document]
ndv99: Converted this to a label

src/app/zones/views/ZoneDetails/ZoneDetailsHeader/DeleteConfirm/DeleteConfirm.test.tsx
```diff
@@ -44,15 +45,18 @@ describe("DeleteConfirm", () => {
       </Provider>
     );
 
-    wrapper.find("button[data-testid='delete-az']").simulate("click");
+    // wrapper.find("button[data-testid='delete-az']").simulate("click");
+    // expect(onConfirm).toHaveBeenCalled();
```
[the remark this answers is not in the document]
ndv99: Removed!

src/app/zones/views/ZoneDetails/ZoneDetailsForm/ZoneDetailsForm.test.tsx
ndv99: nvm i did it

petermakowski (approved):
LGTM 🎉

PR #4168, opened 2022-06-30
Migrate settings components to RTL

ndv99:
## Done

Migrated the following files to RTL:
- src/app/settings/components/Nav/Nav.test.tsx
- src/app/settings/components/Routes/Routes.test.tsx
- src/app/settings/components/SettingsTable/SettingsTable.test.tsx

This will close [this](https://github.com/canonical-web-and-design/app-tribe/issues/1050) issue.

### DIFF
--- a/src/app/settings/components/Routes/Routes.test.tsx
+++ b/src/app/settings/components/Routes/Routes.test.tsx
@@ -1,6 +1,6 @@
 import Routes from "./Routes";
 
-import settingsURLs from "app/settings/urls";
+import urls from "app/base/urls";
 import type { RootState } from "app/store/root/types";
 import {
   rootState as rootStateFactory,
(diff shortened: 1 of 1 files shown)


src/app/settings/components/Routes/Routes.test.tsx
```diff
@@ -1,154 +1,199 @@
...
+    path: settingsURLs.dhcp.add,
+  },
+  {
+    title: `Editing \`${dhcpSnippet.name}\``,
+    path: settingsURLs.dhcp.edit({ id: dhcpSnippet.id }),
+    pattern: `${settingsURLs.dhcp.edit(null)}/*`,
+  },
+  {
+    title: "Package repos",
+    path: settingsURLs.repositories.index,
```
petermakowski, line 157: Would you mind changing this to an absolute url import for consistency with other files?

```tsx
import urls from "app/base/urls"
(...)
path: urls.settings.repositories.index,
```

src/app/settings/components/SettingsTable/SettingsTable.test.tsx
```diff
@@ -54,35 +64,39 @@ describe("SettingsTable", () => {
...
       loading={true}
-    />
+    />,
+    {}
   );
 
-  expect(wrapper.find("Button").props().disabled).toBe(true);
+  const button = screen.getByRole("link", { name: "Add User" });
+  expect(button).toBeInTheDocument();
+  expect(button.classList.contains("is-disabled")).toBe(true);
```
petermakowski, line 87: It's better to test for the disabled attribute (which affects how users can(not) interact with it), `is-disabled`. `className` can be useful to test, but it's more of an implementation detail (as `.is-disabled` className does not guarantee the proper style will be even applied, `testing-library`/`jsdom` doesn't load `css`).

```suggestion
  expect(button).toBeDisabled();
```

petermakowski (approved):

src/app/settings/components/SettingsTable/SettingsTable.test.tsx
ndv99: I tried this, but the test appears to fail when using `.toBeDisabled()`, even though it has the `is-disabled` class:

```
Received element is not disabled:
      <a aria-disabled="true" class="p-button is-disabled" href="/foo" />
```

Issue #4169, opened 2022-06-30
machne details scriptresult not found


petermakowski:
**Describe the bug**
I'm seeing an issue with the maas ui, both locally and on bolla. when I go to a machine details on the commissioning tab, if I select "details" for a script it always says "scriptresult not found"

**maas-ui version**
3.3 (main)

**Additional context**
This was introduced by https://github.com/canonical-web-and-design/maas-ui/commit/ce326f422a81c6c343ff64c052dad6a30bb07373 which isn't in 3.2 so no need for backports.

PR #4170, opened 2022-06-30
fix(ui): machine details routes

petermakowski:
## Done

- fix machine details routes

### MAAS deployment

- [Bolla](/HACKING.md#bolla)

### Running the branch

You can run this branch by:

### QA steps

- Go to machine commissioning details (e.g. `/MAAS/r/machine/pbpncx/commissioning/59674/details` on bolla) and verify the page is displayed correctly (and "Script result could not be found." is not shown)

## Fixes

Fixes: #4169

## Launchpad issue

## Backports

## Screenshots

### DIFF
--- a/cypress/e2e/with-users/machines/details.spec.ts
+++ b/cypress/e2e/with-users/machines/details.spec.ts
@@ -1,10 +1,24 @@
 import { generateMAASURL } from "../../utils";
+import { generateName, generateMac } from "../../utils";
 
 context("Machine details", () => {
   beforeEach(() => {
     cy.login();
     cy.visit(generateMAASURL("/machines"));
   });
+
+  const completeAddMachineForm = () => {
+    const name = generateName("machine");
+    cy.findByRole("button", { name: /Add hardware/i }).click();
+    cy.findByLabelText("submenu").within(() => {
+      cy.findByRole("button", { name: /Machine/i }).click();
+    });
+    cy.findByLabelText("Machine name").type(name);
+    cy.findByLabelText("MAC address").type(generateMac());
+    cy.findByLabelText("Power type").select("Manual");
+    cy.get("button[type='submit']").click();
+    return { name };
+  };
 
   it("hides the subnet column on small screens", () => {
     cy.findByRole("grid").within(() => {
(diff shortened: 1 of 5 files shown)


cypress/e2e/with-users/machines/details.spec.ts
```diff
@@ -24,4 +24,37 @@ context("Machine details", () => {
...
+    cy.waitForPageToLoad();
+    cy.get("[data-testid='section-header-buttons']").within(() =>
+      cy
+        .findByRole("button", {
+          name: /Take action/i,
+        })
+        .click()
+    );
+
+    // abort commissioning
```
huwshimi, line 43: I think this uncovers a bit of an issue with these tests. I presume we need to abort commissioning here because this test expects `machines/add.spec.ts` to have run first and so we assume that the machine that has been added there will still be commissioning.

I think instead we should set up a machine in the `beforeEach` (or potentially `beforeAll` if we think we can safely use the same machine for all tests) of this file to remove the dependent (and leaky) tests as well as removing this race condition.

This will slow these tests down, but make them more reliable.

src/app/machines/views/MachineDetails/MachineDetails.test.tsx
```diff
@@ -45,85 +45,67 @@ describe("MachineDetails", () => {
...
+  ].forEach(({ component, path }) => {
+    it(`Displays: ${component} at: ${path}`, async () => {
       const store = mockStore(state);
       const wrapper = mount(
         <Provider store={store}>
           <MemoryRouter initialEntries={[{ pathname: path }]}>
             <CompatRouter>
               <Routes>
-                <Route element={<MachineDetails />} path={route || "*/:id/*"} />
+                <Route element={<MachineDetails />} path={"/machine/:id/*"} />
```
huwshimi, line 108: This might be better as `urls.machines.machine.index(null)` (as well as the other instances of `/machine/:id` in this file).

src/app/base/components/node/NodeTestsTable/NodeTestsTable.tsx
```diff
@@ -174,6 +174,7 @@ const NodeTestsTable = ({ node, scriptResults }: Props): JSX.Element => {
   return (
     <>
       <MainTable
+        aria-label="Test results"
```
huwshimi, line 177: It doesn't look like you ended up using this label (as far as I can see).

huwshimi (approved):
Thanks for fixing this! A couple of tiny comments, but the main thing is to improve the integration tests.

src/app/base/components/node/NodeTestsTable/NodeTestsTable.tsx
petermakowski: I will, once this is in the dev snap and can be used in tests (right now if I tried `cy.findByRole("grid", { name: "Test results" }` this would have failed.

cypress/e2e/with-users/machines/details.spec.ts
petermakowski: Agreed, testing interactions that involve state updates in the MAAS back-end are tricky. I'm not sure about `beforeEach`, I'm concerned that if we keep going at this pace, we'll end up with the same problem that we had in the past - our test suite taking ages to complete. I'll look into this.

PR #4213, opened 2022-07-06
Site identity - MAAS name emoji help

ndv99:
## Done

- Wrote help about how to use emojis in the MAAS name field with examples.

You can see these changes by navigating to the Settings page, then Configuration --> General


petermakowski (changes requested):
The text content doesn't match the design (look at where the line break is applied and `my-maas my-maas` repetition):

### Implementation
image

### Design
image

ndv99:
Ahh damn I thought I pushed the fix for that cos I spotted it too. One sec!!!!

PR #4237, opened 2022-07-12
fix cypress test failure

petermakowski:
## Done

- fix cypress failure

### MAAS deployment

- [Bolla](/HACKING.md#bolla)

### Running the branch

You can run this branch by:

### QA steps

- Steps for QA.

## Fixes

Fixes: #4174 

## Launchpad issue

## Backports

## Screenshots

### DIFF
--- a/cypress/e2e/with-users/controllers/details.spec.ts
+++ b/cypress/e2e/with-users/controllers/details.spec.ts
@@ -84,14 +84,12 @@ context("Controller details", () => {
   });
 
   it("displays controller commissioning details", () => {
+    cy.findAllByRole("gridcell", { name: "Name" })
+      .first()
+      .within(() => {
+        cy.findByRole("link").click();
+      });
     cy.waitForPageToLoad();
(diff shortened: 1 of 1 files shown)


cypress/e2e/with-users/controllers/details.spec.ts
```diff
@@ -84,14 +84,12 @@ context("Controller details", () => {
   });
 
   it("displays controller commissioning details", () => {
+    cy.findAllByRole("gridcell", { name: "Name" })
```
huwshimi, line 87: It feels like this should happen in the `beforeEach` rather than repeated in each test or is there a way we can get the controller id so we can `cy.visit()` the controller details page directly?

petermakowski: I think it would make sense to use `cy.visit()` directly if we were creating a new controller via API call e.g. in `beforeAll`, but we have no way of doing that (yet). I moved the navigation to controller details page to `beforeEach` for now.

huwshimi (approved):
Looks good!

PR #4246, opened 2022-07-14
fix: configure VLAN DHCP

petermakowski:
## Done

- fix VLAN DHCP configuration form

### MAAS deployment

- [Bolla](/HACKING.md#bolla)

### Running the branch

You can run this branch by:

### QA steps

- Go to https://maas-ui-4246.demos.haus/MAAS/r/vlan/5004 and Click "Configure DHCP"
- Select a subnet
- Form submit button should be enabled

## Fixes

Fixes: #4212

## Launchpad issue

lp#1980818
https://bugs.launchpad.net/bugs/1980818

## Backports

## Screenshots

### DIFF
--- a/src/app/subnets/views/VLANDetails/ConfigureDHCP/ConfigureDHCP.tsx
+++ b/src/app/subnets/views/VLANDetails/ConfigureDHCP/ConfigureDHCP.tsx
@@ -164,7 +164,7 @@ const ConfigureDHCP = ({ closeForm, id }: Props): JSX.Element | null => {
               enableDHCP: true,
               endIP: "",
               gatewayIP: "",
-              primaryRack: vlan.primary_rack || vlan.rack_sids[0] || "",
+              primaryRack: vlan.primary_rack || "",
               relayVLAN: vlan.relay_vlan || "",
               secondaryRack: vlan.secondary_rack || "",
               startIP: "",
--- a/src/app/subnets/views/VLANDetails/ConfigureDHCP/ConfigureDHCP.tsx
+++ b/src/app/subnets/views/VLANDetails/ConfigureDHCP/ConfigureDHCP.tsx
@@ -67,6 +67,9 @@ const ConfigureDHCP = ({ closeForm, id }: Props): JSX.Element | null => {
     vlanSelectors.eventErrorsForVLANs(state, id, "configureDHCP")
   )[0]?.error;
   const [configuredDHCP, resetConfiguredDHCP] = useCycled(!configuringDHCP);
+  const ipRanges = useSelector((state: RootState) =>
+    ipRangeSelectors.getByVLAN(state, id)
+  );
   const saved = configuredDHCP && !configureDHCPError;
   const cleanup = useCallback(() => vlanActions.cleanup(), []);
   const loading =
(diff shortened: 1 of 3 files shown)


huwshimi:
@petermakowski when I submit the form I get an error "dhcp can only be turned on when a dynamic IP range is defined." so does the form also need to be disabled until a subnet has been selected?

Screen Shot 2022-07-18 at 10 14 41 am

petermakowski:
> @petermakowski when I submit the form I get an error "dhcp can only be turned on when a dynamic IP range is defined." so does the form also need to be disabled until a subnet has been selected?
> 
> Screen Shot 2022-07-18 at 10 14 41 am

I'm not sure, I wanted to keep this change to the minimum as it's a bug-fix.The behaviour of a disabled button in forms is often a subject of heated debate, I wouldn't say it's a bug, but I can look into making this more consistent with other forms (and disable the button until the form is valid).

src/app/subnets/views/VLANDetails/ConfigureDHCP/ConfigureDHCP.tsx
```diff
@@ -101,24 +101,32 @@ const ConfigureDHCP = ({ closeForm, id }: Props): JSX.Element | null => {
...
-                path: "subnet",
-              });
+      subnet: Yup.string()
+        .when(["enableDHCP", "dhcpType"], {
+          is: (
+            dhcpEnabled: boolean,
+            dhcpType: DHCPType.CONTROLLERS | DHCPType.RELAY
+          ) => dhcpEnabled && DHCPType.CONTROLLERS === dhcpType,
+          then: Yup.string().required("Subnet is required"),
+        })
```
huwshimi, line 111: It looks like this could do with a test (or at least I couldn't spot one).

src/app/subnets/views/VLANDetails/ConfigureDHCP/ConfigureDHCP.tsx
```diff
@@ -156,7 +164,7 @@ const ConfigureDHCP = ({ closeForm, id }: Props): JSX.Element | null => {
               enableDHCP: true,
               endIP: "",
               gatewayIP: "",
-              primaryRack: vlan.primary_rack || "",
+              primaryRack: vlan.primary_rack || vlan.rack_sids[0] || "",
```
huwshimi, line 167: I think a comment here about why assigning `vlan.rack_sids[0]` might be necessary.

I'm missing a little context here, but is this just picking any controller to appear as selected in the "Primary rack" select box? It seems a little strange that one of the options can be selected in the field without it setting the value in Formik.

petermakowski: Fair point, it might have been a little confusing indeed.

We had a fair bit of complexity here that could lead to more bugs further down the line. I changed the approach and added an empty default option to the field instead which is now compatible with all potential states it can be in (we had no empty - `""` - option on the list previously which caused this bug). The browser selected the only available value which was out of sync with Formik.

This is now consistent with the behaviour of most other forms we have.

src/app/subnets/views/VLANDetails/ConfigureDHCP/ConfigureDHCP.tsx
```diff
@@ -101,24 +101,32 @@ const ConfigureDHCP = ({ closeForm, id }: Props): JSX.Element | null => {
...
-                message: "This subnet has no available IP addresses.",
-                path: "subnet",
-              });
+      subnet: Yup.string()
+        .when(["enableDHCP", "dhcpType"], {
+          is: (
+            dhcpEnabled: boolean,
+            dhcpType: DHCPType.CONTROLLERS | DHCPType.RELAY
+          ) => dhcpEnabled && DHCPType.CONTROLLERS === dhcpType,
+          then: Yup.string().required("Subnet is required"),
```
huwshimi, line 110: When reconfiguring DHCP it looks like it might need to know if a subnet has already been set. If you go to https://maas-ui-4246.demos.haus/MAAS/r/vlan/5004/ and click "Configure DHCP" and then change the radio to "Provide DHCP from rack controller(s)" and select a controller the form is still disabled (which I suspect is because of this change).

huwshimi (approved):
Looking good, one small thing left to fix.

PR #4259, opened 2022-07-19
feat(machine): store result of machine.get

huwshimi:
## Done

- Store the result of machine.get requests using the new details store.

### MAAS deployment

- [Bolla](/HACKING.md#bolla)

### Running the branch

You can run this branch by:

### QA steps

- Load a machine details page and open redux devtools.
- You should see dispatches for `machine/get`, `machine/getStart` and `machine/getSuccess`.
- In the redux state there should an entry for each get request in machine -> details and each one should have `loaded: true`.

## Fixes

Fixes: canonical-web-and-design/app-tribe#1167.

### DIFF
--- a/src/app/store/machine/slice.ts
+++ b/src/app/store/machine/slice.ts
@@ -26,11 +26,14 @@ import type {
   DeletePartitionParams,
   DeleteVolumeGroupParams,
   DeployParams,
+  FilterGroupResponse,
   GetSummaryXmlParams,
   GetSummaryYamlParams,
   LinkSubnetParams,
   Machine,
+  MachineDetails,
   MachineState,
+  MachineStateDetailsItem,
   MarkBrokenParams,
   MountSpecialParams,
   ReleaseParams,
@@ -44,7 +47,6 @@ import type {
   UpdateFilesystemParams,
   UpdateParams,
   UpdateVmfsDatastoreParams,
-  FilterGroupResponse,
 } from "./types";
 
 import type { ScriptResult } from "app/store/scriptresult/types";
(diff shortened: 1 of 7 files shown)


petermakowski:
I noticed that `machine.get` WebSocket message with the same payload is sent to the back-end at least twice when loading the machine details page (and the back-end comes back with a result each time as expected).

This issue doesn't seem to be introduced by this PR - same thing happens on bolla. We might want to address that. I created an issue so this can be followed up on. https://github.com/canonical-web-and-design/maas-ui/issues/4274

src/app/store/machine/slice.ts
```diff
@@ -1015,33 +1016,104 @@ const machineSlice = createSlice({
...
+        const machine = action.payload;
+        // If the item already exists, update it, otherwise
+        // add it to the store.
+        const i = state.items.findIndex(
+          (draftItem: Machine) => draftItem.system_id === machine.system_id
+        );
+        if (i !== -1) {
+          state.items[i] = machine;
+        } else {
+          state.items.push(machine);
```
petermakowski, line 1108: Could you perform this operation this without mutating the original array/object. https://redux.js.org/faq/immutable-data

src/app/store/machine/slice.ts
```diff
@@ -1015,33 +1016,104 @@ const machineSlice = createSlice({
...
+          string,
+          GenericItemMeta<{ system_id: Machine[MachineMeta.PK] }>
+        >
+      ) => {
+        const machine = action.payload;
+        // If the item already exists, update it, otherwise
+        // add it to the store.
+        const i = state.items.findIndex(
+          (draftItem: Machine) => draftItem.system_id === machine.system_id
+        );
```
petermakowski, line 1104: Correct me if I'm wrong, but my understanding is that we will get a complete data for the machine each time we request machine details. We might as well avoid the additional complexity of having to perform the update operation and simply use the machine data as is and discard whatever we had previously instead?

Also storing machine data in an object instead of an array might be a better choice. It would make such operations easier/faster as we wouldn't need to iterate over an array to but simply use an object key lookup.

src/app/store/machine/slice.ts
```diff
@@ -1015,33 +1016,104 @@ const machineSlice = createSlice({
...
-      state.loading = false;
+    getSuccess: {
+      prepare: (
+        item: { system_id: Machine[MachineMeta.PK] },
+        requestId: string,
+        machine: MachineDetails
+      ) => ({
+        meta: {
+          item,
+          requestId,
```
petermakowski, line 1087: I think it's confusing that this `requestId` isn't the same as the `requestId` that already uniquely identifies each WebSocket message exchange.

petermakowski: image
image

src/app/store/machine/slice.ts
huwshimi: Redux Toolkit uses immer behind the scenes so while this looks like a mutation it's actually using using Immer's `produce` when executing the change: https://redux-toolkit.js.org/usage/immer-reducers

src/app/store/machine/slice.ts
huwshimi: > Correct me if I'm wrong, but my understanding is that we will get a complete data for the machine each time we request machine details. We might as well avoid the additional complexity of having to perform the update operation and simply use the machine data as is and discard whatever we had previously instead?

While each request returns all the data needed, `items` contains the data for each `list` and `get` request that is currently being displayed.

> 
> Also storing machine data in an object instead of an array might be a better choice. It would make such operations easier/faster as we wouldn't need to iterate over an array to but simply use an object key lookup.

We could revisit the decision to store data in arrays, but we would want to do that wholesale across the app, not just for machines and something that is out of scope for this feature. I don't remember all the reasons why we made that decision but I remember we did a bit of profiling. It made some operations quicker, but others slower.

src/app/store/machine/slice.ts
huwshimi: Good point! I've stuck this in a new issue as it'll also affect another branch I have: https://github.com/canonical-web-and-design/app-tribe/issues/1226.

src/app/store/machine/slice.ts
petermakowski: Oh interesting, I had no idea redux toolkit already uses Immer behind the scenes.

src/app/store/machine/slice.ts
petermakowski: A common pattern is to store data in object with keys with an additional array of keys to make iterating over items faster/easier. This would solve the problem of such change making certain operations being slower.

```
  machines: {
    abc: {...},
     ...
  },
  machinesById: ["abc", ...]
```

petermakowski: I'm not sure I agree with the sentiment that making a change that's not in line with the current paradigm should be automatically "out of scope". I'm worried that we'll keep doing things "the old way" just for the sake of consistency.

The whole point of server-side machine listing is to make (performance) improvements. If we see an opportunity for change (and we feel that it'd be a change in the right direction), we should make a change.

huwshimi: > A common pattern is to store data in object with keys with an additional array of keys to make iterating over items faster/easier. This would solve the problem of such change making certain operations being slower.
> 
> ```
>   machines: {
>     abc: {...},
>      ...
>   },
>   machinesById: ["abc", ...]
> ```

That's actually the shape that we explored originally, but profiling helped us to realise this wasn't the best way (in maas-ui), particularly as it caused us to have to do more manipulation of the data we were receiving from the API.

huwshimi: > I'm not sure I agree with the sentiment that making a change that's not in line with the current paradigm should be automatically "out of scope". I'm worried that we'll keep doing things "the old way" just for the sake of consistency.

Consistency is an incredibly important part of maintaining a project as large as MAAS. We've been trying not make a change or start a refactor without the whole picture in mind or a plan to finish the refactor otherwise over time we will inevitably end up with a fragmented and unmaintainable codebase.

My statement was that making this change across the codebase is out of scope for this feature.

> 
> The whole point of server-side machine listing is to make (performance) improvements. If we see an opportunity for change (and we feel that it'd be a change in the right direction), we should make a change.

Sure, but right now you're proposing a change that hasn't been shown to be a performance bottleneck for the machine list, in fact the shape of the store will become drastically less of a performance issue as we're no longer needing to load and store all machines in Redux.

petermakowski: > Consistency is an incredibly important part of maintaining a project as large as MAAS. We've been trying not make a change or start a refactor without the whole picture in mind or a plan to finish the refactor otherwise over time we will inevitably end up with a fragmented and unmaintainable codebase.

I agree with you on the importance of consistency. And MAAS is indeed currently very consistent - at least in part thanks to you striving to keep it that way.

I suppose where we disagree is that fragmentation inevitably leads to unmaintainable codebase. I believe it's worth sacrificing _some_ consistency to allow for creating pockets of innovation within the app. That as a result will allow for more experimentation and bolder changes that might ultimately change MAAS for the better. A little bit of disruption might be healthy here in the long run, but I understand if you disagree.

> My statement was that making this change across the codebase is out of scope for this feature.

Approaching any change this way makes it much less likely to ever happen which is my main worry. Trying out changes in a small area, testing them and adopting across the rest of the app once the solution is proven to work well is another way to tackle this.

petermakowski (approved):

src/app/store/machine/slice.ts
huwshimi: Sure, I understand your frustration and having this discussion is worthwhile because you've not been privy to the previous discussions on this topic.

However, I don't think this PR is the right place for this conversation. It took eight days to get this PR approved and in the meantime it was blocking a significant amount of other work.

The shape of how the machines are stored was proposed in the spec some months ago and I think that would have been the appropriate place to discuss changes like this, not in the middle of implementing the feature after we've agreed on the spec.

I think if we want to change this shape in the future I think we need to:
1. Show that there is a performance bottleneck so we're not prematurely optimising.
2. Show that the proposed solution gives a large enough performance improvement.
3. Have a plan for how it could be rolled out for all data models.

MAAS-UI has some unique challenges because it is a very large (and relatively mature) project that has a very small team working on it. I don't disagree at all about innovating and experimenting. We've actually been really good about trying out things and refactoring. But one of the reasons we're able to do those refactors is because of the codebase consistency. The more fragmented, the harder it is to make sweeping changes. I did a masterclass on this topic: ["Maintaining big projects with small teams"](https://docs.google.com/spreadsheets/d/1fFumFWIM3oHwLr9pcBlaANcadAU0bNJxbkfGKwC_1pg/edit#gid=1223088901) which could be an interesting watch if you haven't seen it already.

PR #4278, opened 2022-07-25
feat(redux): add machine lists to the redux store

huwshimi:
## Done

- Add the store of machine list requests to the machine slice and redux store.

## QA
n/a

## Fixes

Fixes: canonical-web-and-design/app-tribe#1142.


src/testing/factories/state.ts
```diff
@@ -240,6 +245,26 @@ export const licenseKeysState = define<LicenseKeysState>({
   ...defaultState,
 });
 
+export const machineStateListGroup = define<MachineStateListGroup>({
+  count: 15,
+  items: () => [],
```
petermakowski, line 250: It might be worth setting some reasonable defaults for items that match the count value.

petermakowski (approved):

src/testing/factories/state.ts
huwshimi: The pattern we've adopted for the factories is to only provide the minimum of what's required by the types (i.e. not providing optional values and leaving arrays etc. empty). This way we've been catching issues where a test passes when a component is expecting a value to exist, but doesn't support an empty/null value.

This also helps us to define what's required for a test to pass when using the factory.

Issue #4274, opened 2022-07-22
redundant websocket messages sent over the network


petermakowski:
`machine` (e.g. `machine.get`) WebSocket messages with the same payload is sent to the back-end multiple times when loading the machine details/machine listing page (and the back-end comes back with a result each time as expected).

This results in sending completely redundant data over the network and definitely does not help performance. This could be happening to other request types, we should look into it.
 
#### Notes
This also applies to machines lists
![image.png](https://images.zenhubusercontent.com/6110f1cbad9628d5cdec1846/9b329162-a84c-4fc0-bb04-22c3e219b7ac)

`nocache`
https://github.com/canonical-web-and-design/maas-ui/blob/main/src/app/base/sagas/websockets.ts#L575-L576

huwshimi:
Maybe in the hook for getting a machine it should check for whether the machine details exist in the store, or otherwise it could check if there is another request in the store for the same id. We still need to store a details request for each place that is using it so that we know when it is safe to unsubscribe from the machine.

petermakowski:
Closing as this is now being tracked on JIRA.

PR #4295, opened 2022-07-28
feat(machines): update machine.list to use new API

huwshimi:
## Done

- Update the slice to handle calls to machine.list to use the new API.

### MAAS deployment

- [Bolla](/HACKING.md#bolla)

### Running the branch

You can run this branch by:

### QA steps

- Go to the machine list. The machines should load.
- Open the redux dev tools and inside the `machines` store you should see `lists` populated with the requests. Note: a number of those requests will get replaced with calls to `machine.count`.

## Fixes

Fixes: canonical-web-and-design/app-tribe#1138.

### DIFF
--- a/src/app/store/machine/slice.ts
+++ b/src/app/store/machine/slice.ts
@@ -1,6 +1,8 @@
 import type { PayloadAction } from "@reduxjs/toolkit";
 import { createSlice } from "@reduxjs/toolkit";
 
+import type { GenericMeta } from "../utils/slice";
+
 import { MachineMeta } from "./types";
 import type {
   Action,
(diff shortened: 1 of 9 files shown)


src/app/store/machine/slice.ts
```diff
@@ -977,44 +993,111 @@ const machineSlice = createSlice({
     exitRescueModeStart: statusHandlers.exitRescueMode.start,
     exitRescueModeSuccess: statusHandlers.exitRescueMode.success,
     fetch: {
-      prepare: (requestId: string) => ({
+      prepare: (requestId: string, params?: FetchParams) => ({
         meta: {
           model: MachineMeta.MODEL,
           method: "list",
+          nocache: true,
```
petermakowski, line 1000: We should probably find a way to keep the cache, atm the machine list is always requested at least twice (I have a ticket that's related to this https://github.com/canonical-web-and-design/maas-ui/issues/4274)

petermakowski (approved):

src/app/store/machine/slice.ts
huwshimi: I suspect we'll end up with just one request as the other requests will change to using the count endpoint.

PR #4313, opened 2022-08-03
Side identity - MAAS theme live preview

ndv99:
## Done

- Added a context for tracking the current MAAS theme
- Allowed users to preview a theme by selecting the radio button
- Added a "Cancel" button should a user change their mind

## QA
- Resets the MAAS theme to their existing choice if they refresh the page or navigate elsewhere

### QA steps

- Log in on the PR preview
- Preview a theme by clicking its radio button
- Save the theme and ensure it sticks if you refresh/navigate
- Preview a different theme
- Ensure it reverts back if you:
  - Hit "Cancel"
  - Navigate to another page
  - Refresh the page 

## Fixes

Fixes: https://github.com/canonical/app-tribe/issues/1133 .

### DIFF
--- a/cypress/e2e/with-users/settings/configuration/general.spec.ts
+++ b/cypress/e2e/with-users/settings/configuration/general.spec.ts
@@ -4,11 +4,64 @@ context("Settings Deploy configuration", () => {
   beforeEach(() => {
     cy.login();
     cy.visit(generateMAASURL("/settings/configuration/general"));
+
+    cy.findByRole("radio", { name: "Default" }).click();
+
+    cy.findByRole("button", { name: "Save" }).then(($btn) => {
+      if ($btn.is(":disabled")) {
+        return;
+      } else {
+        cy.wrap($btn).click();
+      }
+    });
   });
 
   it("displays a live preview of a MAAS theme colour", () => {
     cy.findByRole("radio", { name: "Red" }).click();
 
     cy.findByRole("banner").should("have.class", "p-navigation--red");
+  });
+
+  it("persists the theme choice after saving and refreshing the page", () => {
+    cy.findByRole("radio", { name: "Red" }).click();
+    cy.findByRole("button", { name: "Save" }).click();
+
+    cy.findByRole("button", { name: "Save" }).should("be.disabled");
+
+    cy.reload(true);
+
+    cy.findByRole("banner").should("have.class", "p-navigation--red");
+  });
+
+  it("persists the theme choice after saving and navigating away from page", () => {
+    cy.findByRole("radio", { name: "Red" }).click();
+    cy.findByRole("button", { name: "Save" }).click();
+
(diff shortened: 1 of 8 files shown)


petermakowski (commented):
This is a fairly clean solution to the problem, but sets a new precedent.

We generally try to stick to redux for global state (with a few exceptions). Let's check with @huwshimi before we merge.

Also - would be great to see some integration tests.

huwshimi:
> This is a fairly clean solution to the problem, but sets a new precedent.
> 
> We generally try to stick to redux for global state (with a few exceptions). Let's check with @huwshimi before we merge.

I don't really have strong feelings about this. It's a little bit of a special case. The only other place I can think of that we do any kind of UI state is the selected state for nodes: https://github.com/canonical/maas-ui/blob/main/src/app/store/machine/types/base.ts#L196.

I suppose this could live inside the `config` state in Redux as a `themePreview` attribute, which might require less special case code...

petermakowski:
> I suppose this could live inside the `config` state in Redux as a `themePreview` attribute, which might require less special case code...

We did discuss doing exactly that with @ndv99, but redux comes with lots of boilerplate, Context seems sufficient here IMO.

cypress/e2e/with-users/settings/configuration/general.spec.ts
```diff
@@ -0,0 +1,67 @@
+import { generateMAASURL } from "../../../utils";
+
+context("Settings Deploy configuration", () => {
```
petermakowski, line 3:
```suggestion
context("Settings - General", () => {
```

cypress/e2e/with-users/settings/configuration/general.spec.ts
```diff
@@ -0,0 +1,67 @@
...
+
+    cy.findByRole("radio", { name: "Default" }).click();
+
+    cy.findByRole("button", { name: "Save" }).then(($btn) => {
+      if ($btn.is(":disabled")) {
+        return;
+      } else {
+        cy.wrap($btn).click();
+      }
+    });
```
petermakowski, line 16: What's the reason for doing this before each test, shouldn't the default theme be selected by default? It would be a good idea to clean up after each test though (and revert to default theme).

petermakowski (commented):
Nicely done, your first cypress tests! 👍

cypress/e2e/with-users/settings/configuration/general.spec.ts
ndv99: So in my instance I'm using bolla for my tests, and the theme on that is subject to change - even with a local maas instance, the theme is also subject to change if I were to leave it set to something else while manually testing. This just makes sure that before each test the theme is 100% definitely set to default, and the `if`/`else` is there to prevent any errors if the theme is already set to default

petermakowski (approved):

PR #4319, opened 2022-08-08
Add unsubscribe hook and saga

huwshimi:
## Done

- Add support for unsubscribing to the websocket sagas.
- Add clean up and unsub to fetch and get hooks.

### MAAS deployment

- [Bolla](/HACKING.md#bolla)

### Running the branch

You can run this branch by:

### QA steps

- Go to the machine list.
- Look in `machine` in the redux store and you should see the `lists` requests.
- Navigate away from the machine list, preferably to somewhere that doesn't have machines.
- You should see the dispatches for `machine/unsubscribe` and state should no longer have the lists requests.
- Go to a machine details page and navigate away.
- You should see the dispatches for `machine/unsubscribe` and state should no longer have the details requests.

## Fixes

Fixes: canonical/app-tribe#1141.
Fixes: canonical/app-tribe#1128.

### DIFF
--- a/src/app/base/sagas/actions.ts
+++ b/src/app/base/sagas/actions.ts
@@ -178,24 +178,26 @@ export const generateNextDeleteRecordAction = (
 export function* deleteDomainRecord(
   socketClient: WebSocketClient,
   sendMessage: SendMessage,
-  { payload }: PayloadAction<{ params: DeleteRecordParams }>
+  { payload }: PayloadAction<{ params: DeleteRecordParams } | null>
 ): SagaGenerator<void> {
-  const { domain, rrset } = payload.params;
-  const initialAction = isAddressRecord(rrset.rrtype)
-    ? domainActions.deleteAddressRecord({
-        dnsresource_id: rrset.dnsresource_id,
-        domain,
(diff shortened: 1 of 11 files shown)


petermakowski (approved):

petermakowski:
On machine listing, sometimes only half of `machine.lists` are removed.

To reproduce: 
1. Go to machine listing page
2. refresh the page
3. navigate away to the settings page

On the machine details page all of `machine.details` are removed correctly.

huwshimi:
> On machine listing, sometimes only half of `machine.lists` are removed.

What was happening is that when the machine list loaded the fetch hook was running twice when the filters changed from `undefined` to `{}` which was causing a clean up of the first request and a dispatch for the second request, but I was being defensive in the reducers and creating the entry in the store if the response came back but the call couldn't be found in the store. However, because we were making two calls in quick succession the call came back after it was cleaned up which resulted in orphaned entries in the store.

I've fixed `{}` and falsy filters being equal so that the request only happens once and also fixed the orphaned entries issue. That all might more sense when you see the code :).

petermakowski (approved):

PR #4321, opened 2022-08-08
Migrate Enzyme to RTL - src/app/settings/views/Repositories

ndv99:
## Done

- [src/app/settings/views/Repositories/RepositoriesList/RepositoriesList.test.tsx](https://github.com/canonical/maas-ui/compare/main...ndv99:rtlMigration_settingsRepositories?expand=1#diff-e5ac03d46e56921272833fa3ced7491c5338324548b55b4f74d804c22d32495b)
- [src/app/settings/views/Repositories/RepositoryAdd/RepositoryAdd.test.tsx](https://github.com/canonical/maas-ui/compare/main...ndv99:rtlMigration_settingsRepositories?expand=1#diff-46efd7aaf0d85faf6d0221ec060bd50f0f023dd9c7ac74cbb95884a7fe8d7634)
- [src/app/settings/views/Repositories/RepositoryEdit/RepositoryEdit.test.tsx](https://github.com/canonical/maas-ui/compare/main...ndv99:rtlMigration_settingsRepositories?expand=1#diff-c5a2e1cbc3ee64f6843d3202d95e630961e607ef0c17721866cc233036f667a6)
- [src/app/settings/views/Repositories/RepositoryForm/RepositoryForm.test.tsx](https://github.com/canonical/maas-ui/compare/main...ndv99:rtlMigration_settingsRepositories?expand=1#diff-33cd0881228f00081bbc581817045c2aabca1d7072747f9a80d037397abf772f)
- [src/app/settings/views/Repositories/RepositoryFormFields/RepositoryFormFields.test.tsx](https://github.com/canonical/maas-ui/compare/main...ndv99:rtlMigration_settingsRepositories?expand=1#diff-53b01fb63020fbfeb140be19de5c0bee9257eedfe4ce74280b02a82e2c82cce6)

## Fixes

Closes https://github.com/canonical/app-tribe/issues/1055

### DIFF
--- a/src/app/settings/views/Repositories/RepositoriesList/RepositoriesList.test.tsx
+++ b/src/app/settings/views/Repositories/RepositoriesList/RepositoriesList.test.tsx
@@ -13,6 +13,7 @@ import {
   packageRepositoryState as packageRepositoryStateFactory,
   rootState as rootStateFactory,
 } from "testing/factories";
+import { renderWithMockStore } from "testing/utils";
 
 const mockStore = configureStore();
 
@@ -59,56 +60,50 @@ describe("RepositoriesList", () => {
 
   it("displays a loading component if loading", () => {
     state.packagerepository.loading = true;
-    const store = mockStore(state);
-    render(
-      <Provider store={store}>
-        <MemoryRouter
-          initialEntries={[
-            { pathname: "/settings/repositories", key: "testKey" },
-          ]}
-        >
-          <CompatRouter>
-            <RepositoriesList />
-          </CompatRouter>
-        </MemoryRouter>
-      </Provider>
+    renderWithMockStore(
+      <MemoryRouter
+        initialEntries={[
+          { pathname: "/settings/repositories", key: "testKey" },
+        ]}
+      >
+        <CompatRouter>
+          <RepositoriesList />
+        </CompatRouter>
+      </MemoryRouter>,
+      { state }
     );
     expect(screen.getByText("Loading")).toBeInTheDocument();
   });
 
   it("shows the table if there are repositories and loaded is true", () => {
     state.packagerepository.loaded = true;
-    const store = mockStore(state);
-    render(
-      <Provider store={store}>
-        <MemoryRouter
-          initialEntries={[
-            { pathname: "/settings/repositories", key: "testKey" },
-          ]}
-        >
-          <CompatRouter>
-            <RepositoriesList />
-          </CompatRouter>
-        </MemoryRouter>
-      </Provider>
+    renderWithMockStore(
+      <MemoryRouter
+        initialEntries={[
+          { pathname: "/settings/repositories", key: "testKey" },
+        ]}
+      >
+        <CompatRouter>
+          <RepositoriesList />
+        </CompatRouter>
+      </MemoryRouter>,
+      { state }
     );
     expect(screen.getByLabelText("Package repositories")).toBeInTheDocument();
   });
 
   it("can show a delete confirmation", async () => {
-    const store = mockStore(state);
-    render(
-      <Provider store={store}>
-        <MemoryRouter
-          initialEntries={[
-            { pathname: "/settings/repositories", key: "testKey" },
-          ]}
-        >
-          <CompatRouter>
-            <RepositoriesList />
-          </CompatRouter>
-        </MemoryRouter>
-      </Provider>
+    renderWithMockStore(
+      <MemoryRouter
+        initialEntries={[
+          { pathname: "/settings/repositories", key: "testKey" },
+        ]}
+      >
+        <CompatRouter>
+          <RepositoriesList />
+        </CompatRouter>
+      </MemoryRouter>,
+      { state }
     );
 
     expect(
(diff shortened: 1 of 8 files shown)


src/app/settings/views/Repositories/RepositoriesList/RepositoriesList.test.tsx
```diff
@@ -109,17 +110,31 @@ describe("RepositoriesList", () => {
...
-    let row = wrapper.find("TableRow[data-testid='repository-row']").at(2);
-    expect(row.hasClass("is-active")).toBe(false);
-    // Click on the delete button:
-    wrapper.find("TableRow").at(3).find("Button").at(1).simulate("click");
-    row = wrapper.find("TableRow[data-testid='repository-row']").at(2);
-    expect(row.hasClass("is-active")).toBe(true);
+
+    expect(
+      screen.queryByRole("row", {
+        name: "secret_archive http://www.website.com No Edit Delete",
```
huwshimi, line 116: This feels a little fragile. It might be better to add an aria-label to the row. Same for the other instances of `screen.queryByRole("row" ...` in this file.

src/app/settings/views/Repositories/RepositoriesList/RepositoriesList.test.tsx
```diff
@@ -133,20 +148,29 @@ describe("RepositoriesList", () => {
       </Provider>
     );
     // Click on the delete button:
-    wrapper.find("TableRow").at(3).find("Button").at(1).simulate("click");
+    await userEvent.click(
+      within(
+        screen.getByRole("row", {
+          name: "secret_archive http://www.website.com No Edit Delete",
+        })
+      ).getByTestId("table-actions-delete")
```
huwshimi, line 156: This might be better to be `getByRole("button", {name: Label.Delete})` (which will require adding a Label enum in TableActions.tsx).

src/app/settings/views/Repositories/RepositoriesList/RepositoriesList.test.tsx
```diff
@@ -133,20 +148,29 @@ describe("RepositoriesList", () => {
...
-      .find("TableRow")
-      .at(3)
-      .find("ActionButton[data-testid='action-confirm']")
-      .simulate("click");
+    await userEvent.click(
+      within(
+        screen.getByRole("row", {
+          name: `secret_archive http://www.website.com No Edit Delete Are you sure you want to delete repository "secret_archive"? This action is permanent and can not be undone. Cancel Delete`,
+        })
+      ).getByTestId("action-confirm")
```
huwshimi, line 165: Same here about targeting the role.

src/app/settings/views/Repositories/RepositoryFormFields/RepositoryFormFields.test.tsx
```diff
@@ -60,10 +61,21 @@ describe("RepositoryFormFields", () => {
...
+    expect(
+      screen.getByRole("textbox", { name: "Distributions" })
+    ).toBeInTheDocument();
+
+    expect(
+      screen.getByRole("textbox", { name: "Components" })
+    ).toBeInTheDocument();
+  });
+
+  it("doesn't display disitribution and component inputs if type is ppa", () => {
```
huwshimi, line 74: "disitribution" :)

huwshimi (approved):
I see you've added some Label enums here which is great! I'd be good practice to use those for the other files in this branch. I know it's a little more work but it'll make these tests more robust. It'd also be great to use the `renderWith...` functions if possible when migrating these tests.

src/app/settings/views/Repositories/RepositoriesList/RepositoriesList.test.tsx
ndv99: The reason I didn't do that for this case is that there are two delete buttons for every row, both with the name "Delete", so getting them by role this way returns two of them. I could do getAllByRole and get them by list index but that feels a little janky to me

src/app/settings/views/Repositories/RepositoriesList/RepositoriesList.test.tsx
ndv99: (see above comment)

src/app/settings/views/Repositories/RepositoriesList/RepositoriesList.test.tsx
ndv99: I'm not 100% sure on how to do this, I think we'd have to modify the MainTable React component, specifically MainTableRow, to take an aria-label as a prop

src/app/settings/views/Repositories/RepositoryFormFields/RepositoryFormFields.test.tsx
ndv99: :man_facepalming: :joy:

src/app/settings/views/Repositories/RepositoriesList/RepositoriesList.test.tsx
huwshimi: Ah OK. One thing you could try is to add an `aira-label` to [each column that match the headers](https://github.com/canonical/maas-ui/blob/main/src/app/subnets/components/ReservedRanges/ReservedRanges.tsx#L124)  which you could then use to target the actions column: `.getAllByRole("gridcell", { name: Label.Actions })` and then find the delete button inside the column.

src/app/settings/views/Repositories/RepositoriesList/RepositoriesList.test.tsx
huwshimi: You should just be able to add it to the [row's object](https://github.com/canonical/maas-ui/blob/main/src/app/settings/views/Repositories/RepositoriesList/RepositoriesList.tsx#L37) which gets passed in addition to the defined row props: https://github.com/canonical/react-components/blob/main/src/components/MainTable/MainTable.tsx#L232.

ndv99: I've tried that, but if I use "aria-label", the code thinks it's an arithmetic operation (which makes sense because it's using a `-` in an object and not an HTML tag), and if I use `ariaLabel`, I get the following error:

`Warning: Invalid ARIA attribute `ariaLabel`. Did you mean `aria-label`?`

Which makes me think that the only remaining way to make it work is to modify that react component

ndv99: Ignore that, I forgot you can add props with the name wrapped in quotation marks :joy: :man_facepalming:

src/app/settings/views/Repositories/RepositoriesList/RepositoriesList.test.tsx
ndv99: Got this sorted!

PR #4344, opened 2022-08-15
feat(machines): paginate using the new API

huwshimi:
## Done

- Add pagination to the machine list using the new API.

### MAAS deployment

- [Bolla](/HACKING.md#bolla)

### Running the branch

You can run this branch by:

### QA steps

To test this you either need more than one page of machines using [sampledata](https://github.com/canonical/maas-ui/blob/main/docs/HACKING.md#sample-data) or change [`const pageSize = DEFAULT_PAGE_SIZE;`](https://github.com/canonical/maas-ui/pull/4344/files#diff-2b86f4686bb74434cdb1fafb6f4ee8837c1cfa1afaf9d92934a85ee1a90072c8R56) in `MachineList.tsx` to something smaller than the number of machines you have.
- Go to the machine list.
- Scroll down and should see the pagination.
- Open the redux dev tools and go to `machine` -> `lists` and you should see a few requests.
- Clicking on the pagination buttons should navigate through the list and you should see one of the requests getting removed and replaced with a new one.

## Fixes

Fixes: canonical/app-tribe#1101.
Fixes: canonical/app-tribe#1121.

### DIFF
--- a/src/app/machines/views/MachineList/MachineList.tsx
+++ b/src/app/machines/views/MachineList/MachineList.tsx
@@ -75,7 +75,7 @@ const MachineList = ({
     "grouping",
     FetchGroupKey.Status
   );
-  const pageSize = DEFAULT_PAGE_SIZE;
+  const pageSize = DEFAULTS.pageSize;
   const { machineCount, machines, machinesErrors } = useFetchMachines(
     parseFilters(filters),
     grouping,
--- a/src/app/machines/views/MachineList/MachineList.tsx
+++ b/src/app/machines/views/MachineList/MachineList.tsx
@@ -62,6 +62,7 @@ const MachineList = ({
   const dispatch = useDispatch();
   const errors = useSelector(machineSelectors.errors);
   const selectedIDs = useSelector(machineSelectors.selectedIDs);
+  const [currentPage, setCurrentPage] = useState(1);
   const [sortKey, setSortKey] = useState<FetchGroupKey | null>(
     DEFAULTS.sortKey
   );
@@ -74,9 +75,12 @@ const MachineList = ({
     "grouping",
     FetchGroupKey.Status
   );
-  const { machines, machinesErrors } = useFetchMachines(
+  const pageSize = DEFAULT_PAGE_SIZE;
+  const { machineCount, machines, machinesErrors } = useFetchMachines(
     parseFilters(filters),
     grouping,
+    pageSize,
+    currentPage,
     sortKey,
     mapSortDirection(sortDirection)
   );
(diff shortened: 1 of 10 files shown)


src/app/machines/views/MachineList/MachineList.tsx
```diff
@@ -45,15 +45,19 @@ const MachineList = ({
   const dispatch = useDispatch();
   const errors = useSelector(machineSelectors.errors);
   const selectedIDs = useSelector(machineSelectors.selectedIDs);
+  const [currentPage, setCurrentPage] = useState(1);
```
petermakowski, line 48: Might be worth covering this with tests.

petermakowski:
Kapture 2022-08-18 at 09 27 23

When navigating to another page, the entire table disappears including the pagination controls which feels like a suboptimal user experience.

It might be a good idea to keep the current page displayed until the new one is ready while indicating the loading state in the meantime instead (e.g. only if it's taking longer than 2 seconds).
Alternatively we could use a skeleton pattern that's mentioned here: https://github.com/canonical/app-tribe/issues/1104#issuecomment-1216348969). Might be good to get @amylily1011 thoughts on this.

Ideally we'd want to make the loading transparent and not disruptive to the user experience. For reference - this is how pagination behaves works on the `main` branch (where we load all results in advance, but a similar result should be possible with the new API):

Kapture 2022-08-18 at 09 30 51

amylily1011:
@petermakowski I think it's a great idea. When the pagination is gone after click it looks really disruptive. I think the skeleton loading is a perfect place for this as well and it will make the experience more consistent if we use the skeleton loading for tables.

huwshimi:
> When navigating to another page, the entire table disappears including the pagination controls which feels like a suboptimal user experience.
> 
> It might be a good idea to keep the current page displayed until the new one is ready while indicating the loading state in the meantime instead (e.g. only if it's taking longer than 2 seconds). Alternatively we could use a skeleton pattern that's mentioned here: [canonical/app-tribe#1104 (comment)](https://github.com/canonical/app-tribe/issues/1104#issuecomment-1216348969)). Might be good to get @amylily1011 thoughts on this.

Oh yeah there's actually an issue proposing the skeleton pattern here: https://github.com/canonical/app-tribe/issues/1099 and an issue to implement it here: https://github.com/canonical/app-tribe/issues/1100.

src/app/machines/views/MachineList/MachineList.tsx
huwshimi: Done.

PR #4361, opened 2022-08-22
feat(machines): pass loading states to machine list

huwshimi:
## Done

- Pass the loading state from the list request to the machine list header and table.

### MAAS deployment

- [Bolla](/HACKING.md#bolla)

### Running the branch

You can run this branch by:

### QA steps

- Go to the machine list.
- You should see a spinner in the header and table while the list and counts load.
- Do a search (e.g. `owner:admin`). You should see a spinner in the table (this will be replaced by the loading skeleton in https://github.com/canonical/app-tribe/issues/1100.).

## Fixes

Fixes: canonical/app-tribe#1280.

### DIFF
--- a/src/app/store/machine/selectors.ts
+++ b/src/app/store/machine/selectors.ts
@@ -431,6 +431,28 @@ const listGroups = createSelector(
 );
 
 /**
+ * Get the loaded state for a machine list request with a given callId.
+ */
+const listLoaded = createSelector(
+  [
+    machineState,
+    (_state: RootState, callId: string | null | undefined) => callId,
+  ],
+  (machineState, callId) => getList(machineState, callId)?.loaded ?? null
+);
+
+/**
+ * Get the loading stateo for a machine list request with a given callId.
+ */
+const listLoading = createSelector(
+  [
+    machineState,
+    (_state: RootState, callId: string | null | undefined) => callId,
+  ],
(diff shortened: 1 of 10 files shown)


src/app/store/machine/selectors.ts
```diff
@@ -386,6 +386,28 @@ const listCount = createSelector(
...
 
+/**
+ * Get the loaded state for a machine list request with a given callId.
+ */
+const listLoaded = createSelector(
+  [
+    machineState,
+    (_state: RootState, callId: string | null | undefined) => callId,
+  ],
+  (machineState, callId) => getList(machineState, callId)?.loaded ?? null
```
petermakowski, line 397: Do we need the additional state complexity by adding `null`? Couldn't we just default to `false`?

petermakowski (approved):

src/app/store/machine/selectors.ts
petermakowski: In case we want to keep the `null`, would be good to add tests coverage for it.

huwshimi: Hmm maybe we don't need this. The `null` case would be if we're trying to select a list that isn't in the store for some reason, but I can't think of a case where we'd need to know that for the loading/loaded states. We can always add this back in if we find we need it.

PR #4365, opened 2022-08-25
feat(machines): parse search DSL and send to new API

huwshimi:
## Done

- Parse the search DSL string and send the filters to the API call.

### MAAS deployment

- [Bolla](/HACKING.md#bolla)

### Running the branch

You can run this branch by:

### QA steps

- Go to the machine list.
- Type in some text that will match some param of a machine e.g. part of a hostname, a user's name etc. and the table should update with the results.
- Type in a search for a specific param e.g. status: new, tags:..., hostname:... etc. and the table should update.
- Add a workload annotation (if using the demo service I've added an annotation so you can skip to the last step):
  - go to the network tab and find the websocket URL.
  - Open the console and run the following with the websocket URL: `const websocket = new WebSocket("<MAAS_WEBSOCKET_URL>");`
  - Find the system id of a machine.
  - Now run the following with the system id:
```
websocket.send(
  JSON.stringify({
    method: "machine.set_workload_annotations",
    params: {
      workload_annotations: {
        type: "production",
      },
      system_id: "<SYSTEM_ID_OF_MACHINE>",
    },
    request_id: 999,
    type: 0,
  })
);
```
  - Now go to the machine list and type `workload-type:prod` and you should see the machine.

## Fixes

Fixes: canonical/app-tribe#1263.

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -97,7 +97,6 @@ module.exports = {
           },
         ],
         "react/jsx-sort-props": "error",
-        "react/no-multi-comp": ["error", { ignoreStateless: false }],
       },
       settings: {
         "import/resolver": {
@@ -115,6 +114,12 @@ module.exports = {
       files: ["src/**/*.js?(x)"],
       rules: {
         "no-unused-vars": 2,
+      },
+    },
+    {
+      files: ["src/**/*.tsx"],
+      rules: {
+        "react/no-multi-comp": ["error", { ignoreStateless: false }],
       },
     },
     {
--- a/src/app/machines/views/MachineList/MachineList.test.tsx
+++ b/src/app/machines/views/MachineList/MachineList.test.tsx
@@ -28,8 +28,9 @@ import {
   machineStateList as machineStateListFactory,
   machineStateListGroup as machineStateListGroupFactory,
 } from "testing/factories";
+import { renderWithBrowserRouter } from "testing/utils";
 
-const mockStore = configureStore();
+const mockStore = configureStore<RootState, {}>();
 
 jest.useFakeTimers();
 
@@ -374,6 +375,7 @@ describe("MachineList", () => {
       </Provider>
     );
     expect(wrapper.find("Notification").exists()).toBe(true);
+    // eslint-disable-next-line testing-library/no-node-access
     expect(wrapper.find("Notification").props().children).toBe("Uh oh!");
   });
 
@@ -393,6 +395,7 @@ describe("MachineList", () => {
       </Provider>
     );
     expect(wrapper.find("Notification").exists()).toBe(true);
+    // eslint-disable-next-line testing-library/no-node-access
     expect(wrapper.find("Notification").props().children).toBe(
       "tag: No such constraint."
     );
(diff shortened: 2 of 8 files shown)


.eslintrc.js
```diff
@@ -117,6 +116,12 @@ module.exports = {
         "no-unused-vars": 2,
       },
     },
+    {
+      files: ["src/**/*.tsx"],
+      rules: {
+        "react/no-multi-comp": ["error", { ignoreStateless: false }],
```
huwshimi, line 122: I changed this to only `tsx` files as it was getting false positives from functions that return functions so hopefully this might help reduce that.

src/app/machines/views/MachineList/MachineList.test.tsx
```diff
@@ -374,6 +375,7 @@ describe("MachineList", () => {
       </Provider>
     );
     expect(wrapper.find("Notification").exists()).toBe(true);
+    // eslint-disable-next-line testing-library/no-node-access
```
huwshimi, line 378: This whole file needs to be migrated to RTL, but in the meantime I've disabled these warnings.

src/app/store/machine/types/actions.ts
```diff
@@ -254,54 +254,53 @@ export enum FetchSortDirection {
...
-  ip_addresses: NodeIpAddress["ip"];
-  spaces: Space["name"];
   workloads: string;
-  status: FetchNodeStatus;
-}> & {
-  mem: string;
-  cpu_count: string;
-  cpu_speed: string;
-  link_speed: string;
-};
```
huwshimi, line 287: This was mostly ordering by alphabet, but the `mem`, `cpu_count` etc. filters have been moved into the `ArrayOrValue` block as the API now supports arrays for all the filters.

petermakowski:
Not sure if it's something that I did wrong in following instructions (there's a room for error for sure although I did try more than once) or something broke on the API end at some point, because I wasn't able to QA this successfully (no machines found).

huwshimi:
> Not sure if it's something that I did wrong in following instructions (there's a room for error for sure although I did try more than once) or something broke on the API end at some point, because I wasn't able to QA this successfully (no machines found).

Hmm strange! It's working today for me both on the demo service and locally. Not sure if something has changed since you tried! I did notice that in my QA instructions `workload-service:prod` should have been `workload-type:prod` (I've updated the instructions), but that should have only affected the last bit of the QA.

@ndv99 any chance you wouldn't mind trying the QA for this one?

ndv99:
> @ndv99 any chance you wouldn't mind trying the QA for this one?

Managed to QA it. All looks good to me, got through the steps and got the expected behaviour. One thing I noticed though was when using the filter dropdown to select a filter instead of typing it, it types in the text wrong, e.g. instead of `status:failed_testing` (which correctly shows the machine that failed testing), it puts in `status:(=failed_testing)`, which results in an error message for an invalid filter choice

huwshimi:
> > @ndv99 any chance you wouldn't mind trying the QA for this one?
> 
> Managed to QA it. All looks good to me, got through the steps and got the expected behaviour. One thing I noticed though was when using the filter dropdown to select a filter instead of typing it, it types in the text wrong, e.g. instead of `status:failed_testing` (which correctly shows the machine that failed testing), it puts in `status:(=failed_testing)`, which results in an error message for an invalid filter choice

Great! Thanks. There is a fix coming for those filters on the API side.

PR #4366, opened 2022-08-25
fix(UI): node overview

ndv99:
## Note

The commits in this PR are being merged by #4377. 

## Done

- Extend vertical grid dividers to top of summary card
- Add padding on either side of horizontal divider on summary card
- Details items (Owner, Host, Zone etc) have a max gap of 1rem
- If there is extra space, tags are allowed to stop truncating first
- Align test failed text to other links in top half of overview
- Redistribute space after tags finish truncating
- Hide "Host" on physical machines that don't have a VM host

### MAAS deployment

- [Bolla](/HACKING.md#bolla)

### Running the branch

You can run this branch by:

### QA steps

- Navigate to Machines and click a machine
- Stretch the window horizontally and ensure that the tags are allowed to truncate
- Squeeze the window horizontally and ensure none of the details (labels or info) are wrapping before the layout changes

## Fixes

Fixes https://github.com/canonical/app-tribe/issues/1281

### Before
Screenshot from 2022-08-25 09-42-56

### After
image

### DIFF
--- a/src/app/base/components/node/OverviewCard/MachineStatusCard/MachineStatusCard.tsx
+++ b/src/app/base/components/node/OverviewCard/MachineStatusCard/MachineStatusCard.tsx
@@ -131,14 +131,17 @@ const MachineStatusCard = ({ machine }: Props): JSX.Element => {
       </div>
       {showFailedTestsWarning(machine) ? (
         <div
-          className="overview-card__test-warning u-flex-bottom"
+          className="overview-card__test-warning u-flex--vertically"
           data-testid="failed-test-warning"
         >
-          <i className="p-icon--warning">Warning:</i>
-          <span className="u-nudge-right--x-small">
-            {" "}
-            Some tests failed, use with caution.
-          </span>
+          <ul className="overview-card__test-warning-text p-inline-list u-no-margin--bottom">
+            <li className="p-inline-list__item">
+              <span>
+                <i className="p-icon--warning">Warning:</i> Some tests failed,
+                use with caution.
+              </span>
+            </li>
+          </ul>
         </div>
       ) : null}
     </>
--- a/src/app/base/components/node/OverviewCard/MachineStatusCard/MachineStatusCard.tsx
+++ b/src/app/base/components/node/OverviewCard/MachineStatusCard/MachineStatusCard.tsx
@@ -134,7 +134,7 @@ const MachineStatusCard = ({ machine }: Props): JSX.Element => {
           className="overview-card__test-warning u-flex--vertically"
           data-testid="failed-test-warning"
         >
-          <ul className="overview-card__test-warning-text p-inline-list u-no-margin--bottom">
+          <ul className="overview-card__test-text p-inline-list u-no-margin--bottom">
             <li className="p-inline-list__item">
               <span>
                 <i className="p-icon--warning">Warning:</i> Some tests failed,
--- a/src/app/base/components/node/OverviewCard/_index.scss
+++ b/src/app/base/components/node/OverviewCard/_index.scss
@@ -72,14 +72,14 @@
       }
 
       &.for-controller {
-        grid-template-columns: minmax(0, 1fr) minmax(0, 1fr) minmax(0, 1fr) minmax(
+        grid-template-columns: minmax(0, 1fr) minmax(0, 1fr) minmax(
             0,
             4fr
           );
       }
 
       &.for-machine {
-        grid-template-columns: repeat(7, minmax(0, 1fr));
+        grid-template-columns: minmax(0, 8rem) minmax(0, 8rem) minmax(0, 8rem) minmax(0, 8rem) minmax(0, 8rem) minmax(0, 3fr);
         padding-left: 0;
       }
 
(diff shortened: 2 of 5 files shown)


sparkiegeek:
I've noticed recently (?) that for non VMs we have a "Host: None" field in the overview section which is unexpected.

image

We should just omit the Host field when we're on a physical machine that doesn't have a VM host.

ndv99:
> I've noticed recently (?) that for non VMs we have a "Host: None" field in the overview section which is unexpected.
> 
> (image)
> 
> We should just omit the Host field when we're on a physical machine that doesn't have a VM host.

That's fine yeah, I can add that as a todo for this PR

huwshimi:
@ndv99 I might hold off reviewing this one until @lyubomir-popov has given the OK.

src/app/base/components/node/OverviewCard/_index.scss
```diff
@@ -13,92 +13,96 @@
...
+
+    .overview-card__memory-test-text {
+      margin-bottom: 0.1rem;
+      padding-top: calc(0.4rem - 1px);
+    }
+
+    .overview-card__cpu-test-text {
+      margin-bottom: 0.1rem;
+      padding-top: calc(0.4rem - 1px);
+    }
```
huwshimi, line 62: These classes all look like they're setting the same things. Could it be a single class that is applied to all of the columns?

This is also using custom spacing params. Could they be based off the spacing variables instead?

src/app/base/components/node/OverviewCard/_index.scss
```diff
@@ -13,92 +13,96 @@
...
-            0,
-            4fr
-          );
-      }
-
-      &.for-machine {
-        grid-template-columns: repeat(7, minmax(0, 1fr));
-      }
+      padding: $spv--large 0;
+      margin: 0 1rem 0 1rem ;
```
huwshimi, line 90: This looks like it should be using spacing variables and also for all the other 1rems in this file.

src/app/base/components/node/OverviewCard/MachineStatusCard/MachineStatusCard.tsx
```diff
@@ -131,14 +131,17 @@ const MachineStatusCard = ({ machine }: Props): JSX.Element => {
...
-          className="overview-card__test-warning u-flex-bottom"
+          className="overview-card__test-warning u-flex--vertically"
           data-testid="failed-test-warning"
         >
-          <i className="p-icon--warning">Warning:</i>
-          <span className="u-nudge-right--x-small">
-            {" "}
-            Some tests failed, use with caution.
-          </span>
+          <ul className="overview-card__test-text p-inline-list u-no-margin--bottom">
```
huwshimi, line 137: Is this using a `ul` to get the right spacing? I'm not sure it's the right element in this case.

It actually looks like it could be using the [borderless](https://vanillaframework.io/docs/patterns/notification#borderless) notification (it could be a followup to fix this if need be).

huwshimi (approved):
I few suggestions here. One thing I noticed is that the tags are wrapping instead of truncating like your screenshot:

Screen Shot 2022-08-31 at 1 57 50 pm

src/app/base/components/node/OverviewCard/_index.scss
ndv99: The class can definitely be simplified, that's easy enough. As for spacing variables, I guess they could be done as a calculation of them? e.g. `$spv--large * 0.1` for 0.1rem and `$spv--large * 0.4* for 0.4rem

src/app/base/components/node/OverviewCard/MachineStatusCard/MachineStatusCard.tsx
huwshimi: Ah OK.

src/app/base/components/node/OverviewCard/_index.scss
huwshimi: I guess that would at least scale with the spacing variables though usually I find that if we're having to use very specific spacing then something else is amiss (i.e. using the wrong class or element or applying the spacing to the wrong element).

I did notice that the element with `overview-card__test-text` also has `u-no-margin--bottom` so the `margin-bottom: 0.1rem;` is being overridden, so you could probably remove the margin-bottom styles.

Also, just looking at the figma file it looks like there might just be a 1 rem gap between these links and the content above where as this looks more like 2.5 rem. If that's correct you might be able to drop the top padding and also the bottom padding on these elements: https://github.com/canonical/maas-ui/pull/4366/files#diff-b7088aac56f8eab6260081604edc7246aa32511490d6b86d8f9c27b1292da145L50.

It could be worth double checking with @lyubomir-popov (if you haven't already chatted about this - I might have missed a conversation about this spacing). Either way feel free to land this with whatever you come up with.

ndv99: I've gone and removed the margin-bottom stuff as well as that top and bottom padding, and it looks like that's allowed me to delete that padding calculation entirely. Amy's asked me to merge this and bios boot mode into a separate branch for QA so they can see it all in one, should finally be able to get this merged after that

ndv99:
This PR will be deleted, since its commits are being merged in #4377, which combines another PR that has design changes on the same page.

PR #4368, opened 2022-08-26
refactor(tags): use new API for deployed machines table

huwshimi:
## Done

- Update the table for displaying deployed machines for a tag to use the new API.

### MAAS deployment

- [Bolla](/HACKING.md#bolla)

### Running the branch

You can run this branch by:

### QA steps

- Go to the machine list and tick some deployed machines.
- Use the take action menu to open the tag form.
- type in a new tag name and click the create tag button to open the modal.
- type something into the kernel options field and click create.
- Save the tag form.
- Go to the tags tab and click the delete icon next to the tag you just created.
- Inside the kernel options warning click "Show the deployed machines".
- The table should appear with the machines you just tagged.

## Fixes

Fixes: canonical/app-tribe#1120.


src/app/tags/views/TagMachines/TagMachines.tsx
```diff
@@ -1,38 +1,61 @@
-import { useEffect } from "react";
+import { useEffect, useState } from "react";
 
+import type { ValueOf } from "@canonical/react-components";
 import { Spinner } from "@canonical/react-components";
 import { useDispatch, useSelector } from "react-redux";
 
+import ErrorsNotification from "../../../machines/views/MachineList/ErrorsNotification";
```
petermakowski, line 7: Could use absolute import for consistency

src/app/tags/views/TagMachines/TagMachines.tsx
```diff
@@ -1,38 +1,61 @@
...
+  let filters: FetchFilters = { status: FetchNodeStatus.DEPLOYED };
+  if (tag) {
+    filters.tags = [tag.name];
+  }
+  const { callId, loading, machineCount, machines, machinesErrors } =
+    useFetchMachines({
+      currentPage,
+      filters,
+      pageSize: PAGE_SIZE,
+      sortDirection: mapSortDirection(sortDirection),
```
petermakowski, line 56: `useFetchMachines` hook should handle value mapping by default IMO.

src/app/store/machine/utils/common.ts
```diff
@@ -71,3 +75,19 @@ export const getHasSyncFailed = (machine?: Machine | null): boolean => {
...
+): FetchSortDirection | null => {
+  switch (sortDirection) {
+    case SortDirection.ASCENDING:
+      return FetchSortDirection.Ascending;
+    case SortDirection.DESCENDING:
+      return FetchSortDirection.Descending;
+    default:
+      return null;
+  }
+};
```
petermakowski, line 93: This layer of abstraction could be avoided - SortDirection and FetchSortDirection values are exactly the same with the exception of `null`/`none`. I don't believe we will be adding another sort direction anytime soon and might as well use the same values on the front end and back-end.

petermakowski (approved):

src/app/tags/views/TagMachines/TagMachines.tsx
huwshimi: Whoops! Thanks.

src/app/tags/views/TagMachines/TagMachines.tsx
huwshimi: I miiiight leave this until we see how the hook gets used in the DHCP and clone forms etc. as this might tie the hook to the table sort but we might not use that in all places (or maybe we will).

I've added an issue to do the update once we've taken a look at the forms: https://github.com/canonical/app-tribe/issues/1320.

src/app/store/machine/utils/common.ts
huwshimi: It's even worse because it's based on a type in react-components: https://github.com/canonical/react-components/blob/fd1b4e25ea97a88f205b4d8f2b0f03d8f6301302/src/types/index.ts#L19 but we also have our own version: https://github.com/canonical/maas-ui/blob/main/src/app/base/types.ts#L17.

I can't remember if/why MainTable needed the "none" option but I've added an issue to investigate to see if we can clean up these types: https://github.com/canonical/app-tribe/issues/1321 (until then we'll need this mapping so that we're not sending "none" to the backend).

PR #4376, opened 2022-09-01
feat(machines): collapse groups via the API

huwshimi:
## Done

- Update the machine list to toggle collapsed groups via the API.

### MAAS deployment

- [Bolla](/HACKING.md#bolla)

### Running the branch

You can run this branch by:

### QA steps

- Go to the machine list.
- On one of the group headers click the "-" button.
- The machine list should reload and the group should be collapsed.
- Refresh the page and the machine list should load with the group collapsed.

## Fixes

Fixes: canonical/app-tribe#1123.


src/app/machines/views/MachineList/MachineList.tsx
```diff
@@ -46,20 +46,21 @@ const MachineList = ({
     "grouping",
     FetchGroupKey.Status
   );
+  const [hiddenGroups, setHiddenGroups] = useStorageState<(string | null)[]>(
+    localStorage,
+    "hiddenGroups",
+    []
+  );
```
petermakowski, line 53: I wonder if we need this functionality and if it's worth the additional complexity. What's the use case for this?

And if so - why local storage, shouldn't it be persisted in the URL like the rest of parameters?

huwshimi: > I wonder if we need this functionality and if it's worth the additional complexity. What's the use case for this?

I'm not sure of the use cases (I can't seem to find any of the docs from when the feature was designed). I remember it was discussed a couple of times about whether it could be dropped for this rework as it required API support, but it was decided it was required (see the [acceptance criteria]).(https://docs.google.com/document/d/1B8CQXRxzapUq8dbSILzCZdrNzU49brVS8K5KQAZeQiw/edit#heading=h.weidp47i2rfi).


> And if so - why local storage, shouldn't it be persisted in the URL like the rest of parameters?

It uses local storage so that when users revisit the machine list the [collapsed groups are persisted](https://docs.google.com/document/d/19yTeNST6cp_w6xE4zP-KPIi4Om6aT-FXvKCqBd5tKJw/edit#heading=h.k6grss44lvox) (as opposed to URL params which don't get persisted) - it was originally discussed that this might become a user setting saved in the DB if we get support for such things so that it gets persisted across logins.

petermakowski (approved):

PR #4391, opened 2022-09-07
test(machines): add test for no search results

huwshimi:
## Done

- Add a test for when there are no search results in the machine list (I couldn't reproduce this message not appearing - it can take a while if you enter a number of filters that don't match anything so it's possible I was just being impatient).

## QA

n/a

## Fixes

Fixes: canonical/app-tribe#1327.


src/app/machines/views/MachineList/MachineListTable/MachineListTable.tsx
```diff
@@ -49,6 +49,7 @@ import { actions as zoneActions } from "app/store/zone";
 export enum Label {
   Loading = "Loading machines",
   Machines = "Machines",
+  NoResults = "No machines match the search criteria.",
```
petermakowski, line 52: I'm of the opinion it's helpful to make the name reflect the actual value (e.g. `NoMachines` instead of `NoResults`) where that makes sense.

This - arguably - makes for better readability as it doesn't require you to jump through sections of the code as much as you're reading it to understand what will be displayed.

What are your thoughts on this @huwshimi ?

petermakowski (approved):
Looks good. One thing that would be good to get your thoughts on.

PR #4400, opened 2022-09-08
Migrate Enzyme to RTL - src/app/dashboard

ndv99:
## Done

- [src/app/dashboard/views/DashboardHeader/ClearAllForm/ClearAllForm.test.tsx](https://github.com/canonical/maas-ui/compare/main...ndv99:rtlMigration_dashboard?expand=1#diff-61bb807e7eb19e21fba90ca17891d227acd88a920f652ac52fcb35c2f34e1365)
- [src/app/dashboard/views/DashboardHeader/DashboardHeader.test.tsx](https://github.com/canonical/maas-ui/compare/main...ndv99:rtlMigration_dashboard?expand=1#diff-412c376be80a5465c55dc86cc4b79a0fb9198843ab492207d98f561b80e7152c)
- [src/app/dashboard/views/DiscoveryAddForm/DiscoveryAddForm.test.tsx](https://github.com/canonical/maas-ui/compare/main...ndv99:rtlMigration_dashboard?expand=1#diff-c69888ce48fc3fbddcabe3d0de024b340fbed40da093dc626641083f381d460d)
- [src/app/dashboard/views/DiscoveryAddForm/DiscoveryAddFormFields/DiscoveryAddFormFields.test.tsx](https://github.com/canonical/maas-ui/compare/main...ndv99:rtlMigration_dashboard?expand=1#diff-b8f22684143c1355380179d91486c2b029382a549d03467a2e11b1ad75d33813)
- [src/app/dashboard/views/DashboardConfigurationForm/DashboardConfigurationSubnetForm/DashboardConfigurationSubnetForm.test.tsx](https://github.com/canonical/maas-ui/compare/main...ndv99:rtlMigration_dashboard?expand=1#diff-9ae399b4fe88523131cd4615caad111f0877eeac7e8a99673b9c56dc6c94ff22)

## Fixes

Fixes: https://github.com/canonical/app-tribe/issues/1311


src/app/dashboard/views/DiscoveryAddForm/DiscoveryAddForm.test.tsx
```diff
@@ -84,61 +141,68 @@ describe("DiscoveryAddForm", () => {
...
-      </Provider>
+    renderWithBrowserRouter(
+      <DiscoveryAddForm discovery={discovery} onClose={jest.fn()} />,
+      { route: "/dashboard", wrapperProps: { store } }
     );
-    expect(wrapper.find("Spinner").exists()).toBe(true);
+    expect(screen.getByText("Loading")).toBeInTheDocument();
   });
 
-  it("maps name errors to hostname", () => {
```
ndv99, line 101: @huwshimi Could you take a look at this one for me? With RTL we can't directly access this prop like the original test tries to, and the only visual way of telling there's an error is to see if the save button is disabled

huwshimi: This was a bit confusing to figure out. What this test is intended to do is test that name errors returned by the API are displayed (by mapping them to the hostname field). However, I discovered a little issue with how the formik errors are handled so that the errors in state weren't being displayed on first render. I think this is only an issue when we're writing a test like this.

We could potentially fix it by updating [useFormikErrors](https://github.com/canonical/maas-ui/blob/main/src/app/base/hooks/forms.ts#L19) to not store the initial value of the errors like this `const previousErrors = usePrevious(errors, false);`, but I'm not sure if that's intentionally storing the value for some reason.

The other solution would be to work around this by rerendering the form with the new state to simulate an update from the API. Something like this:

```typescript
it("maps name errors to hostname", async () => {
  // Render the form with default state.
  const { rerender } = renderWithBrowserRouter(
    <DiscoveryAddForm discovery={discovery} onClose={jest.fn()} />,
    { route: "/dashboard", wrapperProps: { state } }
  );
  const error = "Name is invalid";
  // Change the device state to included the errors (as if it has changed via an API response).
  state.device.errors = { name: error };
  // Rerender the form to simulate the state change.
  rerender(<DiscoveryAddForm discovery={discovery} onClose={jest.fn()} />);
  expect(
    screen.getByRole("textbox", {
      name: `${FormFieldLabels.Hostname} (optional)`,
    })
  // react-components uses aria-errormessage to link the errors to the inputs so we can use the toHaveErrorMessage helper here.
  ).toHaveErrorMessage(`Error: ${error}`);
});
```

Does that make sense? Let me know if not!

ndv99: That works perfectly! And yeah that makes a lot of sense, that's pretty clean

src/app/dashboard/views/DashboardConfigurationForm/DashboardConfigurationSubnetForm/DashboardConfigurationSubnetForm.test.tsx
```diff
@@ -141,22 +115,16 @@ describe("DashboardConfigurationSubnetForm", () => {
...
-      3: true, // false to true = changed
-      4: false, // false to false = unchanged
+    renderWithBrowserRouter(<DashboardConfigurationSubnetForm />, {
+      wrapperProps: { store },
     });
 
+    const checkboxes = screen.getAllByRole("checkbox");
+
+    await userEvent.click(checkboxes[1]);
+    await userEvent.click(checkboxes[2]);
```
petermakowski, line 125: It might be better to target each checkbox by label instead of an index so that we don't rely on a particular order of elements (and the relationship between data and rendered form is more clear). I'm happy for you to leave this as is not to increase the scope of this migration work unnecessarily.

petermakowski (approved):
Good work 👍

src/app/dashboard/views/DashboardConfigurationForm/DashboardConfigurationSubnetForm/DashboardConfigurationSubnetForm.test.tsx
ndv99: Gonna leave this as is, from a brief bit of testing trying to select them by label is too large of a rabbit hole, since the aria-labelling on them doesn't appear to work properly (label comes out as [Object object]). Might be worth making another issue to fix thsi

PR #4403, opened 2022-09-09
Migrate Enzyme to RTL - src/app/dashboard part 2

ndv99:
## Done

Migrated the following tests to RTL
- src/app/dashboard/views/DiscoveriesList/DiscoveriesList.test.tsx
- src/app/dashboard/views/DiscoveriesList/DiscoveriesFilterAccordion/DiscoveriesFilterAccordion.test.tsx

*Note: other files shown are identical to https://github.com/canonical/maas-ui/pull/4400*

## Fixes

Fixes https://github.com/canonical/app-tribe/issues/1311


src/app/dashboard/views/DiscoveriesList/DiscoveriesList.test.tsx
```diff
@@ -30,23 +92,33 @@ describe("DiscoveriesList", () => {
...
-    );
+    renderWithBrowserRouter(<DiscoveriesList />, {
+      route: route,
+      wrapperProps: { state },
+    });
 
-    expect(wrapper.text().includes("my-discovery-test")).toBe(true);
-    expect(wrapper.text().includes("another-test")).toBe(true);
+    // expect(wrapper.text().includes("my-discovery-test")).toBe(true);
+    // expect(wrapper.text().includes("another-test")).toBe(true);
```
petermakowski, line 119: Looks like you forgot to remove this.

src/app/base/components/FilterAccordion/FilterAccordion.tsx
```diff
@@ -23,6 +23,7 @@ type FilterValues = Map<FilterValue, number>;
 type FilterSections = Map<FilterKey, FilterValues>;
 
 export type Props<I, PK extends keyof I> = {
+  ariaLabel?: string;
```
petermakowski, line 26: While all DOM attributes are camelCased in React, the `aria-` are an exception. II'd suggest to stick to the convention and use a standard aria attribute with a hyphen (`aria-label`). 

Even better if you use the standard aria type from react so that it's always in sync with what React expects even if that changes in future versions. You can do that by extending AriaAttributes from React, e.g.:

```tsx
import type { AriaAttributes } from "react";
export type Props<I, PK extends keyof I> = {
 ...
} & Pick<AriaAttributes, "aria-label">;
```

src/app/base/components/FilterAccordion/FilterAccordion.tsx
```diff
@@ -104,6 +105,7 @@ const sortByFilterKey = (
 };
 
 const FilterAccordion = <I, PK extends keyof I>({
+  ariaLabel,
```
petermakowski, line 108: If not using a props spread you can pass this down like so:
```tsx
const FilterAccordion = <I, PK extends keyof I>({
  disabled,
  (...)
  "aria-label": ariaLabel
/>

  return (
    <ContextualMenu
      aria-label={ariaLabel}
```

Alternatively when using a props spread: 


```tsx
const FilterAccordion = <I, PK extends keyof I>({
  disabled,
  (...)
  {...props}
/>

  return (
    <ContextualMenu
      aria-label={props["aria-label"]}
      ```

File: src/app/dashboard/views/DiscoveriesList/DiscoveriesFilterAccordion/DiscoveriesFilterAccordion.test.tsx [hunk 4]
```diff
@@ -25,32 +25,20 @@ describe("DiscoveriesFilterAccordion", () => {
...
-      <Provider store={store}>
-        <MemoryRouter
-          initialEntries={[{ pathname: "/discoveries", key: "testKey" }]}
-        >
-          <DiscoveriesFilterAccordion searchText="" setSearchText={jest.fn()} />
-        </MemoryRouter>
-      </Provider>
+    renderWithBrowserRouter(
+      <DiscoveriesFilterAccordion searchText="" setSearchText={jest.fn()} />,
+      { route: route, wrapperProps: { state } }
```
petermakowski, line 30:
```suggestion
      { route, wrapperProps: { state } }
```

src/app/dashboard/views/DiscoveriesList/DiscoveriesFilterAccordion/DiscoveriesFilterAccordion.test.tsx
```diff
@@ -25,32 +25,20 @@ describe("DiscoveriesFilterAccordion", () => {
...
```
petermakowski, line 38:
```suggestion
      { route, wrapperProps: { state } }
```

src/app/dashboard/views/DiscoveryAddForm/DiscoveryAddForm.tsx
```diff
@@ -181,6 +181,7 @@ const DiscoveryAddForm = ({ discovery, onClose }: Props): JSX.Element => {
   return (
     <FormikForm<DiscoveryAddValues>
       allowUnchanged
+      aria-label="Add discovery form"
```
petermakowski, line 184: `form` is already expressed in the form element type and generally unnecessary.
```suggestion
      aria-label="Add discovery"
```

src/app/dashboard/views/DiscoveriesList/DiscoveriesFilterAccordion/DiscoveriesFilterAccordion.tsx
```diff
@@ -30,6 +34,7 @@ const DiscoveriesFilterAccordion = ({
 
   return (
     <FilterAccordion
+      ariaLabel="Filter discoveries"
```
petermakowski, line 37: Doesn't look like you're using the label you just created above.

src/app/dashboard/views/DiscoveriesList/DiscoveriesFilterAccordion/DiscoveriesFilterAccordion.tsx
```diff
@@ -12,6 +12,10 @@ type Props = {
   setSearchText: (searchText: string) => void;
 };
 
+export enum Labels {
+  AccordionLabel = "Filter discoveries",
```
petermakowski, line 16: I'm of the opinion it's helpful to make the name reflect the actual value (e.g. `FilterDiscoveries` instead of `AccordionLabel`) where that makes sense. 

This - arguably - makes for better readability as it doesn't require you to jump through sections of the code as much as you're reading it to understand what will be displayed.

src/app/base/components/FilterAccordion/FilterAccordion.tsx
petermakowski: More on this in react accessibility docs: https://reactjs.org/docs/accessibility.html#wai-aria

src/app/dashboard/views/DiscoveriesList/DiscoveriesList.tsx
```diff
@@ -27,8 +27,12 @@ import type { Discovery } from "app/store/discovery/types";
 import { DiscoveryMeta } from "app/store/discovery/types";
 import type { RootState } from "app/store/root/types";
 
-export enum Label {
+export enum Labels {
   Title = "Discoveries list",
```
petermakowski, line 31: Would be good to update this too so that we at least don't have inconsistencies within the same file.
```suggestion
  DiscoveriesList = "Discoveries list",
```

petermakowski (approved):

src/app/dashboard/views/DiscoveriesList/DiscoveriesList.tsx
ndv99: Done!

PR #4489, opened 2022-10-18
feat(vault): add backend integration

ndv99:
## Done

- Create selector to get vaultEnabled status (true/false)
- Add vaultConfigured to base controller type
- Create selector to split region controllers by vault configuration status (true/false)
- Add vaultEnabledState to general base type
- Write appropriate tests for the above and update others where necessary
- Add vaultEnabledState and vaultConfigured to appropriate factories

## Fixes

Fixes https://github.com/canonical/app-tribe/issues/1405

### DIFF
--- a/src/app/store/controller/selectors.ts
+++ b/src/app/store/controller/selectors.ts
@@ -375,11 +375,11 @@ const getByFabricId = createSelector(
 );
 
 /**
- * Get all region controllers and separate by their vault configuration status.
+ * Get all region/region-and-rack controllers.
  * @param state - The redux state.
- * @returns Two lists of region controllers - one where none are configured with vault, the other where all are configured with vault.
+ * @returns A list of all region/region-and-rack controllers.
  */
-const getVaultConfiguredControllers = createSelector(
+const getRegionControllers = createSelector(
   [defaultSelectors.all],
   (controllers: Controller[]) => {
     const regionControllers = controllers.filter((controller) => {
@@ -388,17 +388,30 @@ const getVaultConfiguredControllers = createSelector(
         controller.node_type === NodeType.REGION_AND_RACK_CONTROLLER
       );
     });
-    const unconfiguredControllers = regionControllers.filter((controller) => {
+
+    return regionControllers;
+  }
+);
+
+/**
+ * Get controllers separated by their vault configuration status.
+ * @param state - The redux state.
+ * @returns Two lists of region controllers - one where none are configured with vault, the other where all are configured with vault.
+ */
+const getVaultConfiguredControllers = createSelector(
+  [getRegionControllers],
+  (controllers: Controller[]) => {
+    const unconfiguredControllers = controllers.filter((controller) => {
       return (
         controller.vault_configured === false ||
         controller.vault_configured === undefined
       );
     });
-    const configuredControllers = regionControllers.filter((controller) => {
+    const configuredControllers = controllers.filter((controller) => {
       return controller.vault_configured === true;
     });
 
-    return [unconfiguredControllers, configuredControllers];
+    return { unconfiguredControllers, configuredControllers };
   }
 );
 
@@ -412,6 +425,7 @@ const selectors = {
   getStatuses,
   getStatusForController,
   getVaultConfiguredControllers,
+  getRegionControllers,
   imageSyncStatuses,
   imageSyncStatusesForController,
   processing,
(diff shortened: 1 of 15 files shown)


src/app/store/controller/selectors.ts
```diff
@@ -373,6 +374,34 @@ const getByFabricId = createSelector(
...
 
+/**
+ * Get all region controllers and separate by their vault configuration status.
+ * @param state - The redux state.
+ * @returns Two lists of region controllers - one where none are configured with vault, the other where all are configured with vault.
+ */
+const getVaultConfiguredControllers = createSelector(
+  [defaultSelectors.all],
+  (controllers: Controller[]) => {
+    const regionControllers = controllers.filter((controller) => {
```
petermakowski, line 385: Ideally `regionControllers` should be its own selector which value would be cached.

src/app/store/controller/selectors.ts
```diff
@@ -373,6 +374,34 @@ const getByFabricId = createSelector(
...
+        controller.node_type === NodeType.REGION_AND_RACK_CONTROLLER
+      );
+    });
+    const unconfiguredControllers = regionControllers.filter((controller) => {
+      return (
+        controller.vault_configured === false ||
+        controller.vault_configured === undefined
+      );
+    });
+    const configuredControllers = regionControllers.filter((controller) => {
```
petermakowski, line 397: If still think it's better to use a single selector, might as well do it a single loop, e.g. using reduce (everything that's _not_ a configured controller will be an unconfigured controller)?

src/app/store/controller/selectors.ts
```diff
@@ -373,6 +374,34 @@ const getByFabricId = createSelector(
...
+      return controller.vault_configured === true;
+    });
+
+    return [unconfiguredControllers, configuredControllers];
```
petermakowski, line 401: Returning an object instead (e.g. `{ unconfigured, configured }`) might help avoid mistakes (and is more consistent with other selectors).

src/app/store/controller/selectors.ts
```diff
@@ -373,6 +374,34 @@ const getByFabricId = createSelector(
...
+        controller.node_type === NodeType.REGION_CONTROLLER ||
+        controller.node_type === NodeType.REGION_AND_RACK_CONTROLLER
+      );
+    });
+    const unconfiguredControllers = regionControllers.filter((controller) => {
+      return (
+        controller.vault_configured === false ||
+        controller.vault_configured === undefined
+      );
+    });
```
ndv99, line 396: In the case that a controller returns `undefined`, this should be counted as `false`.

src/app/store/general/types/base.ts
```diff
@@ -234,6 +234,13 @@ export type TLSCertificateState = {
   loading: boolean;
 };
 
+export type VaultEnabledState = {
+  errors: APIError;
+  data: boolean;
```
ndv99, line 239: Originally had this as a VaultEnabled type that I created, but this didn't return values properly and generally overcomplicated things. Although I'm happy to try and make that work if it's cleaner

src/app/store/controller/selectors.ts
```diff
@@ -373,6 +374,34 @@ const getByFabricId = createSelector(
       }, [])
 );
 
+/**
+ * Get all region controllers and separate by their vault configuration status.
```
petermakowski, line 378: I'd personally break this down into 2 selectors, would make it easier to understand and make the comment mostly redundant.
e.g. `vaultConfiguredControllers` `vaultUnconfiguredControllers`

otherwise maybe something like `getByValueConfigurationStatus` returning `{ configured, unconfigured }`.

petermakowski (commented):
Looks good, just a few ideas how it could be improved. I really think the selector should be broken down (I'll leave it up to you to decide how exactly).

src/app/store/controller/selectors.ts
ndv99: Just to save time on implementation I'm gonna go with wrapping the two in an object, since these two return values are always needed together in their implementations

src/app/store/controller/selectors.ts
ndv99: I'll definitely do that!

src/app/store/controller/selectors.ts
ndv99: That's fine, implementing that

src/app/store/controller/selectors.ts
ndv99: Same as the other one really - these two return values are always needed together in their implementations, and changing to an object instead of a list would require a lot less refactoring than separating it into separate selectors for configured and unconfigured

petermakowski (approved):

PR #4504, opened 2022-10-19
feat(vault): Display warning icon next to controller nav link if not all region controllers are configured with Vault

ndv99:
## Done

- Add warning icon next to controller nav link if not all region controllers are configured with Vault
- Add test to check if icon is displayed or not

### MAAS deployment

- [Bolla](/HACKING.md#bolla)

### Running the branch

You can run this branch by:

### QA steps

- Log in
- Ensure no warning icon is displayed next to controllers in nav bar 

## Fixes

Fixes https://github.com/canonical/app-tribe/issues/1415

### Not all region controllers configured with Vault
image


src/app/base/components/Header/Header.tsx
```diff
@@ -155,7 +165,20 @@ const generateItems = (
...
     key: link.url,
-    label: link.label,
+    label:
+      link.label === "Controllers" && vaultIncomplete ? ( // check if vault is set up on all controllers
+        <>
+          <Icon
+            className="p-navigation--item-icon"
+            data-testid="warning-icon"
+            name="security-warning-grey"
+          />
```
petermakowski, line 175: This isn't purely decorative and conveys important information, for that reason it shouldn't be ignored by assistive technology. You can fix this by adding `aria-label="warning"`.

src/app/base/components/Header/Header.test.tsx
```diff
@@ -218,6 +223,33 @@ it("highlights sub-urls", () => {
...
 });
 
+it("displays a warning icon next to controllers if vault is not fully configured", () => {
+  state.controller.items = [
+    controllerFactory({ vault_configured: true }),
+    controllerFactory({ vault_configured: false }),
+  ];
+  renderWithBrowserRouter(<Header />, { route: "/", wrapperProps: { state } });
+
+  const controllerLink = screen.getByRole("link", { name: "Controllers" });
```
petermakowski, line 233: Could you update this to take the icon's aria-label into account.

src/app/base/components/Header/Header.tsx
ndv99: Can do yeah. Only reason I didn't was because of the test mentioned below - means I'm looking for a link with name "warning Controllers", but I'll implement it and update the test to account.

petermakowski: There's another technique which you could try and might be more appropriate here - keeping the icon hidden and use `aria-description`/`aria-describedby` on the link instead. Then you couldn test for the description on the link element using `expect().toHaveAccessibleDescription()`. That way the link itself would remain as `Controllers`.

petermakowski (approved):

PR #4517, opened 2022-10-21
fix(vault): Change icon colours to accessible grey

ndv99:
## Done

- Change Vault status icon colours to accessible grey `#666`
Requested by @lyubomir-popov 

### Before
image
image
image
image

### After
image
image
image
image

### DIFF
--- a/src/scss/_patterns_icons.scss
+++ b/src/scss/_patterns_icons.scss
@@ -77,12 +77,12 @@
 
   .p-icon--security-tick {
     @extend %icon;
-    @include maas-icon-security-tick($color-positive)
+    @include maas-icon-security-tick($color-mid-dark)
   }
 
   .p-icon--security-warning {
     @extend %icon;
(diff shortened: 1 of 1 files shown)


petermakowski (approved):

lyubomir-popov:
@ndv99 but there's still far too many colours here - there was a directive to only highlight in colour things that are errors/failures. I think we can should replace green with gray too

ndv99:
> @ndv99 but there's still far too many colours here - there was a directive to only highlight in colour things that are errors/failures. I think we can should replace green with gray too

What do you mean? All of the vault status icons are only showing in grey now, even the ones that are previously green (see "After" in description)

lyubomir-popov:
Yes those are fine but the rest is like a Xmas tree - 
image
image

lyubomir-popov:
hmm on second thought, we should probably provide those options in vanilla rather than doing it here, please ignore the above  two comments. I'll raise it in Vanilla.

PR #4581, opened 2022-11-23
feat(vault): render error message instead of content if vault is sealed

ndv99:
## Done

- Render an error message instead of full page content if an error message for vault is received.

### MAAS deployment

### Running the branch

You can run this branch by:

### QA steps

- Follow the setup instructions shown [here](https://docs.google.com/document/d/1ssfthB3g1kzELDaKgKdr4Mzt5n-eLQx11mnotvNbKJM/edit?disco=AAAAktcPosI)
- Migrate all secrets to Vault (instructions at /settings/configuration/security)
- Seal Vault: `vault operator seal`
- Refresh the page
- Ensure that an error message is displayed indicating that there was a Vault request error
- Unseal vault: `vault operator unseal` 
- Refresh the page
- Ensure that MAAS can be accessed again

## Fixes

Fixes https://github.com/canonical/app-tribe/issues/1411

### DIFF
--- a/src/app/App.test.tsx
+++ b/src/app/App.test.tsx
@@ -69,7 +69,7 @@ describe("App", () => {
     );
   });
 
-  it("displays an error if vault is unreachable (sealed)", () => {
+  it("displays an error if vault is sealed", () => {
     state.config.errors = "Vault request failed";
     state.status.authenticated = true;
     state.status.error = null;
@@ -79,6 +79,20 @@ describe("App", () => {
     expect(
       screen.getByText(
         /The server connection failed with the error "Vault request failed"/
+      )
+    ).toBeInTheDocument();
+  });
+
+  it("displays an error if vault is unreachable", () => {
+    state.config.errors = "Vault connection failed";
+    state.status.authenticated = true;
+    state.status.error = null;
+    state.status.connected = true;
+    renderWithBrowserRouter(<App />, { route: "/settings", state });
+    expect(screen.getByText("Failed to connect")).toBeInTheDocument();
+    expect(
+      screen.getByText(
+        /The server connection failed with the error "Vault connection failed"/
       )
     ).toBeInTheDocument();
   });
--- a/src/app/App.test.tsx
+++ b/src/app/App.test.tsx
@@ -1,4 +1,5 @@
 import * as reactComponentHooks from "@canonical/react-components/dist/hooks";
+import { screen } from "@testing-library/react";
 import { mount } from "enzyme";
 import { Provider } from "react-redux";
 import { MemoryRouter } from "react-router-dom";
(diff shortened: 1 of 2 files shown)


src/app/App.test.tsx
```diff
@@ -67,6 +69,17 @@ describe("App", () => {
     );
   });
 
+  it("displays an error if vault is unreachable (sealed)", () => {
+    state.config.errors = "Error!!!!";
```
petermakowski, line 73: Could you use an actual error message that we can expect here instead?

src/app/App.test.tsx
```diff
@@ -67,6 +69,17 @@ describe("App", () => {
...
+    state.status.authenticated = true;
+    const store = mockStore(state);
+    renderWithBrowserRouter(<App />, { route: "/settings", store });
+    expect(screen.getByText("Failed to connect")).toBeInTheDocument();
+    expect(
+      screen.getByText(/The server connection failed./)
+    ).toBeInTheDocument();
```
petermakowski, line 80: Don't we want to display the vault error message in this case as per design?

src/app/App.tsx
```diff
@@ -83,6 +85,15 @@ export const App = (): JSX.Element => {
         </Notification>
       </Section>
     );
+  } else if (configErrors) {
```
petermakowski, line 88: This is dangerous in a sense that if any generic configuration error comes through a user will be unable to use MAAS UI. We need to narrow this down to Vault errors or display a simple notification instead of replacing entire content of the app.

src/app/App.test.tsx
```diff
@@ -67,6 +69,17 @@ describe("App", () => {
     );
   });
 
+  it("displays an error if vault is unreachable (sealed)", () => {
+    state.config.errors = "Error!!!!";
+    state.status.authenticated = true;
+    const store = mockStore(state);
+    renderWithBrowserRouter(<App />, { route: "/settings", store });
```
petermakowski, line 76: You can pass state directly to `renderWithBrowserRouter` (meaning you can remove line `75, 1 line of code less).

petermakowski (commented):
I think in principle this can work, but we need to look more closely into a few things. More below.

petermakowski:
QA Note: Warning notifications are still being rendered

image

src/app/App.tsx
```diff
@@ -83,6 +85,17 @@ export const App = (): JSX.Element => {
         </Notification>
       </Section>
     );
+  } else if (
+    configErrors === "Vault request failed" ||
+    configErrors === "Vault connection failed"
```
petermakowski, line 90: It might be better to move these values to a const/enum and add as potential values to the configErrors return type in the redux store instead of using a generic APIError so that the data flow is clearer. 

 ```tsx
    const VAULT_ERRORS = {
      REQUEST_FAILED: "Vault request failed",
      CONNECTION_FAILED: "Vault connection failed"
    }
```

petermakowski (approved):
Feel free to address the comment now or in a separate PR.

petermakowski:
We should try to fix the notifications issue mentioned above if it doesn't add too much work. Please look into this @ndv99

ndv99:
> We should try to fix the notifications issue mentioned above if it doesn't add too much work. Please look into this @ ndv99

Fixed the error value storage error and put them in an enum, I'll work on the warning notification in another PR

PR #4594, opened 2022-11-28
test: Migrate remaining tests in controllers/components to RTL

ndv99:
## Done

- Migrated all tests in src/app/controllers/components to RTL

### DIFF
--- a/src/app/controllers/components/ControllerHeaderForms/ControllerHeaderForms.test.tsx
+++ b/src/app/controllers/components/ControllerHeaderForms/ControllerHeaderForms.test.tsx
@@ -1,7 +1,4 @@
-import { mount } from "enzyme";
-import { Provider } from "react-redux";
-import { MemoryRouter } from "react-router-dom";
-import configureStore from "redux-mock-store";
+import { screen } from "@testing-library/react";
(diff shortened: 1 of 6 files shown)


src/app/controllers/components/ControllerHeaderForms/ControllerHeaderForms.test.tsx
```diff
@@ -10,41 +7,40 @@ import {
...
+        controllers={[controllerFactory()]}
+        headerContent={{ view: ControllerHeaderViews.SET_ZONE_CONTROLLER }}
+        setHeaderContent={jest.fn()}
+      />,
+      { state }
     );
-    expect(wrapper.find("ControllerActionFormWrapper").exists()).toBe(true);
+
+    expect(
+      screen.getByText(/Cannot set zone of 1 controller/)
```
petermakowski, line 25: It might be better to either update the state in this test so that the form actually gets rendered (instead of the error message). Alternatively change the test description so that it reflects what's being tested here (this test case checks that a warning is displayed, the form won't show up here).

petermakowski (approved):

PR #4737, opened 2023-02-06
fix: display original lastUpdate and lastDeployed

petermakowski:
## Done

- display original UTC lastUpdate and lastDeployed time strings

This came after clarifying in the spec that an unmodified UTC time string is preferred.

### MAAS deployment

- [Bolla](/HACKING.md#bolla)

### Running the branch

You can run this branch by:

### QA steps

- Go to MAAS/r/images
- Add a new image (e.g. CentOS 7)
- Update selection
- Wait for the image to sync and note the time displayed
- Change your system's time zone to GMT+1
- Make sure the last sync time is still displaying the same UTC time

## Launchpad issue

## Backports

## Screenshots


ndv99 (approved):
Looks good, although would it make sense to add "UTC" after the timestamp for clarity?

petermakowski:
> Looks good, although would it make sense to add "UTC" after the timestamp for clarity?

It might, but apparently operators are used to time being displayed in UTC by default already. I think this would be a bigger change. We might need to update all timestamps across the app not to cause confusion if we decided to go with this.

PR #4767, opened 2023-02-16
feat(nav): Only display "Virsh" if user has Virsh KVM hosts MAASENG-1334

ndv99:
## Done

- Hide "Virsh" in nav if user has no virsh kvms

### MAAS deployment

- [Bolla](/HACKING.md#bolla)

### Running the branch

You can run this branch by:

### QA steps

- Ensure you have mock data with Virsh KVM hosts
- Go to the UI
- Ensure the Virsh tab is shown in the nav and you can click it
- If you have multipass or LXD, set up a new instance with a clean install of the maas snap (no mock data)
- Go to the UI (and skip through the intro)
- Ensure the Virsh tab is ***not*** shown in the nav

## Fixes

Fixes [MAASENG-1334](https://warthogs.atlassian.net/browse/MAASENG-1334)

[MAASENG-1334]: https://warthogs.atlassian.net/browse/MAASENG-1334?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/app/base/components/AppSideNavigation/AppSideNavigation.test.tsx
+++ b/src/app/base/components/AppSideNavigation/AppSideNavigation.test.tsx
@@ -16,6 +16,8 @@ import {
   configState as configStateFactory,
   controller as controllerFactory,
   controllerState as controllerStateFactory,
+  pod as podFactory,
+  podState as podStateFactory,
   rootState as rootStateFactory,
   user as userFactory,
   userState as userStateFactory,
@@ -56,6 +58,15 @@ describe("GlobalSideNav", () => {
       }),
       controller: controllerStateFactory({
         items: [controllerFactory()],
+        loaded: true,
+      }),
+      pod: podStateFactory({
+        loaded: true,
+        items: [
+          podFactory({
+            type: "virsh",
+          }),
+        ],
       }),
       user: userStateFactory({
         auth: authStateFactory({
@@ -339,5 +350,22 @@ describe("GlobalSideNav", () => {
     await waitFor(() =>
       expect(history.location.pathname).toBe(urls.intro.images)
     );
+  });
+
+  it("hides the 'Virsh' link if the user does not have any Virsh KVM hosts", () => {
+    const { rerender } = renderWithBrowserRouter(<AppSideNavigation />, {
+      route: "/machines",
+      state,
+    });
+
+    expect(screen.getByRole("link", { name: "Virsh" })).toBeInTheDocument();
+
+    state.pod.items = [];
+
+    rerender(<AppSideNavigation />);
+
+    expect(
+      screen.queryByRole("link", { name: "Virsh" })
+    ).not.toBeInTheDocument();
   });
 });
--- a/src/app/base/components/AppSideNavigation/AppSideNavigation.tsx
+++ b/src/app/base/components/AppSideNavigation/AppSideNavigation.tsx
@@ -23,6 +23,7 @@ import authSelectors from "app/store/auth/selectors";
 import configSelectors from "app/store/config/selectors";
 import { actions as controllerActions } from "app/store/controller";
 import controllerSelectors from "app/store/controller/selectors";
+import { actions as podActions } from "app/store/pod";
 import podSelectors from "app/store/pod/selectors";
 import type { RootState } from "app/store/root/types";
 import { actions as statusActions } from "app/store/status";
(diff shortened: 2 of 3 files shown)


src/app/base/components/AppSideNavigation/AppSideNavigation.tsx
```diff
@@ -80,6 +81,9 @@ const AppSideNavigation = (): JSX.Element => {
     dispatch(controllerActions.fetch());
   }, [dispatch]);
 
+  const virshKvms = useSelector(podSelectors.virsh);
+  const hasVirsh = virshKvms.length > 0;
```
petermakowski, line 85: In order to be able to check thi (and display the virsh nav item when available), the pods need to be loaded. We need to make sure we fetch them regardless of which page we're on (atm it only happens on some pages).

Also it might be a good idea to use optimistic updates (meaning display first, hide later once data is loaded).

ndv99: Ahhh righto, would that just be in a useEffect? Something like:

```ts
useEffect(() => {
    dispatch(podActions.fetch());
  }, [dispatch]);
```

src/app/base/components/AppSideNavigation/AppSideNavigation.tsx
```diff
@@ -138,6 +147,7 @@ const AppSideNavigation = (): JSX.Element => {
                 <AppSideNavItems
                   authUser={authUser}
                   groups={navGroups}
+                  hasVirsh={hasVirsh}
```
petermakowski, line 150: I think we should avoid adding a prop to the entire group of navigation items where it concerns only a single item.

How about you filter the navigation items array that's passed down to this component instead (making sure it's memoized and not filtered on every render)?

src/app/base/components/AppSideNavigation/AppSideNavigation.test.tsx
```diff
@@ -340,4 +351,21 @@ describe("GlobalSideNav", () => {
       expect(history.location.pathname).toBe(urls.intro.images)
     );
   });
+
+  it("hides the 'Virsh' link if the user does not have any Virsh KVM hosts", () => {
```
petermakowski, line 355: Nice and clean test 👍

src/app/base/components/AppSideNavigation/AppSideNavigation.tsx
```diff
@@ -94,6 +104,24 @@ const AppSideNavigation = (): JSX.Element => {
   });
   const themeColor = theme ? theme : maasTheme ? maasTheme : "default";
 
+  const filteredGroups = useMemo(() => {
+    if (hasVirsh === false && kvmsLoaded) {
```
petermakowski, line 108:
```suggestion
    if (hideVirsh) {
```

src/app/base/components/AppSideNavigation/AppSideNavigation.tsx
```diff
@@ -80,6 +82,14 @@ const AppSideNavigation = (): JSX.Element => {
     dispatch(controllerActions.fetch());
   }, [dispatch]);
 
+  useEffect(() => {
+    dispatch(podActions.fetch());
+  }, [dispatch]);
+
+  const virshKvms = useSelector(podSelectors.virsh);
+  const kvmsLoaded = useSelector(podSelectors.loaded);
+  const hasVirsh = virshKvms.length > 0;
```
petermakowski, line 91:
```suggestion
  const hideVirsh = kvmsLoaded && virshKvms.length < 1;
```

src/app/base/components/AppSideNavigation/AppSideNavigation.tsx
```diff
@@ -94,6 +104,24 @@ const AppSideNavigation = (): JSX.Element => {
...
+        (navLink) => navLink.label === "Virsh"
+      );
+
+      if (virshItemIndex > -1) {
+        navGroups[kvmGroupIndex].navLinks.splice(virshItemIndex, 1);
+      }
+    }
+
+    return navGroups;
+  }, [hasVirsh]);
```
petermakowski, line 123:
```suggestion
  }, [hideVirsh]);
```

petermakowski: Ideally we should use something like an array of hidden links instead to make it easier to extend in the future, but I'm not gonna push for it. It's ok as is.

petermakowski (approved):
Looks good. 👍  The only thing is that now we make an unnecessary request which results in a flash of a loading indicator when you go to one of KVM pages.

This is because we don't check if the pods are already loaded and perform another request regardless. But that's outside of scope of this PR.

PR #4824, opened 2023-03-15
test: Fix docs urls to ensure no redirects

ndv99:
## Done

- Updated docs URLs to prevent redirects and fix failing cypress test


petermakowski (approved):
LGTM!

PR #4837, opened 2023-03-21
feat(machines): Add button set on machine listing page MAASENG-1425

ndv99:
## Done

- Replace node action menu with node action button group on machine list
- Move machine list controls inside of header
- Add button to clear selection

## Todo

- [x] Fix cypress tests
- [x] Fix unit/integration tests
- [x] Add unit tests for button group (dis)appear functionality
- [x] Add unit test for clear selection button
- [x] Get advice on mobile styling

### MAAS deployment

- [Bolla](/HACKING.md#bolla)

### Running the branch

You can run this branch by:

### QA steps

- Go to machine list
  - [x] Ensure machine list controls render correctly in header
- Select a machine / machines
  - [x] Ensure the filters, searchbox, and grouping select are replaced by the action buttons
- Dispatch some actions with selections
  - [x] Ensure the correct call is made for the corresponding actions and machines
- Click the "Clear selection" button
  - [x] Ensure all machines are deselected
  - [x] Ensure the action button group is replaced by the machine list controls

## Fixes

Fixes https://warthogs.atlassian.net/browse/MAASENG-1425

### Before

image

### After

**Without selection**
image

**With selection**
image


ndv99:
@petermakowski @amylily1011 I'm struggling for ideas on how to style this on mobile - it currently looks pretty horrible on small breakpoints and I'm not 100% sure what the best way is to go about fixing it

image

image

amylily1011:
You can collapse it into 1 button like the second and remove `add hardware` button. Although, it is ok to set mobile as out of scope, because compared to the impact that it brings, I'm not sure it's worth the implementation time.

ndv99:
> I'm not sure it's worth the implementation time.

All I needed to hear :sunglasses:

petermakowski:
@amylily1011 correct me if I'm wrong but seems that the "Add hardware" button should have the new regular vanilla styles (to match "Filters")
image

amylily1011:
Yes, that's correct @petermakowski. But for mobile version, you can remove it all together.

src/app/machines/views/MachineList/MachineListControls/MachineListControls.tsx
```diff
@@ -27,42 +39,138 @@ const MachineListControls = ({
...
+        </>
+      ) : (
+        <>
+          <Col size={8}>
+            <NodeActionMenuGroup
+              alwaysShowLifecycle
+              excludeActions={[NodeActions.IMPORT_IMAGES]}
+              getTitle={getTitle}
+              hasSelection={hasSelection}
+              key="machine-list-action-menu"
```
petermakowski, line 116: Is there a reason why you've set the react key manually here? Keys are useful in dynamic lists as they help React identify which items have changed. In practice this usually means an array (especially in which the order of items may change). This is not the case here.

 You have multiple elements with the same key now which results in the following warning: `Encountered two children with the same key, `machine-list-action-menu``

petermakowski:
### QA Review

When the side navigation is expanded the "Add hardware" button gets wrapped which is inconsistent with the design (this is using the standard 15 inch screen size)

Pasted Graphic 8

cypress/e2e/with-users/machines/actions.spec.ts
```diff
@@ -1,26 +1,45 @@
...
-  "Mark fixed",
-  "Mark broken",
-  "Override failed testing",
-  "Lock",
-  "Unlock",
-  "Tag",
-  "Set zone",
-  "Set pool",
-  "Delete",
+type ActionGroup = {
```
petermakowski, line 3: No need to manually define the type here, you can just set the `MACHINE_ACTIONS_GROUPS` as read-only and let TS know this will never change.

cypress/e2e/with-users/machines/actions.spec.ts
```diff
@@ -1,26 +1,45 @@
...
+  label: string;
+  actions: string[];
+};
+
+const MACHINE_ACTIONS_GROUPS: ActionGroup[] = [
```
petermakowski, line 8:
```suggestion
const MACHINE_ACTIONS_GROUPS = [
```

cypress/e2e/with-users/machines/actions.spec.ts
```diff
@@ -1,26 +1,45 @@
...
+  },
+  {
+    label: "Categorise",
+    actions: ["Tag", "Set zone", "Set pool"],
+  },
+  {
+    label: "Lock",
+    actions: ["Lock", "Unlock"],
+  },
 ];
```
petermakowski, line 43:
```suggestion
] as const;
```

src/app/base/components/NodeActionMenuGroup/index.scss
```diff
@@ -11,11 +11,15 @@
     }
 
     .p-button, .p-switch {
-      margin: $spv--x-small 0;
-      padding: $spv--x-small $spv--large;
+      margin: calc(0.4rem - 1px) 0;
```
petermakowski, line 14: Do you remember why you chose this specific value? I wonder if we could use an existing variable here instead.

This would help stay in sync with Vanilla when there are changes to spacing and/or other styles in the future.

src/app/machines/views/MachineList/MachineListControls/MachineListControls.test.tsx
```diff
@@ -52,4 +69,116 @@ describe("MachineListControls", () => {
...
+    ).not.toBeInTheDocument();
+    expect(
+      screen.queryByRole("searchbox", { name: "Search" })
+    ).not.toBeInTheDocument();
+    expect(
+      screen.queryByRole("combobox", { name: "Group by" })
+    ).not.toBeInTheDocument();
+  });
+
+  it("dispatches an action to clear selected machines when the 'Clear seleciton' button is clicked", async () => {
```
petermakowski, line 156:
```suggestion
  it("dispatches an action to clear selected machines when the 'Clear selection' button is clicked", async () => {
```

src/app/machines/views/MachineList/MachineListControls/MachineListControls.tsx
ndv99: Huh, I thought I'd changed this one here. I was basically following the implementation of the normal Node Action Menu, which also has a key there since it (was) rendered out in a list by the section header. 

Thought I'd changed this to be "machine-list-action-menu-group", but since we're no longer mapping them in a list and just displaying them as children directly, I'd be happy to just remove the keys here

src/app/machines/views/MachineList/MachineListControls/MachineListControls.tsx
```diff
@@ -27,42 +39,138 @@ const MachineListControls = ({
...
-        />
-      </Col>
-      <Col size={3}>
-        <div className="u-flex--align-baseline">
-          <div className="u-flex--grow">
-            <GroupSelect
-              grouping={grouping}
-              setGrouping={setGrouping}
-              setHiddenGroups={setHiddenGroups}
+      {!hasSelection ? (
```
petermakowski, line 79: This means that filter groups are fetched again every time you clear the selection.

Kapture 2023-03-29 at 10 56 26

This can be mitigated by adding a check for whether filter groups have been loaded before dispatching the `machineActions.filterGroups` action in MachinesFilterAccordion component.

I added a separate task for this: https://warthogs.atlassian.net/browse/MAASENG-1522

src/app/machines/views/MachineList/MachineListControls/MachineListControls.tsx
petermakowski: Yes please!

src/app/base/components/NodeActionMenuGroup/index.scss
ndv99: my reaction to that
image

I'll get that committed, cheers!

ndv99: Actually - I've just committed that, and the action button group looks too thin now:
image

Will investigate

ndv99: image

Looks like the padding comes out as `0` for some reason...

petermakowski (approved):

PR #4890, opened 2023-05-09
test: Migrate four base components to RTL

ndv99:
## Done

- src/app/base/components/node/ActionConfirm
- src/app/base/components/node/DeleteForm
- src/app/base/components/node/DiskBootStatus
- src/app/base/components/node/FieldlessForm

### DIFF
--- a/src/app/base/components/node/DeleteForm/DeleteForm.test.tsx
+++ b/src/app/base/components/node/DeleteForm/DeleteForm.test.tsx
@@ -3,13 +3,11 @@ import configureStore from "redux-mock-store";
 import DeleteForm from "./DeleteForm";
 
 import type { RootState } from "app/store/root/types";
-import { machine as machineFactory } from "testing/factories";
 import {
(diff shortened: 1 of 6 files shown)


src/app/base/components/node/DeleteForm/DeleteForm.test.tsx
```diff
@@ -1,53 +1,48 @@
...
 
-const mockStore = configureStore();
+const mockStore = configureStore<RootState>();
 
 describe("DeleteForm", () => {
   let state: RootState;
 
   beforeEach(() => {
-    state = rootStateFactory();
+    state = getTestState();
```
petermakowski, line 20: Sorry, having thought about this more I think it makes more sense to make this change it in bulk. It'll be easier as it can be automated. I'll update the epic description and add a separate task for it. 
```suggestion
    state = rootStateFactory();
```

Jay-Topher (approved):
Looks good

PR #4896, opened 2023-05-10
test: Migrate NetworkTableConfirmation.test to RTL MAASENG-1618

ndv99:
## Done

- Migrated src/app/machines/views/MachineDetails/MachineNetwork/NetworkTable/NetworkTableConfirmation/NetworkTableConfirmation.test.tsx to RTL

## Fixes

Fixes [MAASENG-1618](https://warthogs.atlassian.net/browse/MAASENG-1618)

[MAASENG-1618]: https://warthogs.atlassian.net/browse/MAASENG-1618?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/app/machines/views/MachineDetails/MachineNetwork/NetworkTable/NetworkTableConfirmation/NetworkTableConfirmation.test.tsx
+++ b/src/app/machines/views/MachineDetails/MachineNetwork/NetworkTable/NetworkTableConfirmation/NetworkTableConfirmation.test.tsx
@@ -1,7 +1,3 @@
-import { mount } from "enzyme";
-import { Provider } from "react-redux";
-import { MemoryRouter } from "react-router-dom";
-import { CompatRouter } from "react-router-dom-v5-compat";
 import configureStore from "redux-mock-store";
 
 import NetworkTableConfirmation from "./NetworkTableConfirmation";
@@ -18,8 +14,9 @@ import {
   networkLink as networkLinkFactory,
   rootState as rootStateFactory,
 } from "testing/factories";
+import { renderWithBrowserRouter, screen, userEvent } from "testing/utils";
 
-const mockStore = configureStore();
+const mockStore = configureStore<RootState>();
 
 describe("NetworkTableConfirmation", () => {
   let nic: NetworkInterface;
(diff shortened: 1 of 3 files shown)


src/app/machines/views/MachineDetails/MachineNetwork/NetworkTable/NetworkTableConfirmation/NetworkTableConfirmation.test.tsx
```diff
@@ -225,98 +206,87 @@ describe("NetworkTableConfirmation", () => {
...
+            content: ExpandedState.DISCONNECTED_WARNING,
+          }}
+          nic={nic}
+          setExpanded={jest.fn()}
+          systemId="abc123"
+        />,
+        { store }
+      );
+
+      expect(
```
petermakowski, line 264: This is slightly confusing, you're first getting the element by using a partial string match, then verifying the same element again using a longer regex.

You should be able to use just `getByText` along with `.toBeInTheDocument` as it throws an error when not found. In case you had troubles matching the entire string due to white spaces you can just test for the disconnected warning: `"This interface is disconnected, it cannot be configured unless a cable is connected"`

petermakowski (approved):

src/app/machines/views/MachineDetails/MachineNetwork/NetworkTable/NetworkTableConfirmation/NetworkTableConfirmation.test.tsx
ndv99: I did it this way since looking for it by text wouldn't work, since the message is broken up by multiple elements:

image

I tried using the string you suggested there, but it still results in this same error

petermakowski: You should be able to workaround this using a custom text matcher:

```tsx
const getByTextContent = (text: string | RegExp) => {
  return screen.getByText((_, element) => {
    const hasText = (element: Element | null) => {
      if (element) {
        if (text instanceof RegExp && element.textContent) {
          return text.test(element.textContent);
        } else {
          return element.textContent === text;
        }
      } else {
        return false;
      }
    };
    const elementHasText = hasText(element);
    const childrenDontHaveText = Array.from(
      element?.children || []
    ).every((child) => !hasText(child));
    return elementHasText && childrenDontHaveText;
  });
};
```

@Jay-Topher is already using this in another project.

PR #5050, opened 2023-07-12
refactor: MachineListTable

petermakowski:
## Done

- refactor: MachineListTable
  - extract tableModels to own file

### QA steps

- Go to machine list
- Verify the machine list loaded and pagination, grouping and row selection works as before.


src/app/machines/views/MachineList/MachineListTable/types.ts
```diff
@@ -0,0 +1,63 @@
...
+import type { SortDirection } from "app/base/types";
+import type { MachineColumns } from "app/machines/constants";
+import type { GetMachineMenuToggleHandler } from "app/machines/types";
+import type {
+  Machine,
+  MachineStateListGroup,
+  FetchGroupKey,
+} from "app/store/machine/types";
+
+export type Props = {
```
Jay-Topher, line 15: I think naming the type `Props` was okay when it was a local type, it looks too generic now that it's being shared between files and can be mistaken for another prop type. We could call it `MachineListTableProps`

ndv99: Agreed, marking this as `Code +1 (with changes)`

ndv99, line 15:
```suggestion
export type MachineListTableProps = {
```

ndv99 (changes requested):

src/app/machines/views/MachineList/MachineListTable/types.ts
petermakowski: Good point, will update!

ndv99 (approved):

PR #5069, opened 2023-07-20
fix(navigation): prevent side nav display without auth 3.4 backport

ndv99:
## Done

- Prevent side nav from displaying if user is not connected or authenticated

## QA

QA for this was completed in https://github.com/canonical/maas-ui/pull/5067


petermakowski:
Could you please use git cherry-pick next time to make sure we can easily track commits across different versions.

petermakowski (approved):

PR #5101, opened 2023-08-17
feat: responsive machine columns MAASENG-1505

petermakowski:
## Done

- feat: responsive machine columns MAASENG-1505
- extract machine list table utils to hooks: `usePageSize`, `useGrouping`, `useResponsiveColumns`

### QA steps

- Go to machine list
- Open the columns dropdown
- Make sure you have all columns checked
- WIth the dropdown still open, start resizing the widow down and verify that columns are unchecked progressively as the window size decreases
- With the window size set to the smallest size, enable additional columns
- Without changing the window size, refresh the page
- Columns should remain the same
- Slowly start to increase the window size
- As you increase the window size the columns should reset to the default set for each breakpoint

## Fixes

Fixes:  https://warthogs.atlassian.net/browse/MAASENG-1505

### DIFF
--- a/cypress/e2e/with-users/machines/list.spec.ts
+++ b/cypress/e2e/with-users/machines/list.spec.ts
@@ -108,6 +108,8 @@ context("Machine listing", () => {
 
   it("can hide machine table columns", () => {
     const allHeadersCount = 11;
+    cy.viewport("macbook-15");
+
     cy.findAllByRole("columnheader").should("have.length", allHeadersCount);
 
     cy.findAllByRole("button", { name: "Columns" }).click();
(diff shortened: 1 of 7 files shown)


Jay-Topher:
This works as expected and the code looks good, we may need to add some tests for the hooks you created

petermakowski:
> This works as expected and the code looks good, we may need to add some tests for the hooks you created

Thanks for pointing this out.

These hooks are already covered with e2e and integration tests; they were extracted only for better readability.

I personally do not believe that unit tests for every single function should be a hard requirement - especially for components that are only used in a single place testing. In such cases thorough testing at the integration level should be sufficient.

Jay-Topher:
Oh right, I wasn't aware they were covered in e2e tests because I didn't see any diffs, that makes sense

Jay-Topher (approved):

PR #5110, opened 2023-08-21
test: set BondFormFields initialValues

petermakowski:
## Done

- test: set BondFormFields initialValues
- test: use toHaveAccessibleErrorMessage

Fixes the following error thrown when running `BondFormFields.test.tsx` test:
```
Warning: A component is changing an uncontrolled input to be controlled. This is likely caused by the value changing from undefined to a defined value, which should not happen. Decide between using a controlled or uncontrolled input element for the lifetime of the component. More info: https://reactjs.org/link/controlled-components
```

Fixes the following error across the codebase: 

```
Warning: toHaveErrorMessage has been deprecated and will be removed in future updates. Please use toHaveAccessibleErrorMessage.
```

### DIFF
--- a/src/app/base/components/FormikField/FormikField.test.tsx
+++ b/src/app/base/components/FormikField/FormikField.test.tsx
@@ -29,7 +29,9 @@ describe("FormikField", () => {
       </Formik>
     );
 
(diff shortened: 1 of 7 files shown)


Jay-Topher (approved):
Looks good

PR #5118, opened 2023-08-23
5111 fix docs link tests

Jay-Topher:
## Done

- Update docurls on app
- Handled uncaught exceptions

### QA steps

- Ensure `docs-links.spec.ts` is passing

## Fixes

Fixes: #5111

### DIFF
--- a/cypress/support/e2e.ts
+++ b/cypress/support/e2e.ts
@@ -30,3 +30,7 @@ declare global {
     }
   }
 }
+
+Cypress.on("uncaught:exception", (err, _runnable) => {
+  return false;
+});
--- a/cypress/support/e2e.ts
+++ b/cypress/support/e2e.ts
@@ -31,6 +31,13 @@ declare global {
   }
 }
 
-Cypress.on("uncaught:exception", () => {
-  return false;
+Cypress.on("uncaught:exception", (err, _runnable) => {
+  // run tests if canonicalGlobalNav, hljs, drpNs is not defined
+  if (
+    err.message.includes("canonicalGlobalNav") ||
+    err.message.includes("hljs") ||
+    err.message.includes("drpNs")
+  ) {
+    return false;
+  }
 });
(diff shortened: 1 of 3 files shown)


cypress/support/e2e.ts
```diff
@@ -30,3 +30,7 @@ declare global {
     }
   }
 }
+
+Cypress.on("uncaught:exception", (err, _runnable) => {
+  return false;
+});
```
ndv99, line 36: What is this needed for?

Jay-Topher: When navigating to the MAAS docs, there's an error that says `canonicalGlobalNav is not defined`, along with a few other scripts which are not instrumental to our test. Once this exception is thrown, the test fails automatically. This prevents the test from failing because of the absence of those scripts

petermakowski: This disables failing on uncaught exceptions globally (including tests of maas-ui itself) which is not what we want. As a workaround, this should be as narrow as possible (and target maas.io errors that prevent checks from passing).

This can be done by e.g. checking the error message and only returning false for those that are returned from maas.io when testing links.

Alternatively you can provide a stub response for network requests that we do not care about and that seem to be failing. 
https://docs.cypress.io/api/commands/intercept#Controlling-the-response

Jay-Topher: Does this filter make more sense? 
```javascript
Cypress.on("uncaught:exception", (err, _runnable) => {
  // run tests if canonicalGlobalNav, hljs, drpNs is not defined
  if (
    err.message.includes("canonicalGlobalNav") ||
    err.message.includes("hljs") ||
    err.message.includes("drpNs")
  ) {
    return false;
  }
});
```

petermakowski: It is an improvement, might be worth explaining in the comment why we're doing this.

ndv99 (approved):
I'm happy with this implementation, +1 from me

PR #5123, opened 2023-08-23
chore: upgrade react components

petermakowski:
## Done

- chore: upgrade react components
- update tests to hover a tooltip trigger before assertions
- add `expectTooltipOnHover` test util to reduce repetition
- fix react type resolution issues
- fix type errors

### QA Steps
- Go to machine list and hover over any icons that trigger a tooltip to verify it works correctly
- Navigate to a few other pages ensuring there's nothing unexpected

### DIFF
--- a/src/app/base/components/NodeActionMenu/NodeActionMenu.test.tsx
+++ b/src/app/base/components/NodeActionMenu/NodeActionMenu.test.tsx
@@ -222,27 +222,12 @@ describe("NodeActionMenu", () => {
       />
     );
 
+    await userEvent.click(screen.getByRole("button"));
     expect(
       screen.getByRole("tooltip", {
         name: "Select foobars below to perform an action.",
       })
     ).toBeInTheDocument();
-  });
-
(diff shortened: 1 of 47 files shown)


src/app/base/components/NodeActionMenu/NodeActionMenu.test.tsx
```diff
@@ -222,29 +222,14 @@ describe("NodeActionMenu", () => {
...
 
+    await userEvent.click(screen.getByRole("button"));
     expect(
       screen.getByRole("tooltip", {
         name: "Select foobars below to perform an action.",
       })
     ).toBeInTheDocument();
   });
 
-  it("can change the position of the disabled tooltip", async () => {
```
petermakowski, line 232: This test was failing (even though component works correctly in the browser). As it was testing internal implementation details of a tooltip (which are already tested elsewhere) I decided to remove it.

ndv99 (approved):

PR #5157, opened 2023-09-13
MAASENG-2148 machine details ephemeral indicator

Jay-Topher:
## Done

- Add ephemeral deploy indicator to machine details page

### QA steps

- Go to the details page of any deployed machine
- If that machine is not deployed ephemerally, negate the condition in line 77 of `src/app/base/components/node/OverviewCard/MachineStatusCard/MachineStatusCard.tsx`
- You should see a heading that says 'Deployed in memory'

## Fixes
[MAASENG-2148](https://warthogs.atlassian.net/browse/MAASENG-2148)

## Screenshots
image

[MAASENG-2148]: https://warthogs.atlassian.net/browse/MAASENG-2148?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/app/base/components/node/OverviewCard/MachineStatusCard/MachineStatusCard.tsx
+++ b/src/app/base/components/node/OverviewCard/MachineStatusCard/MachineStatusCard.tsx
@@ -16,7 +16,7 @@ import {
   NodeStatusCode,
   TestStatusStatus,
 } from "app/store/types/node";
-import { breakLines } from "app/utils";
+import { breakLines, isEphemerallyDeployed } from "app/utils";
 
 type Props = {
   machine: MachineDetails;
@@ -74,6 +74,9 @@ const MachineStatusCard = ({ machine }: Props): JSX.Element => {
             </i>
           )}
           {machine.status}
+          {isEphemerallyDeployed(machine) && (
+            <span className="p-text--small"> in memory</span>
+          )}
         </h4>
 
         {machine.show_os_info ? (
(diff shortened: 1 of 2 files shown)


src/app/base/components/node/OverviewCard/MachineStatusCard/MachineStatusCard.tsx
```diff
@@ -74,6 +74,9 @@ const MachineStatusCard = ({ machine }: Props): JSX.Element => {
             </i>
           )}
           {machine.status}
+          {isEphemerallyDeployed(machine) && (
```
ndv99, line 77: In theory, `ephemerally_deployed` could only be true if the machine also has a "deployed" status, but just to be sure I'd still add a check here just in case.

```suggestion
          {machine.status === NodeStatus.DEPLOYED && isEphemerallyDeployed(machine) && (
```

Jay-Topher: I think this check is already met on the utility function like so:

```javascript
export const isEphemerallyDeployed = (machine: Machine | null) => {
  return (
    machine &&
    machine.status_code === NodeStatusCode.DEPLOYED &&
    machine.ephemeral_deploy
  );
};
```
I could add this (`machine.status === NodeStatus.DEPLOYED`) also to the check to make it more thorough, what do you think?

ndv99: Oh, if it's already in the util function then that's fine IMO, please disregard!

ndv99 (approved):
LGTM

PR #5168, opened 2023-09-20
docs: add story for MachineActionMenu

petermakowski:
## Done

- docs: add story for MachineActionMenu

### QA steps

- run `yarn storybook`
- go to http://localhost:6006/?path=/story/sections-machine-machineactionmenu--example
- verify the component is displayed

## Screenshots
Google Chrome screenshot 000840@2x

### DIFF
--- a/src/app/machines/views/MachineList/MachineListControls/MachineActionMenu/MachineActionMenu.stories.tsx
+++ b/src/app/machines/views/MachineList/MachineListControls/MachineActionMenu/MachineActionMenu.stories.tsx
@@ -1,0 +1,16 @@
+import type { Meta } from "@storybook/react";
+
+import MachineActionMenu from ".";
+
+const meta: Meta<typeof MachineActionMenu> = {
+  title: "Sections/Machine/MachineActionMenu",
+  component: MachineActionMenu,
+  tags: ["autodocs"],
+};
+export default meta;
+
+export const Example = {
+  args: {
+    hasSelection: true,
+  },
+};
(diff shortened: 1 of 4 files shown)


src/app/machines/views/MachineList/MachineListControls/MachineActionMenu/MachineActionMenu.stories.tsx
```diff
@@ -0,0 +1,16 @@
...
+  title: "Sections/Machine/MachineActionMenu",
+  component: MachineActionMenu,
+  tags: ["autodocs"],
+};
+export default meta;
+
+export const Example = {
+  args: {
+    hasSelection: true,
+  },
```
ndv99, line 15: Is it possible to set `setSidePanelContent` here as just a void function? This would avoid error messages that display when clicked in the storybook

ndv99 (approved):
Just one comment below, but otherwise LGTM

src/app/machines/views/MachineList/MachineListControls/MachineActionMenu/MachineActionMenu.stories.tsx
petermakowski: Thanks for pointing this out!

I'd be cautious when introducing additional manual handlers for props, especially when it's writing additional code that has no other function other than avoiding an error on interaction that we don't really support anyway at the moment.

I'll resolve this in a followup PR by addressing state settters (set* type of props) globally.

PR #5207, opened 2023-11-10
chore: Migrate maas-ui from CRA to Vite

ndv99:
# This PR is work in progress, and will break a lot of things.

## Done
- Removed react-scripts
- Installed Vite
- Resolve absolute file paths with "@/"
- Replaced `process.env` with `import.meta.env`
- Replaced `REACT_` in env vars with `VITE_`
- Updated scripts in package to use vite
- Replaced jest with vitest

## To do
- [x] Migrate to Vitest from Jest
- [x] Fix any failing tests
- [x] ~~Update GitHub and Usabilla env var keys~~
- [x] Build the UI, ask someone from core very nicely for some help using it with MAAS 
- [x] Fix the show-ready script so that dotrun can pass 
- [x] Fix font asset resolution at build time (https://github.com/vitejs/vite/issues/15109)
- [x] Resolve merge conflicts with #5209 
- [x] Find out why those lint errors are happening
- [x] Get dotrun to use NodeJS v20 , OR create a working "build" check

## QA steps

- [ ] 

## Fixes

Fixes:


ndv99:
There is an [issue with Vite](https://github.com/vitejs/vite/issues/9520) that causes the dev server to return a 404 response if the request does not contain an "Accept" header, or if that header is set to `*/*`. We use a utility `wait-on` in some of our scripts that sets its Accept header to this. The last commit before this comment contains a workaround, but this will need to be removed once this issue is fixed.

petermakowski:
I'd encourage breaking this down. `vitest` migration could be done in a separate PR.

ndv99:
@petermakowski That was my plan initially, but to do that you need to install jest and configure it - I tried this, and it ended up being a lot more hassle than it was worth for something that we'll just be getting rid of in the next PR

ndv99:
Regarding the broken font resolution at built time, I've filed an issue on the Vite repo: https://github.com/vitejs/vite/issues/15109

ndv99:
My fix for URL decoding in Vite has landed! https://github.com/vitejs/vite/pull/15246

Once that fix makes it into a release, I'll upgrade the Vite version here

ndv99:
I've decided to leave lazy-loading for a follow-up PR once this lands. It's causing too many flakey tests at the moment for me to justify spending more time getting it to work just now

SK1Y101:
Verified latest build as working in my development container running maas 3.5.0~alpha1

ndv99:
Now that https://code.launchpad.net/~nickdv99/maas/+git/maas/+merge/457420 has landed, and the vite build has been shown to work in MAAS proper, I think this is no longer a breaking change. Will be removing the label

ndv99:
I've finally created a successful build check!

petermakowski:
> I've finally created a successful build check!

OK, I'll update and remove run-dotrun from merge requirements for now.

petermakowski:
@ndv99 Running `yarn start` results in an error:

```
[0] [nodemon] starting `node ./scripts/proxy.js`
[0] file:///Users/makowski/Documents/GitHub/maas-ui/scripts/proxy.js:1
[0] require("dotenv-flow").config();
[0] ^
[0]
[0] ReferenceError: require is not defined in ES module scope, you can use import instead
[0] This file is being treated as an ES module because it has a '.js' file extension and '/Users/makowski/Documents/GitHub/maas-ui/package.json' contains "type": "module". To treat it as a CommonJS script, rename it to use the '.cjs' file extension.
```

We should refactor both `proxy.js` and `proxy-ready.js` to use `import`.

ndv99:
> @ndv99 Running `yarn start` results in an error:
> 
> ```
> [0] [nodemon] starting `node ./scripts/proxy.js`
> [0] require("dotenv-flow").config();
> [0] ^
> [0]
> [0] ReferenceError: require is not defined in ES module scope, you can use import instead
> [0] This file is being treated as an ES module because it has a '.js' file extension and 'maas-ui/package.json' contains "type": "module". To treat it as a CommonJS script, rename it to use the '.cjs' file extension.
> ```
> 
> We should refactor both `proxy.js` and `proxy-ready.js` to use `import`.

Fixed!

petermakowski:
There's a bunch of warnings around the basename when running tests: `Warning: You are attempting to use a basename on a page whose URL path does not begin with the basename. Expected path "/" to begin with "/MAAS/r".`

We're also getting a virsh memoised selector error on this PR which does not seem to happen on main (it might just be that it only surfaced now that we changed the build pipeline, but still worth investigating).
```
src/app/base/components/AppSideNavigation/AppSideNavigation.test.tsx > GlobalSideNav > persists collapsed state
Selector virsh returned a different result when called with the same parameters. This can lead to unnecessary rerenders.
Selectors that return a new reference (such as an object or an array) should be memoized: https://redux.js.org/usage/deriving-data-selectors#optimizing-selectors-with-memoization {
```

package.json
```diff
@@ -62,10 +64,9 @@
     "react-dom": "18.2.0",
     "react-dropzone": "14.2.3",
     "react-portal": "4.2.2",
-    "react-redux": "8.0.5",
+    "react-redux": "9.0.4",
```
petermakowski, line 67: Regarding the selector error, this is most likely the culprit. Is upgrade to version 9 required for vite upgrade to work?

ndv99: Huh, that worked! Thanks mate :)

petermakowski: Great!

petermakowski (approved):
Great work @ndv99 👏 
This is ready to go to main. Let's monitor for any issues closely in the following days.

ndv99:
> Great work @ndv99 👏 This is ready to go to main. Let's monitor for any issues closely in the following days.

Thank you so much!

PR #5210, opened 2023-11-15
ci: bump node from v16 to v20 LTS

ndv99:
## Done
- Changed node version to 20.X in CI pipeline
- Added `.nvmrc` with latest node LTS for nvm users


.nvmrc
```diff
@@ -0,0 +1 @@
+v20.9.0
```
petermakowski, line 1: How about we specify the major version only to match test.yml workflow?

```suggestion
v20
```

ndv99: Let me test that locally to see if it works as intended. One sec

ndv99: Yep, that works!

petermakowski: Even better if we centralised this and read the node version from .nvmrc file in the github action: https://github.com/actions/setup-node/issues/32#issuecomment-1003854758

petermakowski: @ndv99  This is more relevant for us: https://github.com/actions/setup-node/issues/32#issuecomment-872375574

ndv99:
oh hey that worked

.github/workflows/test.yml
```diff
@@ -6,12 +6,19 @@ on:
...
+      - name: "Read .nvmrc"
+        run: "echo NVMRC=`cat .nvmrc` >> $GITHUB_ENV"
   lint:
     name: Lint
+    needs: read-node-version
     runs-on: ubuntu-22.04
     strategy:
       matrix:
-        node-version: [16.x]
+        node-version: [20.x]
```
petermakowski: This can be removed since we use a single version of node specified in the nvmrc.

ndv99: Pushed a commit resolving this :+1:

.github/workflows/test.yml
```diff
@@ -6,12 +6,16 @@ on:
   pull_request:
     types: [opened, synchronize, reopened]
 jobs:
+  read-node-version:
+    name: "Read NodeJS version"
+    runs-on: ubuntu-22.04
+    steps:
+      - name: "Read .nvmrc"
+        run: "echo NVMRC=`cat .nvmrc` >> $GITHUB_ENV"
```
petermakowski, line 14: You need to checkout the repo before reading the version in order for the file to be available for the runner.

petermakowski: https://github.com/canonical/maas-react-components/blob/main/.github/workflows/CI.yml

petermakowski (approved):
Much cleaner, nice one!

PR #5244, opened 2023-12-12
feat: use ContentSection to wrap header and content in MainContentSection WIP

ndv99:
## Blocked by

https://warthogs.atlassian.net/browse/MAASENG-2440

## Done
- Wrapped header and content of MainContentSection in ContentSection components

## QA steps

TBD

### Before
image

### After
image


petermakowski:
Could you try and update this with latest changes in `maas-react-components`?

ndv99:
> Could you try and update this with latest changes in `maas-react-components`?

Of course! Sorry, been carried away with Vite stuff today.

ndv99:
This is still rough, hence the comments and failing checks, but I just wanted to check that I'm at least on the right track here

This change is going to affect a lot of pages - at first glance, it actually all looks alright to me, aside from some padding here and there, but have a look about and just make sure that nothing is super broken or anything

src/app/base/components/MainContentSection/MainContentSection.tsx
```diff
@@ -23,11 +24,14 @@ const MainContentSection = ({
...
-      <div>
+      <ContentSection>
         {header ? (
-          <header aria-label="main content" className="row">
-            <Col size={12}>{header}</Col>
-          </header>
+          // <header aria-label="main content" className="row">
+          //   <Col size={12}>{header}</Col>
+          // </header>
+          <ContentSection.Header aria-label="main content">
```
petermakowski, line 32: Do we need to do this? Does rendering a header without content cause any issues?

src/app/base/components/MainContentSection/MainContentSection.tsx
```diff
@@ -36,11 +40,13 @@ const MainContentSection = ({
             </Col>
           )}
           <Col size={(sidebar ? TOTAL - SIDEBAR : TOTAL) as ColSize}>
-            {!isNotificationListHidden && <NotificationList />}
-            {children}
+            <ContentSection.Content>
```
petermakowski, line 43: This means that a potential footer will be nested within content. This shouldn't break anything, but we don't have such an example in the component library. Could you please add it, then this should be good to go.

src/app/base/components/MainContentSection/MainContentSection.tsx
petermakowski: Actually, this is unnecessary (and Header component does not exist). Composition of the header will depend on the page and it can either be Title or MainToolbar, which will be handled by a child component.

src/app/base/components/MainContentSection/MainContentSection.tsx
```diff
@@ -23,11 +24,14 @@ const MainContentSection = ({
   const { SIDEBAR, TOTAL } = COL_SIZES;
   return (
     <div {...props} id="main-content-section">
-      <div>
+      <ContentSection>
```
petermakowski, line 27: Once we add this, we'll also need to remove duplicated `ContentSection` wrapper from settings pages.

petermakowski: Which leaves a question - what feels more natural - handling this globally in PageContent component - or individually per page?

On one hand, handling it per page definitely gives us more flexibility but that can also lead to inconsistencies in the future.

@ndv99 How do you think?

src/app/base/components/MainContentSection/MainContentSection.tsx
petermakowski: Before All pages use such components, we should remove all instances of `.section-header` className and associated styles.

src/app/base/components/MainContentSection/MainContentSection.tsx
ndv99: I think I'm leaning more towards handling it globally. It may be less flexible, but given that pages in maas-ui follow a very similar layout to each other, I think it makes sense here and reduces repetition.

petermakowski: Same here!

src/app/base/components/MainContentSection/MainContentSection.test.tsx
```diff
@@ -22,9 +22,10 @@ it("can render without a header", () => {
   renderWithMockStore(
     <MainContentSection header={null}>content</MainContentSection>
   );
-  expect(
-    screen.queryByRole("banner", { name: "main content" })
-  ).not.toBeInTheDocument();
+
+  // ContentSection.Header doesn't have an accessible role, and also
+  // does not accept aria labels or test ids in its props.
+  expect(document.querySelector(".content-section__header")).toBeNull();
```
ndv99, line 28: Normally I wouldn't do this, but as the comment explains, it's not currently possible to query for this element using testing-library methods. 

A fix for this could be to alter `ContentSection.Header` to use a `header` element as opposed to a `div` - this would give it an accessible role of "banner", and would make it possible to query for using testing-library.

src/app/base/components/MainContentSection/MainContentSection.tsx
ndv99: So, after some messing around, removing `.section-header` styles pretty much breaks the header of every page that isn't using `MainToolbar`. For this reason, I'm gonna put this PR on hold and instead focus on knocking out the subtasks in [this ticket](https://warthogs.atlassian.net/browse/MAASENG-2440).

ndv99:
This PR is redundant now, closing.

(fix implemented by https://github.com/canonical/maas-ui/pull/5361)

PR #5250, opened 2024-01-04
refactor: remove local Stepper component

petermakowski:
## Done
refactor: remove local Stepper component
- update @canonical/maas-react-components to v1.14.0

## QA steps

- [ ] Go to lxd page
- [ ] Click Add LXD host
- [ ] Ensure the stepper is displayed with correct steps and active steps as you proceed

## Screenshots
Google Chrome screenshot 001254@2x


ndv99:
"Credentials" is the first step, but with this implementation it's being shown as completed, with the second step highlighted, even though the form is from the first step.

Expected:
image

Actual:
image

petermakowski:
> "Credentials" is the first step, but with this implementation it's being shown as completed, with the second step highlighted, even though the form is from the first step.

Good spot! The problem lies in `Object.values(AddLxdSteps)`. The object values aren't in the correct order which I missed when refactoring to make it more concise.

Jay-Topher (approved):

PR #5262, opened 2024-01-09
fix: main toolbar padding

petermakowski:
## Done
- add bottom padding to the main toolbar

## QA steps

- [ ] 

## Fixes

Fixes: 

### Before
Google Chrome screenshot 001278@2x


ndv99 (approved):
LGTM

PR #5283, opened 2024-01-15
test: fix Cypress domains list test

ndv99:
## Done
- Updated test id for finding header on domains list

## QA steps

- [x] Run cypress test for domains list and ensure it passes

## Notes

This is one PR in what will be a series of PRs to fix our failing Cypress runs.

### DIFF
--- a/cypress/e2e/with-users/domains/list.spec.ts
+++ b/cypress/e2e/with-users/domains/list.spec.ts
@@ -7,6 +7,6 @@ context("DNS", () => {
   });
 
   it("renders the correct heading", () => {
-    cy.get("[data-testid='section-header-title']").contains("DNS");
+    cy.get("[data-testid='main-toolbar-heading']").contains("DNS");
   });
 });


cypress/e2e/with-users/domains/list.spec.ts
```diff
@@ -7,6 +7,6 @@ context("DNS", () => {
   });
 
   it("renders the correct heading", () => {
-    cy.get("[data-testid='section-header-title']").contains("DNS");
+    cy.get("[data-testid='main-toolbar-heading']").contains("DNS");
```
petermakowski, line 10: Depending on how many failures are related to this, it might make more sense to rename the testid in toolbar itself instead. There are more places where we use this pattern.

ndv99: I think it's only 4 other failures in Cypress.

petermakowski (approved):

PR #5290, opened 2024-01-17
feat: move domain set default form to sidepanel

Jay-Topher:
## Done
- Moved set default domain form to sidepanel

## QA steps

- [ ] visit `/domains` and open the set default domain form
- [ ] Ensure that it opens in a sidepanel

## Fixes

Fixes: [Move domain set default form to side panel](https://warthogs.atlassian.net/browse/MAASENG-2575)

### Before
image

### After
image


petermakowski:
@Jay-Topher Also, once you change a default and attempt to change it back to the previous, the side panel is empty and no action can be taken.

src/app/base/components/ModelActionForm/ModelActionForm.tsx
```diff
@@ -7,14 +7,18 @@ import type { EmptyObject } from "@/app/base/types";
 type Props = {
   modelType: string;
   message?: string;
+  showSubtext?: boolean;
+  subtext?: string;
```
petermakowski, line 11: Why not simply output everything as a single string for simplicity? I think we can safely drop the light text styling for the sake of reducing complexity.

As a side note, next time in a similar scenario I'd recommend to simplify, `showSubtext` prop is redundant in this case. As `subtext` is already optional we can assume showSubtext will always be false if subtext is falsy (undefined) and remove it.

PR #5291, opened 2024-01-18
feat: move SSH Key delete form to sidepanel

Jay-Topher:
## Done
- moved SSH Key delete form to sidepanel

## QA steps

- [x] visit /account/prefs/ssh-keys
- [x] If you don't have an SSH key added already, add one
- [x] Click on the delete button on the table row
- [x] ensure the side panel opens up to delete the SSH key
- [x] ensure deletion works

## Fixes

Fixes: [MAASENG-2576](https://warthogs.atlassian.net/browse/MAASENG-2576)

### Before
image

### After
image

## Notes

[MAASENG-2576]: https://warthogs.atlassian.net/browse/MAASENG-2576?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/app/base/components/SSHKeyList/SSHKeyList.tsx
+++ b/src/app/base/components/SSHKeyList/SSHKeyList.tsx
@@ -1,26 +1,19 @@
 import type { ReactNode } from "react";
-import { useState } from "react";
 
 import { Button, Notification } from "@canonical/react-components";
-import { useDispatch, useSelector } from "react-redux";
-import type { Dispatch } from "redux";
+import { useSelector } from "react-redux";
+import type { NavigateFunction } from "react-router-dom-v5-compat";
+import { useNavigate } from "react-router-dom-v5-compat";
 
-import TableDeleteConfirm from "@/app/base/components/TableDeleteConfirm";
-import { useFetchActions, useAddMessage } from "@/app/base/hooks";
+import { useFetchActions } from "@/app/base/hooks";
+import urls from "@/app/preferences/urls";
 import SettingsTable from "@/app/settings/components/SettingsTable";
 import type { Props as SettingsTableProps } from "@/app/settings/components/SettingsTable/SettingsTable";
 import { actions as sshkeyActions } from "@/app/store/sshkey";
 import sshkeySelectors from "@/app/store/sshkey/selectors";
-import type {
-  KeySource,
-  SSHKey,
-  SSHKeyMeta,
-  SSHKeyState,
-} from "@/app/store/sshkey/types";
+import type { KeySource, SSHKey } from "@/app/store/sshkey/types";
 
-type Props = {
-  sidebar?: boolean;
-} & Partial<SettingsTableProps>;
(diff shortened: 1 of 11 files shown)


src/app/base/components/SSHKeyList/SSHKeyList.tsx
```diff
@@ -1,26 +1,20 @@
...
-  SSHKey,
-  SSHKeyMeta,
-  SSHKeyState,
-} from "@/app/store/sshkey/types";
+import type { KeySource, SSHKey } from "@/app/store/sshkey/types";
 
-type Props = {
-  sidebar?: boolean;
-} & Partial<SettingsTableProps>;
+type Props = {} & Partial<SettingsTableProps>;
```
petermakowski, line 17:
```suggestion
type Props = Partial<SettingsTableProps>;
```

src/app/preferences/components/Routes/Routes.tsx
```diff
@@ -18,6 +21,7 @@ import { getRelativeRoute } from "@/app/utils";
 
 const Routes = (): JSX.Element => {
   const base = urls.preferences.index;
+  const { sidePanelContent } = useSidePanel();
```
petermakowski, line 24: Delete SSHKey sidePanel should be setup in a way that's consistent with API key delete side panel. It's activated via own route `/r/account/prefs/api-keys/{id}/delete`.

```suggestion
```

petermakowski (changes requested):

src/app/preferences/components/Routes/Routes.tsx
Jay-Topher: This would be problematic because the ssh key deletion takes an array of values for multiple deleting, when the table is grouped, having it as a URL route will only account for a single ID

petermakowski: Passing multiple ids is perfectly valid as well via query parameters.

e.g. `/r/account/prefs/ssh-keys/delete?ids=1,2`

src/app/preferences/components/Routes/Routes.tsx
```diff
@@ -76,7 +80,10 @@ const Routes = (): JSX.Element => {
       />
       <Route
         element={
-          <PageContent sidePanelContent={null} sidePanelTitle={null}>
+          <PageContent
+            sidePanelContent={sidePanelContent && <DeleteSSHKey />}
+            sidePanelTitle="Delete SSH key"
+          >
```
petermakowski, line 86: This is no longer necessary.
```suggestion
          <PageContent sidePanelContent={null} sidePanelTitle={null}>
```

petermakowski (commented):
Just a small change and this should be ready to go.

petermakowski (approved):

PR #5306, opened 2024-01-25
ci(cypress): create issue only on main branches

petermakowski:
## Done
- ci(cypress): create issue only on main branches

## QA steps

- [ ] 

## Fixes

Fixes:

### DIFF
--- a/.github/workflows/cypress.yml
+++ b/.github/workflows/cypress.yml
@@ -61,7 +61,7 @@ jobs:
           path: cypress/screenshots
       - name: Create issue on failure
         # Create an issue if the job fails on push to main or 3.* branches
-        if: failure() && github.event_name == 'push' && contains(fromJson('["main", "3.*", "*cypress*"]'), env.BRANCH_NAME)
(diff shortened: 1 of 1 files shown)


.github/workflows/cypress.yml
```diff
@@ -60,7 +60,8 @@ jobs:
           name: cypress-screenshots
           path: cypress/screenshots
       - name: Create issue on failure
-        if: failure()
+        # Create an issue if the job fails on push to main or 3.* branches
+        if: failure() && github.event_name == 'push' && contains(fromJson('["main", "3.*", "*cypress*"]'), env.BRANCH_NAME)
```
ndv99, line 64: Do we need it to create an issue on `*cypress*` branches? I thought those branches would be for fixing cypress tests - if it fails, it would just show as a status in the PR, no?

petermakowski: Correct. I thought it's good to keep it here for ensuring the workflow still functions correctly. Having said that, github doesn't require you to merge changes in order to test out updates to workflows, so... Yup, let's remove.

ndv99 (approved):

PR #5308, opened 2024-01-30
chore: use formatBytes from maas react components

petermakowski:
## Done
- chore: use formatBytes from maas react components
  - update maas react components

## QA steps

- [ ] Go to machine details page and verify that values on the storage page reflect values for the same machine on http://maas-ui-demo.internal:5240/MAAS/r/machines
- [ ] Repeat this for LXD


ndv99 (approved):
LGTM

PR #5318, opened 2024-02-22
fix: update machine deploy form UI behavior

Jay-Topher:
## Done
- Made sidepanel header sticky
- Removed unwanted `hr` and processing text from deploy form

## QA steps

- [ ] Goto `/machines`
- [ ] Select a machine > Actions > Deploy
- [ ] Ensure the form displays correctly and header is sticky when you scroll
- [ ] Click on 'Deploy' and ensure that the button does not jump

## Fixes

Fixes: [MAASENG-2756](https://warthogs.atlassian.net/browse/MAASENG-2756)

### Before
[Screencast from 02-22-2024 10:14:12 AM.webm](https://github.com/canonical/maas-ui/assets/47540149/c18c74ad-b06d-4ab3-ac1f-0e815fdc1c5d)

### After
[Screencast from 02-22-2024 10:16:54 AM.webm](https://github.com/canonical/maas-ui/assets/47540149/478a4e1b-a91f-4d0b-b6af-1e463774b81a)

## Notes

[MAASENG-2756]: https://warthogs.atlassian.net/browse/MAASENG-2756?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/app/machines/components/MachineForms/MachineActionFormWrapper/DeployForm/DeployForm.tsx
+++ b/src/app/machines/components/MachineForms/MachineActionFormWrapper/DeployForm/DeployForm.tsx
@@ -171,6 +171,7 @@ export const DeployForm = ({
       onSuccess={clearSidePanelContent}
       processingCount={processingCount}
       selectedCount={machines ? machines.length : selectedCount ?? 0}
+      showProcessingCount={false}
       validationSchema={DeploySchema}
     >
       <DeployFormFields />
--- a/src/app/machines/components/MachineForms/MachineActionFormWrapper/DeployForm/DeployFormFields/DeployFormFields.tsx
+++ b/src/app/machines/components/MachineForms/MachineActionFormWrapper/DeployForm/DeployFormFields/DeployFormFields.tsx
@@ -286,7 +286,6 @@ export const DeployFormFields = (): JSX.Element => {
             </Col>
           </Row>
         )}
-        <hr className="u-sv1" />
       </div>
     </>
   );


src/app/base/components/SectionHeader/_index.scss
```diff
@@ -1,6 +1,11 @@
 @mixin SectionHeader {
   .section-header {
     padding-top: $spv--medium;
+    position: sticky;
+    top: 0;
+    background: $color-x-light;
+    border-bottom: $border;
+    z-index: 2; // Ensure title displays above notification bar when scrolled
```
petermakowski, line 8: This change should land in `maas-react-components` as part of the [content section](https://canonical.github.io/maas-react-components/?path=/docs/sections-contentsection--docs#with-form ).

But first we should confirm with @dgtlntv if that's how we want this component to behave. Content section is also used across various forms and it's not clear if it should have a sticky header everywhere (or at all).

petermakowski (commented):
This is a step in the right direction! Appreciate before and after videos, very helpful. Please take a look at the comment below.

src/app/base/components/SectionHeader/_index.scss
dgtlntv: That the header is sticky is nice imo. I can't say at the moment whether it's something that should be used everywhere, as I don't know in what other contexts the content area is used.

petermakowski: @Jay-Topher Thanks. But please open a PR for maas-react-components with this change. I'd suggest the sticky header behaviour being applied through a prop passed onto the Header component. It should probably be disabled by default.

petermakowski (approved):

PR #5326, opened 2024-02-29
fix: composing machine does not update list lp#1842524

petermakowski:
## Done
This ensures we refetch machine list on compose success via `dispatch(machineActions.invalidateQueries());` and update the list immediately. It also adds `stale: false` to success actions to ensure consistent behaviour.

- fix: composing machine does not update list

## QA steps

- [ ] Go to KVM
- [ ] Add a new machine using "Add VM" button
- [ ] Ensure that it's displayed on the list immediately when operation is successful

## Fixes

Fixes: lp#1842524


sparkiegeek:
> Fixes: lp#231842524

:see_no_evil:

Jay-Topher:
Having difficulties QAing this (no cores available), @ndv99 could you have a look?

ndv99:
Currently setting up a local env to QA this in, hopefully shouldn't be too long before I can +1 this

ndv99 (approved):
Finally, I can say: LGTM!

PR #5336, opened 2024-03-11
feat: move machine details forms to side panel

Jay-Topher:
## Done
- Mark connected
- Edit Physical
- Add VLAN
- Remove Physical
Have all been moved to side panels

## QA steps

- [ ] Visit machine details page
- [ ] Navigate to the network tab
- [ ] Ensure all the action forms on the table open up in a side panel
- [ ] Ensure the forms submit successfully

## Fixes

Fixes: [Dev: Display Edit physical, Remove bridge forms in a side panel](https://warthogs.atlassian.net/browse/MAASENG-2796)

### Before
image
image
image

### After
image
image
image

### DIFF
--- a/src/app/base/components/ModelActionForm/ModelActionForm.tsx
+++ b/src/app/base/components/ModelActionForm/ModelActionForm.tsx
@@ -1,3 +1,5 @@
+import type { ReactNode } from "react";
+
 import { Col, Row } from "@canonical/react-components";
 
 import type { Props as FormikFormProps } from "@/app/base/components/FormikForm/FormikForm";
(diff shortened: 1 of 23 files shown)


ndv99:
Just my take, but I really think the "Edit physical"  form should use the "large" side panel size, if we want to display that table at the top.

Jay-Topher:
> Just my take, but I really think the "Edit physical" form should use the "large" side panel size, if we want to display that table at the top.

Sounds fair, but do we really need a table for this use case?

ndv99:
> > Just my take, but I really think the "Edit physical" form should use the "large" side panel size, if we want to display that table at the top.
> 
> Sounds fair, but do we really need a table for this use case?

That's also a fair point. I'll leave it to you then - either make the side panel larger, or keep the current size and remove that table row from the form

PR #5344, opened 2024-03-13
fix: enable name column on partitions table on side panel close

Jay-Topher:
## Done
- Enable name column on partitions table on side panel close

## QA steps

- [ ] Goto `/machine/:id/storage` for a ready machine
- [ ] Select a disk under the available disks and partitions table
- [ ] Click on create volume group
- [ ] Ensure that the name column of the table is disabled
- [ ] Close the side panel and ensure that the name column is enabled

## Fixes

Fixes: [Dev: Enable name column after side panel is closed](https://warthogs.atlassian.net/browse/MAASENG-2799)

## Screenshots
image
image

### DIFF
--- a/src/app/base/components/node/StorageTables/AvailableStorageTable/AvailableStorageTable.tsx
+++ b/src/app/base/components/node/StorageTables/AvailableStorageTable/AvailableStorageTable.tsx
@@ -20,6 +20,7 @@ import ActionConfirm from "@/app/base/components/node/ActionConfirm";
 import DiskBootStatus from "@/app/base/components/node/DiskBootStatus";
 import DiskNumaNodes from "@/app/base/components/node/DiskNumaNodes";
 import DiskTestStatus from "@/app/base/components/node/DiskTestStatus";
+import { useSidePanel } from "@/app/base/side-panel-context";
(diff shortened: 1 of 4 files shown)


src/app/base/components/node/StorageTables/AvailableStorageTable/AvailableStorageTable.tsx
```diff
@@ -286,8 +287,9 @@ const AvailableStorageTable = ({
   const dispatch = useDispatch();
   const [expanded, setExpanded] = useState<Expanded | null>(null);
   const [selected, setSelected] = useState<(Disk | Partition)[]>([]);
-  const [bulkAction, setBulkAction] = useState<BulkAction | null>(null);
+  const [, setBulkAction] = useState<BulkAction | null>(null);
```
ndv99, line 290: If `bulkAction` was only used for deciding whether or not actions should be disabled, and that value is now unneeded, we should also be able to remove all calls for `setBulkAction` without any side effects, and then we can remove this state entirely.

ndv99 (changes requested):
One more change, and I think this should be good to merge.

ndv99 (approved):
LGTM

PR #5358, opened 2024-03-19
fix: login error message spacing MAASENG-2885

petermakowski:
## Done
fix: login error message spacing
  - test: add login error message test
  - test: refactor Login.test.tsx to use renderWithBrowserRouter

## QA steps

- [x] Use incorrect credentials to login
- [x] Press login
- [x] Verify that the error message is displayed in the center with a correct padding

## Fixes

Fixes: MAASENG-2885

### Before
Google Chrome screenshot 001675@2x

### After
Google Chrome screenshot 001673@2x

### DIFF
--- a/src/app/App.tsx
+++ b/src/app/App.tsx
@@ -14,7 +14,6 @@ import PageContent from "./base/components/PageContent/PageContent";
 import SectionHeader from "./base/components/SectionHeader";
 import ThemePreviewContextProvider from "./base/theme-context";
 import { MAAS_UI_ID } from "./constants";
-import { formatErrors } from "./utils";
 
 import AppSideNavigation from "@/app/base/components/AppSideNavigation";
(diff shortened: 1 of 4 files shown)


Jay-Topher:
QA works as expected, but I noticed that the whole page repaints while making the login request, we should make this better in a separate ticket

Jay-Topher (approved):
+1

petermakowski:
> QA works as expected, but I noticed that the whole page repaints while making the login request, we should make this better in a separate ticket

@Jay-Topher It won't be easy to change. This was intentional and prevents flashing of the loading screen when websocket connection is lost and restored. See: https://github.com/canonical/maas-ui/blob/44467ed1ceebdc34d6232e53e0658b527c6d6d42/src/app/App.tsx#L124-L126

Created an issue for this: https://warthogs.atlassian.net/browse/MAASENG-2887

PR #5367, opened 2024-03-21
fix: align side panel content section header

petermakowski:
## Done
- fix: side panel content section

## QA steps

- [ ] Verify that the header of the side panel is aligned with the header of the main section

## Fixes

Fixes: 

### Before
Google Chrome screenshot 001717@2x

### After
Google Chrome screenshot 001715@2x


ndv99 (approved):
LGTM -  thanks for fixing this

Jay-Topher (approved):

PR #5377, opened 2024-03-26
fix: change source side panel lp#2058340

petermakowski:
## Done
- fix: change source side panel lp#2058340
  - wrap application layout overrides in l-application for higher specificity

## QA steps

- [ ] Go to images
- [ ] Press Change source
- [ ] Paste in `http://images.maas.io/ephemeral-v3/candidate` as URL
- [ ] Press connect
- [ ] Verify that the next step is displayed using a large side panel size

## Fixes

Fixes: 

### Before

Google Chrome screenshot 001739@2x

### After

Google Chrome screenshot 001737@2x
Google Chrome screenshot 001735@2x


ndv99 (approved):
LGTM

PR #5383, opened 2024-04-02
chore: replace react-portal with react-use-portal

Jay-Topher:
## Done
- Replaced instance of react-portal with react-use-portal

## QA steps

- [ ] Ensure that app works as expected
- [ ] Ensure that the "Create new tag" dialog in the controller configuration page opens as expected

## Fixes

Fixes: [1599](https://warthogs.atlassian.net/browse/MAASENG-1599)

## Screenshots
image


package.json
```diff
@@ -110,7 +110,6 @@
     "@types/pluralize": "0.0.33",
     "@types/react": "18.2.73",
     "@types/react-dom": "18.2.23",
-    "@types/react-portal": "4.0.7",
```
petermakowski, line 113: Don't forget to remove `react-portal` as well.

src/app/base/components/NodeConfigurationFields/NodeConfigurationFields.tsx
```diff
@@ -50,14 +50,15 @@ const NodeConfigurationFields = (): JSX.Element => {
             name="description"
           />
           <TagIdField
+            addTagRef={ref}
```
petermakowski, line 53: Could you explain what happened here, why is this necessary?

Jay-Topher: The normal implementation calls for the event object to be passed to the `openPortal` function, in a situation where this can't be done, the alternative is to pass a ref as shown [here](https://www.npmjs.com/package/react-useportal#usage-with-a-ref). Without the ref, it throws an error in the console (attached below)
image

petermakowski: I can see you added the ref to the `li` element.

It's not an interactive element.`ref` needs to be added to an element you're _interacting with_ (that triggers the portal), in this case it's a add new tag _button_. This is for many reasons, one of which accessibility.

I'd recommend to modify the existing `onAddNewTag` to include the event so that it can be passed to the openPortal handler, e.g.:

```tsx
onAddNewTag={(name) => (event) => {
  setNewTagName(name);
  openPortal(event);
 }}
```
By using a curried function, you ensure that any handler that doesn't care about the event will still work as expected. 

That way, you can avoid passing down the ref entirely.

src/app/base/components/TagSelector/TagSelector.tsx
```diff
@@ -13,6 +13,7 @@ export type Tag = {
 };
 
 export type Props = {
+  addTagRef?: React.RefObject<HTMLLIElement>;
```
petermakowski, line 16:
```suggestion
```

src/app/base/components/TagSelector/TagSelector.tsx
```diff
@@ -66,6 +67,7 @@ const generateDropdownItems = ({
   tags,
   updateTags,
   generateDropdownEntry,
+  addTagRef,
```
petermakowski, line 70:
```suggestion
```

src/app/base/components/TagSelector/TagSelector.tsx
```diff
@@ -202,6 +205,7 @@ export const TagSelector = ({
   showSelectedTags = true,
   tags = [],
   disabledTags = [],
+  addTagRef,
```
petermakowski, line 208:
```suggestion
```

src/app/base/components/TagSelector/TagSelector.tsx
```diff
@@ -237,6 +241,7 @@ export const TagSelector = ({
     tags,
     updateTags,
     generateDropdownEntry,
+    addTagRef,
```
petermakowski, line 244:
```suggestion
```

src/app/base/components/TagSelector/TagSelector.tsx
```diff
@@ -20,7 +20,7 @@ export type Props = {
   help?: string;
   initialSelected?: Tag[];
   label?: string | null;
-  onAddNewTag?: (name: string) => void;
+  onAddNewTag?: (name: string, event?: React.SyntheticEvent) => void;
```
petermakowski, line 23:
```suggestion
  onAddNewTag?: (name: string) => (event: React.SyntheticEvent) => void;
```

src/app/base/components/TagSelector/TagSelector.tsx
```diff
@@ -272,7 +272,7 @@ export const TagSelector = ({
                 if (allowNewTags) {
                   const cleanedFilter = sanitiseFilter(filter);
                   if (onAddNewTag) {
-                    onAddNewTag(cleanedFilter.name);
+                    onAddNewTag(cleanedFilter.name, e);
```
petermakowski, line 275:
```suggestion
                    onAddNewTag(cleanedFilter.name)?.(e);
```

src/app/base/components/TagSelector/TagSelector.test.tsx
```diff
@@ -179,7 +179,10 @@ describe("TagSelector", () => {
       "new-tag"
     );
     await userEvent.click(screen.getByTestId("new-tag"));
-    expect(onAddNewTag).toHaveBeenCalledWith("new-tag");
+    expect(onAddNewTag).toHaveBeenCalledWith(
+      "new-tag",
+      expect.objectContaining({ type: "click", _reactName: "onClick" })
+    );
```
petermakowski, line 185:
```suggestion
    expect(onAddNewTag).toHaveBeenCalledWith("new-tag");
```

petermakowski (commented):
Like I explained in the previous review, I'd recommend to use a curried function here. Accepting my inline review suggestions should make this work.

petermakowski (approved):

PR #5388, opened 2024-04-05
feat: react router data api investigation WIP

ndv99:
# This is a *really* rough proof-of-concept; it feels a bit laggy when running the dev server on my machine, but I imagine this would probably improve when done properly.

I'm also just using this to see what happens to tests. Should be interesting.

## Done
- 

## QA steps

- [ ] 

## Fixes

Fixes:


ndv99:
Wow, only two failing files, and they were exactly the ones I had expected to fail. I was really under the impression that this would be a lot more difficult... Actually, let me not jinx that.

PR #5389, opened 2024-04-05
fix: VLAN reserve range side panels

petermakowski:
## Done
- fix: VLAN reserve range side panels
  - move SubnetActionForms.tsx out of the header directory

## QA steps

- [ ] Go to a VLAN details page (e.g. `/MAAS/r/vlan/35` in demos)
- [ ] Click reserve range
- [ ] Verify you can submit the form and range is added to the list
- [ ] Ensure you can edit and delete it successfully
- [ ] Add a new VLAN and ensure that Delete VLAN side panel works correctly


ndv99:
I can delete a VLAN successfully, but the side panel does not close after this.

src/app/subnets/views/VLANDetails/VLANActionForms/VLANActionForms.tsx
```diff
@@ -0,0 +1,28 @@
...
+  activeForm,
+  setSidePanelContent,
+}: VLANActionFormProps) => {
+  return activeForm === VLANActionTypes.DeleteVLAN ? (
+    <VLANDeleteForm closeForm={() => setSidePanelContent(null)} id={id} />
+  ) : activeForm === VLANActionTypes.ReserveRange ? (
+    <ReservedRangeForm setActiveForm={setSidePanelContent} />
+  ) : activeForm === VLANActionTypes.DeleteReservedRange ? (
+    <ReservedRangeDeleteForm setActiveForm={setSidePanelContent} />
+  ) : null;
```
Jay-Topher, line 25: We should probably represent this better, it's difficult to read

petermakowski: You're right, wanted to limit the scope as it's a bugfix. Updated.

ndv99: Agreed, a simple switch/case would be much more readable here

ndv99:
Idk if this is just happening for me, but I'm still getting the issue with the "Delete reserved range" side panel not closing on success

PR #5405, opened 2024-04-15
fix: html error formatting MAASENG-2935

petermakowski:
## Done
In some rare scenarios HTML content might be returned as a response. This change makes sure that we display text only and not entire raw HTML string.

- fix: html error formatting MAASENG-2935

## QA steps

HTML formatting is hard QA without simulating an actual nginx error and also already covered by tests. Focus on verifying that other errors are displayed as before.

- Attempt to login using incorrect credentials
- Ensure that login validation error is displayed correctly

## Fixes

Fixes: MAASENG-2935

### Before
image

### After
Google Chrome screenshot 001831@2x

### DIFF
--- a/src/app/utils/formatErrors.ts
+++ b/src/app/utils/formatErrors.ts
@@ -73,7 +73,7 @@ const parseHtmlToText = (htmlContent: string): string | null => {
       // Trim any leading or trailing whitespace from the resulting text
       .trim();
 
-    return strippedText !== "" ? strippedText : null;
+    return strippedText || null;
   }
 
   // Return the original content if the body tag is not found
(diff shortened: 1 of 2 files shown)


src/app/utils/formatErrors.ts
```diff
@@ -60,10 +60,34 @@ const parseObjectError = (
...
+    const bodyContent = bodyMatch[1];
+    const strippedText = bodyContent
+      // Remove all HTML tags from the body content
+      .replace(/<[^>]+>/g, "")
+      // Collapse multiple consecutive whitespace characters into a single space
+      .replace(/\s+/g, " ")
+      // Trim any leading or trailing whitespace from the resulting text
+      .trim();
+
+    return strippedText !== "" ? strippedText : null;
```
Jay-Topher, line 76:
```suggestion
    return strippedText || null;
```

Jay-Topher (approved):
+1

PR #5415, opened 2024-04-23
feat: add static dhcp table

Jay-Topher:
## Done
- Added Static DHCP Lease table

## QA steps
A flag has been added to prevent this from showing in production (per the acceptance criteria). as a result, this should be tested in `dev`
- [ ] Visit Reserved IPs page
- [ ] Ensure that the Static DHCP table renders as expected with the mock data.

## Fixes

Fixes: [MAASENG-2974](https://warthogs.atlassian.net/browse/MAASENG-2974)

## Screenshots
image

## Notes
The table height needs to be adjusted on MRC to take off excess white space between the tables as seen in the screenshot

[MAASENG-2974]: https://warthogs.atlassian.net/browse/MAASENG-2974?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/app/store/subnet/types/base.ts
+++ b/src/app/store/subnet/types/base.ts
@@ -9,7 +9,12 @@ import type {
   TimestampedModel,
   TimestampFields,
 } from "@/app/store/types/model";
-import type { NetworkInterface, Node, NodeType } from "@/app/store/types/node";
+import type {
+  NetworkInterface,
+  Node,
+  NodeType,
+  SimpleNode,
+} from "@/app/store/types/node";
 import type { EventError, GenericState } from "@/app/store/types/state";
 import type { User } from "@/app/store/user/types";
 import type { VLAN } from "@/app/store/vlan/types";
@@ -119,4 +124,15 @@ export type SubnetState = GenericState<Subnet, APIError> & {
   active: Subnet[SubnetMeta.PK] | null;
   eventErrors: EventError<Subnet, APIError, SubnetMeta.PK>[];
   statuses: SubnetStatuses;
+};
+
+export type StaticDHCPLease = {
+  id: number;
+  comment?: string | null;
+  ip_address: string;
+  mac_address: string;
+  interface?: string | null;
+  node?: SimpleNode | null;
+  usage?: string | null;
+  owner: string;
 };
--- a/src/app/subnets/views/SubnetDetails/StaticDHCPLease/StaticDHCPLease.tsx
+++ b/src/app/subnets/views/SubnetDetails/StaticDHCPLease/StaticDHCPLease.tsx
@@ -1,12 +1,16 @@
 import { MainToolbar } from "@canonical/maas-react-components";
 import { Button } from "@canonical/react-components";
+import { array } from "cooky-cutter";
 
 import { SubnetActionTypes, SubnetDetailsSidePanelViews } from "../constants";
 
 import { useSidePanel } from "@/app/base/side-panel-context";
+import StaticDHCPTable from "@/app/subnets/views/SubnetDetails/StaticDHCPLease/StaticDHCPTable";
+import { staticDHCPLease } from "@/testing/factories/subnet";
 
 const StaticDHCPLease = () => {
   const { setSidePanelContent } = useSidePanel();
+  const getMockStaticDHCPLeases = array(staticDHCPLease, 5);
 
   return (
     <>
@@ -27,6 +31,7 @@ const StaticDHCPLease = () => {
           </Button>
         </MainToolbar.Controls>
       </MainToolbar>
+      <StaticDHCPTable staticDHCPLeases={getMockStaticDHCPLeases()} />
     </>
   );
 };
--- a/src/app/subnets/views/SubnetDetails/StaticDHCPLease/StaticDHCPTable/StaticDHCPTable.tsx
+++ b/src/app/subnets/views/SubnetDetails/StaticDHCPLease/StaticDHCPTable/StaticDHCPTable.tsx
@@ -1,0 +1,82 @@
+import { DynamicTable, TableCaption } from "@canonical/maas-react-components";
+import { Link } from "react-router-dom";
+
+import TableActions from "@/app/base/components/TableActions";
+import type { StaticDHCPLease } from "@/app/store/subnet/types/base";
+
+const headers = [
+  { content: "IP Address", className: "ip-col", sortKey: "ip_address" },
+  { content: "MAC Address", className: "mac-col", sortKey: "mac_address" },
+  { content: "Node", className: "node-col", sortKey: "node" },
+  {
+    content: "Interface",
+    className: "interface-col",
+    sortKey: "interface",
+  },
+  { content: "Usage", className: "usage-col", sortKey: "usage" },
+  { content: "Owner", className: "oner-col", sortKey: "owner" },
+  { content: "Comment", className: "comment-col", sortKey: "comment" },
+  { content: "Actions", className: "actions-col" },
+] as const;
+
+const generateRows = (dhcpLeases: StaticDHCPLease[]) =>
+  dhcpLeases.map((dhcpLease) => {
+    return (
+      <tr key={dhcpLease.mac_address}>
+        <td>{dhcpLease.ip_address}</td>
+        <td>{dhcpLease.mac_address}</td>
+        <td>
+          {dhcpLease?.node ? (
+            <Link to="#">
+              <strong>{dhcpLease.node.hostname}</strong>.
+              {dhcpLease.node.fqdn.split(".")[1]}
+            </Link>
+          ) : (
+            "—"
+          )}
+        </td>
+        <td>{dhcpLease.interface}</td>
+        <td>{dhcpLease.usage || "—"}</td>
+        <td>{dhcpLease.owner}</td>
+        <td>{dhcpLease.comment || "—"}</td>
+        <td>
+          <TableActions onDelete={() => {}} onEdit={() => {}} />
+        </td>
+      </tr>
+    );
+  });
+
+type Props = {
+  staticDHCPLeases: StaticDHCPLease[];
+};
+
+const StaticDHCPTable = ({ staticDHCPLeases }: Props) => {
+  return (
+    <DynamicTable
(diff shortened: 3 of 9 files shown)


ndv99:
I think the offset calculation function in `DynamicTable` might need some work, or at least something like being able to pass in a prop to alter its behaviour. In its current state, the tbody is definitely too big and leaves a lot of whitespace

For a really insane example, this is what happens if the window is 2400px in height: 
image

ndv99:
After some more thought, I think the best way to solve this would be to add a `tableHeight` prop to DynamicTable, using `vh` as the unit. It should default to `100` if not provided, and then this can be passed into the `style` prop on the tbody, instead of just always subtracting from 100vh.

@petermakowski what do you think?

src/app/subnets/views/SubnetDetails/SubnetStaticIPs/SubnetStaticIPs.tsx
```diff
@@ -27,6 +34,7 @@ const SubnetStaticIPs = () => {
           </Button>
         </MainToolbar.Controls>
       </MainToolbar>
+      {IS_DEV && <StaticDHCPTable staticDHCPLeases={mockStaticDHCPLeases} />}
```
ndv99, line 37: All the static IP work is already hidden if `VITE_APP_STATIC_IPS_ENABLED` isn't true, so you can remove the condition here and just always display the table

Jay-Topher: Okay, this means we'd need to update the acceptance criteria for this.

Jay-Topher:
> After some more thought, I think the best way to solve this would be to add a `tableHeight` prop to DynamicTable, using `vh` as the unit. It should default to `100` if not provided, and then this can be passed into the `style` prop on the tbody, instead of just always subtracting from 100vh.
> 
> @petermakowski what do you think?

This sounds like a good idea, we could call it `tableHeightInVH` or `tableVH` to make the unit more obvious.

petermakowski:
> @petermakowski what do you think?

@Jay-Topher @ndv99 Please wait with merging this PR. I'll get back to this later today.

src/app/subnets/views/SubnetDetails/SubnetStaticIPs/StaticDHCPTable/StaticDHCPTable.test.tsx
```diff
@@ -0,0 +1,14 @@
+import StaticDHCPTable from "./StaticDHCPTable";
+
+import { renderWithBrowserRouter, screen } from "@/testing/utils";
+
+it("renders a static DHCP table", () => {
+  renderWithBrowserRouter(<StaticDHCPTable staticDHCPLeases={[]} />);
```
petermakowski, line 6: Please add a test for a scenario where there are rows in the table. This will require you adding a new factory (similar to the ones in `src/testing/factories/subnet.ts`.

src/app/subnets/views/SubnetDetails/constants.ts
```diff
@@ -59,3 +61,67 @@ export type SubnetDetailsSidePanelContent = SidePanelContent<
...
+    id: 5,
+    comment: "Cillum adipisicing nisi deserunt dolore ex.",
+    ip_address: "238.59.124.16",
+    mac_address: "22:45:92:55:21",
+    interface: "eno2",
+    node: "hoover",
+    usage: "",
+    owner: "cgrabowski",
+  },
+];
```
petermakowski, line 127: Please create and use a factory to generate test data.

petermakowski:
Tab name should now be "Address reservation" and not "Reserved IPs", table name "Static DHCP leases" as per design: https://app.zeplin.io/project/6616700150a22355a6c55c7f/screen/6622792ea06cad349d46714e

Could you please update, ideally including rename of components so that it's not confusing (`SubnetStaticIPs` -> `StaticDHCPLeases`)

petermakowski (changes requested):
Please review the changes requested above.

petermakowski:
> I think the offset calculation function in `DynamicTable` might need some work, or at least something like being able to pass in a prop to alter its behaviour. In its current state, the tbody is definitely too big and leaves a lot of whitespace

@ndv99  Thanks for raising this. DynamicTable was only intended and tested for use as the only table on the page at a time. Let's leave this discussion out of this PR and move over to maas-react-components: https://github.com/canonical/maas-react-components/issues/128

Jay-Topher:
> Tab name should now be "Address reservation" and not "Reserved IPs", table name "Static DHCP leases" as per design: https://app.zeplin.io/project/6616700150a22355a6c55c7f/screen/6622792ea06cad349d46714e
> 
> Could you please update, ideally including rename of components so that it's not confusing (`SubnetStaticIPs` -> `StaticDHCPLeases`). Also the button should be renamed from "Reserve static IP" to "Reserve static DHCP lease".

I started tackling this on a different ticker because it's beyond the scope of this one

src/app/store/subnet/types/base.ts
```diff
@@ -120,3 +125,14 @@ export type SubnetState = GenericState<Subnet, APIError> & {
...
   statuses: SubnetStatuses;
 };
+
+export type StaticDHCPLease = {
+  id: number;
+  comment: string | null;
+  ip_address: string;
+  mac_address: string;
+  interface: string;
+  node: SimpleNode;
```
ndv99, line 135: I had some reservations about passing through a node, but given we'll need the ID as well to make the links work, I think this should be fine for now.

src/app/store/subnet/types/base.ts
```diff
@@ -120,3 +125,14 @@ export type SubnetState = GenericState<Subnet, APIError> & {
...
+  usage: string | null;
+  owner: string;
+};
```
ndv99, line 138: `node`, `interface`, `usage` and `comment` should all be optional. A reservation can be made for a node that hasn't been commissioned yet, so that data isn't guaranteed.

petermakowski: Let's just make sure all optional keys have `| null`, (`node?: SimpleNode | null;`) and not optional key (`node?: SimpleNode;`). With an optional key it can be unclear whether the property was omitted intentionally or by mistake. `null` states explicitly that a property has been intentionally omitted.

src/app/store/subnet/types/base.ts
```diff
@@ -120,3 +125,14 @@ export type SubnetState = GenericState<Subnet, APIError> & {
...
 };
+
+export type StaticDHCPLease = {
+  id: number;
+  comment?: string | null;
+  ip_address: string;
+  mac_address: string;
+  interface?: string;
+  node?: SimpleNode;
+  usage?: string | null;
```
petermakowski, line 136: Let's make sure all optional keys have `| null`, (`node?: SimpleNode | null;`) and not optional key (`node?: SimpleNode;`). With an optional key it can be unclear whether the property was omitted intentionally or by mistake. `null` states explicitly that a value has been intentionally omitted.

petermakowski (changes requested):

src/app/subnets/views/SubnetDetails/StaticDHCPLease/StaticDHCPTable/StaticDHCPTable.tsx
```diff
@@ -0,0 +1,82 @@
...
+  {
+    content: "Interface",
+    className: "interface-col",
+    sortKey: "interface",
+  },
+  { content: "Usage", className: "usage-col", sortKey: "usage" },
+  { content: "Owner", className: "oner-col", sortKey: "owner" },
+  { content: "Comment", className: "comment-col", sortKey: "comment" },
+  { content: "Actions", className: "actions-col" },
+];
```
petermakowski, line 20:
```suggestion
] as const;
```

src/testing/factories/subnet.ts
```diff
@@ -110,3 +112,14 @@ export const subnet = extend<TimestampedModel, BaseSubnet>(timestampedModel, {
...
+export const staticDHCPLease = define<StaticDHCPLease>({
+  id: sequence,
+  comment: "random comment",
+  ip_address: (i: number) => `192.168.1.${i}`,
+  mac_address: (i: number) => `00:00:00:00:00:${i}`,
+  interface: (i: number) => `eth${i}`,
+  usage: "Device",
+  node: simpleNode,
+  owner: "test-owner",
+});
```
petermakowski, line 125: Much better 👍

src/app/store/subnet/types/base.ts
Jay-Topher: Adding `null` to the node type throws a lint error when creating the DHCPLease factory, could you please advise on this?

petermakowski: You could try resolving it with `() => simpleNode() ?? null,` although there is likely a better way to do this.

petermakowski: e.g. something like:
```tsx
export const staticDHCPLease = define<StaticDHCPLease>({
  node: derive<StaticDHCPLease, SimpleNode | null>(simpleNode),
  ```

Comment (Jay-Topher) [reply src/app/store/subnet/types/base.ts]: Thanks, this works

File: src/app/subnets/views/SubnetDetails/StaticDHCPLease/StaticDHCPLease.tsx [hunk 9]
```diff
@@ -27,6 +31,7 @@ const StaticDHCPLease = () => {
           </Button>
         </MainToolbar.Controls>
       </MainToolbar>
+      <StaticDHCPTable staticDHCPLeases={getMockStaticDHCPLeases()} />
```
petermakowski, line 34: This will need to be updated in order for tree-shaking to work correctly (make sure we only import mock data in development). As is, cooky-cutter is included in the production bundle increasing its size. Long term solution is to serve mock data using `msw`, but it's outside of scope of this. I'll followup with a PR to update it.

petermakowski (approved):

PR #5424, opened 2024-05-06
chore: upgrade prettier to v3

petermakowski:
## Done
- chore: upgrade prettier to v3
  - set "trailingComma" to "es5" to maintain existing style (default has changed as of v3 - https://prettier.io/docs/en/options.html#trailing-commas )
  - fix all lint errors


ndv99 (approved):
LGTM

PR #5429, opened 2024-05-09
refactor: generate common reducers

petermakowski:
## Done
- refactor: generateCommonReducers to use object argument
- refactor: remove redundant state comments


ndv99 (approved):
LGTM, I definitely prefer the props being in this format.

PR #5448, opened 2024-05-29
fix: DHCPTable, TagForm, VLANDetails selectors

petermakowski:
## Done
- fix: DHCPTable, TagForm, VLANDetails selectors

This pull request fixes the selectors in the DHCPTable, TagForm, and VLANDetails components.

In VLANDetails.tsx, the call to subnetSelectors.getByIds has been updated to pass null instead of an empty array when vlan?.subnet_ids is falsy.

In TagForm.tsx, the events array is now prevented from being re-created on every render.

## QA steps

- [ ] Go to VLAN Details page and ensure that DHCP snippets are displayed and there is no warning in the console

## Fixes

Fixes:


ndv99 (approved):
LGTM

PR #5455, opened 2024-06-10
test: add coverage report upload step

petermakowski:
## Done
- test: add coverage report upload step

I mistakenly removed this step before making final changes to the workflow; unfortunately testing workflows before merge is not straightforward.


ndv99 (approved):
LGTM

PR #5459, opened 2024-06-12
ci: fix links checker workflow

petermakowski:
## Done
- ci: fix links checker

This adds `push` instead of `branches` directly under `on`


ndv99 (approved):
LGTM

Issue #5464, opened 2024-06-13
Unusual spacing in parsing of multiple distributions or components when adding a package repository


wyattrees:
**Describe the bug**
When adding a new package repository with multiple distributions or components via the MAAS UI, you need to enter them comma-separated, but with the spaces before the commas, like so: `focal ,jammy ,noble`

**maas-ui version**
3.6.0

**MAAS version**
3.4

**LP issue**
N/A

**To Reproduce**

1. Go to Settings->Package Repos->Add Repository
2. For both the Distributions and Components sections, enter any values separated with commas, but without the space before the comma (e.g. `focal, jammy, noble` or `focal,jammy,noble`)
3. Click Save Repository
4. Inspecting the WS request sent to MAAS will show a singleton list, `["focal, jammy, noble"]`. If you use spaces before the commas in step 2, the list is `["focal", "jammy", "noble"]`

**Expected behavior**
The list sent to MAAS is `["focal", "jammy", "noble"]` when spaces are either omitted or placed after commas.

**Screenshots**
![ui-parsing](https://github.com/canonical/maas-ui/assets/25443786/e8a526ff-5dcd-4a0d-8b94-326dfd5ad2c3)

**Desktop (please complete the following information):**
 - OS: Ubuntu 22.04
 - Browser Chrome
 - Version 126.0.6478.55-1

petermakowski:
Thanks for reporting this! FYI we now use launchpad for bug tracking in maas-ui. https://bugs.launchpad.net/maas-ui

PR #5470, opened 2024-06-19
fix(settings): fix list spacing #5464

petermakowski:
## Done
- Fix values spacing when adding a package repository
- update the parsing logic to correctly split the values and remove any leading or trailing spaces

This fixes an issue where the MAAS UI incorrectly parses comma-separated distributions and components when adding a new package repository.

Fixes: #5464

## QA

1. Go to Settings->Package Repos->Add Repository
2. For both the Distributions and Components sections, enter any values separated with commas (e.g. `focal, jammy, noble` or `focal,jammy,noble`)
3. Click Save Repository
4. Inspecting the WS request sent to MAAS and ensure the list  is `["focal", "jammy", "noble"]`

### DIFF
--- a/src/app/utils/parseCommaSeparatedValues.ts
+++ b/src/app/utils/parseCommaSeparatedValues.ts
@@ -1,0 +1,6 @@
+export const parseCommaSeparatedValues = (input: string): string[] => {
+  return input
+    .split(",")
+    .map((s) => s.trim())
+    .filter(Boolean);
+};
--- a/src/app/settings/views/Repositories/RepositoryForm/RepositoryForm.test.tsx
+++ b/src/app/settings/views/Repositories/RepositoryForm/RepositoryForm.test.tsx
@@ -338,4 +338,52 @@ describe("RepositoryForm", () => {
     ).toBe(true);
     expect(actions.some((action) => action.type === "message/add")).toBe(true);
   });
+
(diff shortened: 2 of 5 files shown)


src/app/utils/parseCommaSeparatedValues.ts
```diff
@@ -0,0 +1,6 @@
+export const parseCommaSeparatedValues = (input: string): string[] => {
+  return input
+    .split(",")
+    .map((s) => s.trim())
+    .filter(Boolean);
```
ndv99, line 5: What does this line do?

ndv99 (approved):
LGTM, just one question below

src/app/utils/parseCommaSeparatedValues.ts
petermakowski: `Boolean` converts any given value to a boolean type. It returns true for truthy values and false for falsy values. Here it's used as an argument in `.filter()`effectively filtering out all falsy values from the array.

petermakowski: Examples below might help if still unclear.

```tsx
"first, , third".split(",")
    .map((s) => s.trim())
```
`['first', '', 'third']`

---

```tsx
"first, , third".split(",")
    .map((s) => s.trim())
    .filter(Boolean);
```
`['first', 'third']`

Issue #5465, opened 2024-06-13
Machine's last commission time shown as time since first commission


wyattrees:
**Describe the bug**
If a machine has had multiple commissioning events on different dates, the UI will display "Last commissioned: {time since _first_ commission event}"

**maas-ui version**
3.6.0

**MAAS version**
3.4

**LP issue**
LP#2069226

**To Reproduce**
1. Find a machine with commissioning events on different days (in labmaas, `chladni` is an example)
2. Click on the machine name, then the "Commissioning" tab
3. Note the "Last commissioned" time in the bottom right corner corresponds to the time since the earliest event in the table

**Expected behavior**
The "Last commissioned" time corresponds to the time since the _latest_ commissioning event

**Screenshots**
From the LP bug: 
<img width="1569" alt="last-commissioned-bug" src="https://github.com/canonical/maas-ui/assets/25443786/55df806f-62e0-499e-b2ac-2de03a57c695">

**Desktop (please complete the following information):**
 - OS: Ubuntu 22.04
 - Browser chrome
 - Version 126.0.6478.55-1

petermakowski:
Thanks for reporting this! FYI we now use launchpad for bug tracking in maas-ui. https://bugs.launchpad.net/maas-ui

PR #5471, opened 2024-06-19
fix: last commissioned time in the status bar

petermakowski:
## Done
- fix: last commissioned time in the status bar

Fixes the bug where the last commissioned time displayed in the UI for machines with multiple commissioning events shows the time since the first commission event instead of the latest commission event.

The issue is resolved by updating the logic in the `getLastCommissionedString` function to correctly calculate the last commissioned time based on the latest commissioning event. The function now checks for commissioning events in the machine's event list and selects the event with the latest timestamp. If no commissioning events are found, it falls back to using the `commissioning_start_time` property.

Unit tests have been added to ensure the correct calculation of the last commissioned time and to handle cases with invalid commissioning event timestamps.

Fixes: #5465.


src/app/base/components/StatusBar/StatusBar.tsx
```diff
@@ -26,14 +26,44 @@ import type { UtcDatetime } from "@/app/store/types/model";
 import { NodeStatus } from "@/app/store/types/node";
 import { formatUtcDatetime, getTimeDistanceString } from "@/app/utils/time";
 
-const getLastCommissionedString = (machine: MachineDetails) => {
+const getLastCommissionedString = (machine: MachineDetails): string => {
```
ndv99, line 29: Bit of a nitpick, but I think this would be a lot more readable if placed *after* `findLastCommissioningTime` and `formatLastCommissioningTime`

ndv99 (approved):
LGTM

PR #5482, opened 2024-06-24
docs: update docs link

petermakowski:
## Done
- docs: update docs link
- test: remove heading docs test

After this change we only ensure that the page itself displays as expected and don't test specific heading id.

Ever since we display a heading for MAAS variant (UI/CLI) headings change very often and it's not feasible to keep these links updated.

## QA steps

- [ ] 

## Fixes

Fixes:


ndv99 (approved):
LGTM. Suppose it doesn't make sense to check for headings any more if they're this volatile

PR #5488, opened 2024-06-26
test: add invalid CSRF token test case

petermakowski:
## Done
- Added test case for invalid CSRF token scenario in `base.test.ts`


tmerten (approved):
Thanks. Nice addition

PR #5499, opened 2024-07-15
feat(subnets): add tooltips for reserved IP comments MAASENG-3517

ndv99:
## Done
- Added tooltips to the "comments" column of the static DHCP leases table

## QA steps

You'll need a backend that has the websocket model for reserved IPs - I have one up and running, ping me if you want to use it

- [ ] Go to `subnet/{id}/address-reservation`
- [ ] Reserve an IP address and enter a long comment (at least 60 characters)
- [ ] Submit the form
- [ ] Hover over the comment in the table, and ensure a tooltip appears with the full text of the comment

## Fixes

Fixes [MAASENG-3517](https://warthogs.atlassian.net/browse/MAASENG-3517)

## Screenshots

image

## Notes

[MAASENG-3517]: https://warthogs.atlassian.net/browse/MAASENG-3517?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/app/subnets/views/SubnetDetails/StaticDHCPLease/StaticDHCPTable/StaticDHCPTable.tsx
+++ b/src/app/subnets/views/SubnetDetails/StaticDHCPLease/StaticDHCPTable/StaticDHCPTable.tsx
@@ -53,7 +53,7 @@ const generateRows = (
             ? getNodeTypeDisplay(reservedIp.node_summary.node_type)
             : "—"}
         </td>
-        <td>{reservedIp.comment || "—"}</td>
+        <td title={reservedIp.comment}>{reservedIp.comment || "—"}</td>
         <td>
           <TableActions
             onDelete={() =>
--- a/src/app/subnets/views/SubnetDetails/StaticDHCPLease/StaticDHCPTable/StaticDHCPTable.tsx
+++ b/src/app/subnets/views/SubnetDetails/StaticDHCPLease/StaticDHCPTable/StaticDHCPTable.tsx
@@ -53,7 +53,9 @@ const generateRows = (
             ? getNodeTypeDisplay(reservedIp.node_summary.node_type)
             : "—"}
         </td>
-        <td title={reservedIp.comment}>{reservedIp.comment || "—"}</td>
+        <td className="u-no-wrap" title={reservedIp.comment}>
+          {reservedIp.comment || "—"}
+        </td>
         <td>
           <TableActions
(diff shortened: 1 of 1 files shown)


tmerten:
Looking at this behavior it feels wrong that comments with spaces (words) are not trucated but comments with a single long string are. 
Other options would be to force a line break for long strings and/or to truncate in both cases and always show a pop up.
WDYT @dgtlntv @amylily1011 ?

dgtlntv:
> Looking at this behavior it feels wrong that comments with spaces (words) are not trucated but comments with a single long string are. Other options would be to force a line break for long strings and/or to truncate in both cases and always show a pop up. WDYT @dgtlntv @amylily1011 ?

I agree. I do wonder though how the truncated Lorem ipsum comment will look like in a title tooltip. @ndv99 Could you create a screenshot of that? I wonder if we might need to use Vanilla (not Vanilla html ^^) tooltips for truncated comments 🤔 Would that be annoying to implement (showing vanilla tooltips only if the comment is truncated)?

ndv99:
> > Looking at this behavior it feels wrong that comments with spaces (words) are not trucated but comments with a single long string are. Other options would be to force a line break for long strings and/or to truncate in both cases and always show a pop up. WDYT @dgtlntv @amylily1011 ?
> 
> I agree. I do wonder though how the truncated Lorem ipsum comment will look like in a title tooltip. @ndv99 Could you create a screenshot of that? I wonder if we might need to use Vanilla (not Vanilla html ^^) tooltips for truncated comments 🤔 Would that be annoying to implement (showing vanilla tooltips only if the comment is truncated)?

Vanilla tooltips were actually the very first thing I tried, but I remember us having problems using them in tables in the past (which resurfaced when I attempted this), which is why we ended up using `title`s everywhere

This is what the `title` for the lorem comment looks like:

image

dgtlntv:
I see! I think the title tooltip is fine. I thought it might not wrap nicely, but i think that is good enough. Thanks! :)

tmerten (approved):
Nice and consistent solution, thanks for updating.

ndv99:
> Nice and consistent solution, thanks for updating.

Thanks for the approval @tmerten, could you set the labels to code +1 and review +1, and remove the "review needed" ones?

PR #5510, opened 2024-07-29
fix: Prevent query client from fetching zones excessively on page load

ndv99:
## Done
- Removed `useEffect` in `useWebsocketAwareQuery` that was causing the zones endpoint to be fetched 50 times on page load
- Added TODO comment with a linked jira ticket to investigate this in future and determine if this code is needed

## QA steps

- [ ] Go to the UI and open dev tools
- [ ] Fully refresh the page (ctrl + F5), and ensure the zones endpoint is only called once
- [ ] Go to zones ("AZs"), and create/update/delete a zone
- [ ] Ensure that the zones list is automatically updated after each of these operations, *without* needing a page refresh

## Screenshots

This is how the network tools looked on initial page load for labmaas:
image

### DIFF
--- a/src/app/api/query/base.ts
+++ b/src/app/api/query/base.ts
@@ -2,11 +2,9 @@ import { useEffect, useCallback, useContext } from "react";
 
 import type { QueryFunction, UseQueryOptions } from "@tanstack/react-query";
 import { useQuery, useQueryClient } from "@tanstack/react-query";
-import { useSelector } from "react-redux";
 
 import type { QueryKey } from "@/app/api/query-client";
 import { WebSocketContext } from "@/app/base/websocket-context";
-import statusSelectors from "@/app/store/status/selectors";
 import type { WebSocketEndpointModel } from "@/websocket-client";
 import { WebSocketMessageType } from "@/websocket-client";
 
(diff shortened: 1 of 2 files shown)


src/app/api/query/base.ts
```diff
@@ -52,14 +50,16 @@ export function useWebsocketAwareQuery<
...
-  useEffect(() => {
-    queryClient.invalidateQueries();
-  }, [connectedCount, queryClient, queryKey]);
+  // TODO: Investigate why the code below causes the zones endpoint to be polled 50 times on page load
+  // https://warthogs.atlassian.net/browse/MAASENG-3558
+
+  // const connectedCount = useSelector(statusSelectors.connectedCount);
+  // useEffect(() => {
+  //   queryClient.invalidateQueries();
+  // }, [connectedCount, queryClient, queryKey]);
```
huwshimi, line 62: It looks like this might be happening because a new `queryKey` array reference is passed in on each render even though the array is always `["zones"]` and useEffect reruns when a reference is changed instead of checking the array's contents.

You could probably fix this bug by either storing the query key in a way that means the reference doesn't change (e.g. as a variable outside the hooks in zones.ts or inside a ref) or by checking whether the queryKey content changes and updating a ref inside this hook.

ndv99:
Closed, as this has been superseded by https://github.com/canonical/maas-ui/pull/5518

PR #5518, opened 2024-08-19
fix(zones): unnecessary query invalidations and API calls

petermakowski:
## Done
- fix(zones): unnecessary query invalidations and API calls
  - refactor: invalidate queries only when connectedCount changes
  - refactor: add zoneKeys

## QA steps

- [ ] Go to AZs page and ensure zones list is only fetched once

## Fixes

Fixes:


ndv99 (approved):
LGTM - welcome back!

PR #5521, opened 2024-08-20
docs: Add note to uninstall MAAS if present before doing LXD setup

ndv99:
## Done
- Added note to `LXD.md` to make sure MAAS is not installed locally before proceeding

### DIFF
--- a/docs/LXD.md
+++ b/docs/LXD.md
@@ -7,6 +7,8 @@ Having a locally running MAAS where you can deploy your own virtual machines can
 - Hardware that meets the [minimum requirements](https://maas.io/docs/reference-installation-requirements)
 - Knowledge of Ubuntu and basic Linux commands
 
+Note: If you've previously installed MAAS locally, make sure you uninstall it completely, or the LXD setup will not work.
+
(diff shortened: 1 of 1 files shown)


docs/LXD.md
```diff
@@ -7,6 +7,8 @@ Having a locally running MAAS where you can deploy your own virtual machines can
 - Hardware that meets the [minimum requirements](https://maas.io/docs/reference-installation-requirements)
 - Knowledge of Ubuntu and basic Linux commands
 
+Note: If you've previously installed MAAS locally, make sure you uninstall it completely, or the LXD setup will not work.
```
amylily1011, line 10: You can also do `sudo snap stop maas` (if you install maas from snap) to stop maas from running. Then once LXD is set up you can do `sudo snap start maas`, as well. I'm not sure if uninstalling maas completely is applicable for deb and probably a good way to start.

amylily1011 (commented):
I tried running 
```sh
lxc network create maas-kvm
lxc network edit maas-kvm
```

but it didn't work without `sudo`.

PR #5531, opened 2024-09-04
chore(deps): Upgrade @playwright/test to 1.45

ndv99:
## Done
- Upgraded playwright to v1.45

## Notes

This is needed in order for Playwight to be able to install its dependencies on 24.04. See issue thread [here](https://github.com/microsoft/playwright/issues/30368)


tmerten (approved):
lgtm, minor version bump

PR #5560, opened 2024-12-06
fix(machines): Allow ports at the end of IPMI addresses LP#2087965

ndv99:
## Done
- Added extra validation steps to IP address power field to account for ports
- Removed IP validation from LXD and Virsh address fields as these can be hostnames or other non-IP-address things.

## QA steps

- [ ] Go to a machine with an IPMI power type
- [ ] Go to "Configuration" and edit the power parameters
- [ ] Enter an IPv4 address with a port and tab out of the field afterwards
- [ ] Ensure no errors are shown
- [ ] Enter an invalid IP address and tab out of the field afterwards
- [ ] Ensure an error is shown
- [ ] Enter an IPv6 address with a port (you'll need to wrap the address in square brackets, e.g. `[2008:d45::1]:8000`
- [ ] Tab out of the field
- [ ] Ensure no errors are shown

## Fixes

Fixes [LP#2087965](https://bugs.launchpad.net/maas/+bug/2087965)
Fixes [MAASENG-4172](https://warthogs.atlassian.net/browse/MAASENG-4172)

[MAASENG-4172]: https://warthogs.atlassian.net/browse/MAASENG-4172?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/app/store/general/utils/powerTypes.ts
+++ b/src/app/store/general/utils/powerTypes.ts
@@ -1,4 +1,4 @@
-import { isIP } from "is-ip";
+import { isIP, isIPv6 } from "is-ip";
 import * as Yup from "yup";
 import type { ObjectShape } from "yup/lib/object";
 
@@ -56,12 +56,32 @@ const getPowerFieldSchema = (fieldType: PowerFieldType) => {
     case PowerFieldType.MULTIPLE_CHOICE:
       return Yup.array().of(Yup.string());
     case PowerFieldType.IP_ADDRESS:
-    case PowerFieldType.VIRSH_ADDRESS:
-    case PowerFieldType.LXD_ADDRESS:
       return Yup.string().test({
         name: "is-ip-address",
         message: "Please enter a valid IP address.",
-        test: (value) => isIP(value as string),
+        test: (value) => {
+          if (typeof value !== "string") {
+            return false;
+          }
+          if (value.includes("[") && value.includes("]")) {
+            // This is an IPv6 address with a port number
+            const openingBracketIndex = value.indexOf("[");
+            const closingBracketIndex = value.indexOf("]");
+            const ip = value.slice(
+              openingBracketIndex + 1,
+              closingBracketIndex
+            );
+            // We use +2 here to include the `:` before the port number
+            const port = value.slice(closingBracketIndex + 2);
+            return isIPv6(ip) && !isNaN(parseInt(port));
+          } else if (value.split(":").length === 2) {
+            // This is an IPv4 address with a port number
+            const [ip, port] = value.split(":");
+            return isIP(ip) && !isNaN(parseInt(port));
+          } else {
+            return isIP(value);
+          }
+        },
       });
     default:
       return Yup.string();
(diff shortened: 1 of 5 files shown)


r00ta (approved):
LGTM (up to my knowledge of react/ts :) )

src/app/store/general/utils/powerTypes.ts
```diff
@@ -56,12 +56,32 @@ const getPowerFieldSchema = (fieldType: PowerFieldType) => {
...
+            // This is an IPv6 address with a port number
+            const openingBracketIndex = value.indexOf("[");
+            const closingBracketIndex = value.indexOf("]");
+            const ip = value.slice(
+              openingBracketIndex + 1,
+              closingBracketIndex
+            );
+            // We use +2 here to include the `:` before the port number
+            const port = value.slice(closingBracketIndex + 2);
+            return isIPv6(ip) && !isNaN(parseInt(port));
```
tmerten, line 76: This usually works but insane users like myself might try: `[aa:cf:f::34fF]:  4 5 6`. Results in a `parseInt("  4 5 6")` which is `4`. I think it's still OK to merge for now but we might want to make this even tighter later.

src/app/store/general/utils/powerTypes.ts
```diff
@@ -56,12 +56,32 @@ const getPowerFieldSchema = (fieldType: PowerFieldType) => {
...
+          } else if (value.split(":").length === 2) {
+            // This is an IPv4 address with a port number
+            const [ip, port] = value.split(":");
+            return isIP(ip) && !isNaN(parseInt(port));
```
tmerten, line 80: as above. Could error out if the port number has anything but decimals.

tmerten (commented):
QA: Looking good, changed a machine, tried multiple V4 and V6 addresses with and without port.

Code: small edge case nag inline below

src/app/store/general/utils/powerTypes.ts
ndv99: That's interesting! By the looks of it, `parseInt` ignores any leading whitespace, and ignores anything once it reaches the first whitespace after a number (only if the number is also the first non-whitespace text in the string).

I guess the simplest check for this would just be to check for any whitespace in the input, and treat the value as invalid if there is whitespace.

src/app/utils/isValidPortNumber.test.ts
```diff
@@ -0,0 +1,15 @@
+import { isValidPortNumber } from ".";
+
+it("returns true for any number between 0 and 65535", () => {
+  for (let i = 0; i <= 65535; i++) {
```
tmerten, line 4: That is a computationally expensive test for the function it tests.... Edge cases and some common ports might be enough. But I'm happy to merge if it runs quickly.

src/app/utils/isValidPortNumber.ts
```diff
@@ -0,0 +1,9 @@
+/**
+ * Checks if a given port number is valid (between 0 and 65535).
+ *
+ * @param port The port number to check.
+ * @returns True if valid, false otherwise.
+ */
+export const isValidPortNumber = (port: number): boolean => {
```
tmerten, line 7: I like that its `port: number` :)

tmerten (approved):
I really wanted to break it. LTGM.

src/app/utils/isValidPortNumber.test.ts
ndv99: Locally it seems to run quickly enough, but thought I'd get your thoughts first.

src/app/utils/isValidPortNumber.ts
ndv99: me too :)

PR #5576, opened 2025-01-13
feat: Add plugin for custom naming API methods MAASENG-4270

abuyukyi101198:
## Done

- Added a custom plugin that can rename/reparse the auto-generated SDK method names from the URL
- Generated types and SDK with custom naming

## Fixes

Fixes: 

Resolves [MAASENG-4270](https://warthogs.atlassian.net/browse/MAASENG-4270)

[MAASENG-4270]: https://warthogs.atlassian.net/browse/MAASENG-4270?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/app/apiclient/plugin/plugin.ts
```diff
@@ -0,0 +1,24 @@
...
+    Object.keys(paths).forEach((path: string) => {
+      const pathMethods: IR.PathItemObject =
+        paths[path as keyof IR.PathsObject];
+
+      Object.keys(pathMethods).forEach((method: string) => {
+        const methodData: IR.OperationObject = pathMethods[
+          method as keyof IR.PathItemObject
+        ] as IR.OperationObject;
+
+        methodData.id = methodData.id.split("Maas")[0];
```
tmerten, line 20: This looks like it should work. I'm wondering if we can make it more robust (and generic) by stripping off the [client's baseUrl](https://heyapi.dev/openapi-ts/clients/fetch#configuration)?

abuyukyi101198: The default SDK names (i.e. `id`s) are in `<summary><path><method>` format. This is what an example `IR.OperationObject` looks like:

```
put: {
    id: 'updateZoneMaasAV3ZonesZoneIdPut',
    method: 'put',
    path: '/MAAS/a/v3/zones/{zone_id}',
    summary: 'Update Zone',
    tags: [ 'Zones' ],
    parameters: { path: [Object] },
    body: {
      mediaType: 'application/json',
      schema: [Object],
      required: true,
      type: 'json'
    },
    responses: { '200': [Object], '404': [Object], '422': [Object] },
    security: [ [Object] ]
  }
```

Adding a base url did not change the naming convention. The `path` seems to be taken from somewhere else.

abuyukyi101198: What we can do is pass the base url as a custom prop into our plugin and use it there, but I don't think it would change the generated code.

abuyukyi101198: Or we can remove the path of the operation from the name, which we already can access.

openapi-ts.config.ts
```diff
@@ -1,12 +1,13 @@
-import { defineConfig } from "@hey-api/openapi-ts";
+import { defineConfig as defineCustomPlugin } from "./src/app/apiclient/plugin";
 
-export default defineConfig({
+export default {
```
ndv99, line 3: I think you can still wrap this in `defineConfig`, ESLint is complaining locally that the default export hasn't been assigned to a variable

src/app/apiclient/plugin/plugin.ts
```diff
@@ -0,0 +1,38 @@
...
+        paths[path as keyof IR.PathsObject];
+
+      Object.keys(pathMethods).forEach((method: string) => {
+        const methodData: IR.OperationObject = pathMethods[
+          method as keyof IR.PathItemObject
+        ] as IR.OperationObject;
+
+        const pathName = path
+          .split("/")
+          .filter(Boolean)
```
ndv99, line 22: TIL you can do this to get rid of null/undefined in an array. Nice!

ndv99 (approved):
LGTM, one minor change below

openapi-ts.config.ts
abuyukyi101198: Done 👍

abuyukyi101198:
Since the OpenAPI specs now have sensible operationIds for naming the generated methods, this PR is no longer necessary.

PR #5590, opened 2025-01-23
feat(zones): Add delete zones as a table action MAASENG-4308

abuyukyi101198:
## Done

- Refactored DeleteConfirm to a side panel form named DeleteZone
- Added DELETE_ZONE to zone dispatch actions with custom param for zone id
- Added Actions column to zones table, added delete action
- Removed delete action from the zone details view

## QA steps

- [x] Verify the Actions column is visible on the zones table with delete action (trash can) available
- [x] Ensure the delete action is disabled for the default zone
- [x] Ensure clicking the delete action opens the side panel titled "Delete AZ"
- [x] Verify clicking Cancel on the side panel form closes the side panel with no other effects
- [x] Verify clicking Delete on the side panel form closes the side panel and then removed the deleted AZ row
- [x] Verify the Delete AZ button is no longer available in zone details

## Fixes

Resolves: 

[MAASENG-4308](https://warthogs.atlassian.net/browse/MAASENG-4308)

## Screenshots

Screenshot 2025-01-23 at 14 11 18
Screenshot 2025-01-23 at 14 20 04

## Notes

Removal of the deleted AZ row may not be fully responsive due to the lack of support for a notification for when deletion is successful. APIv3 will resolve this problem.

Tests were removed either because they were no longer applicable to this new layout or they will have to completely overhauled with v3 anyway.

[MAASENG-4308]: https://warthogs.atlassian.net/browse/MAASENG-4308?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/app/zones/views/ZonesList/ZonesList.tsx
```diff
@@ -6,9 +8,11 @@ import { useZoneCount } from "@/app/api/query/zones";
 import PageContent from "@/app/base/components/PageContent";
 import { useWindowTitle } from "@/app/base/hooks";
 import { useSidePanel } from "@/app/base/side-panel-context";
+import { getSidePanelTitle } from "@/app/store/utils/node/base";
 import { ZoneActionSidePanelViews } from "@/app/zones/constants";
+import DeleteZone from "@/app/zones/views/ZonesList/ZonesListTable/DeleteZone";
 
-const ZonesList = (): JSX.Element => {
+const ZonesList: React.FC = () => {
```
ndv99, line 15: I've noticed you doing this change with a few components, what's the reasoning? (I honestly don't know which is the better way to type a component)

ndv99 (approved):
LGTM, nice one!

src/app/zones/views/ZonesList/ZonesList.tsx
tmerten: I'm a bit late to the party. In general from what I see `React.FC` seems to be fine.  However, FB [removed it from the default template](https://github.com/facebook/create-react-app/pull/8177) and discussed this move in the PR. TL;DR: With `FC` we buy an explicit return type but we also get a contract that allows child components even if we're not prepared for them. 
I'll add a discussion note for the next sync meeting. Might also be something for the React Guild.

PR #5591, opened 2025-01-23
fix: add Origin header in the proxy so to bypass django CSRF protections

r00ta:
## Done

Fix the e2e tests that are failing due to the new CSRF protections introduced in Django 4.x. We simply add in the proxy the fake Origin header. 

## QA steps

- deploy MAAS on an LXD container
- run the UI with `yarn start` on the host and point to the LXD container for the MAAS_URL
- try to reproduce the bug


ndv99 (approved):
LGTM, you're a legend for fixing this :heart:

PR #5600, opened 2025-02-03
fix: coverage on schedule should analyze main branch

tmerten:
Branch name needs to be explicit as TiCS defaults to `master` and we run on `main`.

## QA steps

Hard to do. Push this to `ci-tics-main` to test if pushing to a `ref` still works as expected. Otherwise need to merge and wait for the next scheduled run on Saturday.

## Fixes

- TiCS analyze step on scheduled runs
- Should also still work running on push to a branch (can be tested by pushing this commit to `ci-tics-main`)


tmerten:
fyi, variables are described here: https://docs.github.com/en/actions/writing-workflows/choosing-what-your-workflow-does/accessing-contextual-information-about-workflow-runs#github-context

abuyukyi101198 (approved):

PR #5605, opened 2025-02-05
ci: add 'deps-dev' as a valid commit scope

ndv99:
## Done
- Added "deps-dev" as a valid commit scope to help out dependabot


abuyukyi101198 (approved):
LGTM 👍

PR #5623, opened 2025-02-26
chore(deps): Update MAAS UI Redux ecosystem MAASENG-4482

abuyukyi101198:
## Done

- Upgraded the packages as outlined in https://github.com/canonical/maas-ui/pull/5610
- Fixed the `getDefaultMiddleware` deprecation error by updating the `redux-store` middleware definition
- Added type casts for common reducers
- Removed generics affecting Redux' type inference
- Removed deprecated `ParametricSelector` and `OutputParametricSelector`

## Fixes

Resolves: 

[MAASENG-4482](https://warthogs.atlassian.net/browse/MAASENG-4482)

[MAASENG-4482]: https://warthogs.atlassian.net/browse/MAASENG-4482?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/app/machines/components/MachineForms/MachineActionFormWrapper/MachineActionFormWrapper.tsx
```diff
@@ -43,7 +43,7 @@ type ContainerProps = Omit<MachineActionFormProps, "processingCount"> & {
 
 type Props = ContainerProps & {
   clearSelectedMachines: () => void;
-  dispatch: Dispatch<AnyAction>;
+  dispatch: Dispatch<Action>;
```
ndv99, line 46: Ideally we should use `UnkownAction` here instead of `Action`
https://redux-toolkit.js.org/usage/migrating-rtk-2#anyaction-deprecated-in-favour-of-unknownaction

ndv99 (changes requested):
One comment inline, but my biggest concern is the type casting to `CaseReducer` - casting types for edge cases is fine IMO, but this is some pretty core stuff. I would first see if we can get around this by updating the return/parameter types of `generateStatusHandlers`, since all of the affected handlers seem to be status handlers. 

If this isn't possible, I'm okay to leave it as-is and merge, but heavy type-casting like this really makes me nervous :sweat_smile:

src/app/machines/components/MachineForms/MachineActionFormWrapper/MachineActionFormWrapper.tsx
abuyukyi101198: `UnknownAction` is not assignable to the type of `MachineActionForm` `dispatch` prop, even if you update the `Props` `dispatch` to accept `Dispatch<UnknownAction>`. So I have to keep it as `Action`.

abuyukyi101198:
> One comment inline, but my biggest concern is the type casting to `CaseReducer` - casting types for edge cases is fine IMO, but this is some pretty core stuff. I would first see if we can get around this by updating the return/parameter types of `generateStatusHandlers`, since all of the affected handlers seem to be status handlers.
> 
> If this isn't possible, I'm okay to leave it as-is and merge, but heavy type-casting like this really makes me nervous 😅

I definitely tried setting the return type of the `generateStatusHandlers` to provide those `CaseReducer`s, instead of adding casts by hand, but that change of types seems unreconcilable with the current return type `{ [x: string]: SliceCaseReducers<S> }`. There is `CaseReducers` (plural) type that could have done the job, but it is deprecated.

ndv99:
> > One comment inline, but my biggest concern is the type casting to `CaseReducer` - casting types for edge cases is fine IMO, but this is some pretty core stuff. I would first see if we can get around this by updating the return/parameter types of `generateStatusHandlers`, since all of the affected handlers seem to be status handlers.
> > If this isn't possible, I'm okay to leave it as-is and merge, but heavy type-casting like this really makes me nervous 😅
> 
> I definitely tried setting the return type of the `generateStatusHandlers` to provide those `CaseReducer`s, instead of adding casts by hand, but that change of types seems unreconcilable with the current return type `{ [x: string]: SliceCaseReducers<S> }`. There is `CaseReducers` (plural) type that could have done the job, but it is deprecated.

That's understandable, thanks for the explanation. In that case, I think this is probably good to merge, especially given that we're aiming to remove the store in the future :+1:

PR #5626, opened 2025-02-27
feat(preferences): Migrate SSH key deletion to API v3 MAASENG-4449

ndv99:
## Done
- Migrated SSH key delete form to API v3
- Added missing tests for ssh key query hooks
- Delete form now fetches IDs from side panel context instead of URL parameters
- Created mock resolver for SSH key deletion
- Added generic type props to ModelActionForm to support v3 API errors

## QA steps

- [x] Go to /account/prefs/ssh-keys
- [x] If you don't have any keys, upload/import some
- [x] Delete a single SSH key 
- [x] Ensure the request succeeds and the list is updated
- [ ] Delete a group of SSH keys
- [ ] Ensure the request succeeds and the list is updated

## Fixes

Resolves [MAASENG-4449](https://warthogs.atlassian.net/browse/MAASENG-4449)

[MAASENG-4449]: https://warthogs.atlassian.net/browse/MAASENG-4449?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/app/preferences/views/SSHKeys/DeleteSSHKey/DeleteSSHKey.tsx
+++ b/src/app/preferences/views/SSHKeys/DeleteSSHKey/DeleteSSHKey.tsx
@@ -1,48 +1,41 @@
-import { useState } from "react";
-
 import { useOnEscapePressed } from "@canonical/react-components";
-import { useDispatch, useSelector } from "react-redux";
-import { useNavigate, useSearchParams } from "react-router-dom";
+import { useNavigate } from "react-router-dom";
 
+import { useDeleteSshKey } from "@/app/api/query/sshKeys";
+import type { DeleteUserSshkeyError } from "@/app/apiclient";
 import ModelActionForm from "@/app/base/components/ModelActionForm";
-import { useAddMessage } from "@/app/base/hooks";
+import { useSidePanel } from "@/app/base/side-panel-context";
+import type { EmptyObject } from "@/app/base/types";
 import urls from "@/app/preferences/urls";
-import { sshkeyActions } from "@/app/store/sshkey";
-import sshkeySelectors from "@/app/store/sshkey/selectors";
-import type { SSHKey, SSHKeyMeta } from "@/app/store/sshkey/types";
 
 const DeleteSSHKey = () => {
-  const [deleting, setDeleting] = useState<SSHKey[SSHKeyMeta.PK][]>([]);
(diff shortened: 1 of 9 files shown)


src/app/preferences/views/SSHKeys/DeleteSSHKey/DeleteSSHKey.tsx
```diff
@@ -1,48 +1,41 @@
...
+  const { sidePanelContent } = useSidePanel();
+
+  const getIds = () => {
+    const extras = sidePanelContent?.extras;
+    if (extras && "group" in extras && extras.group) {
+      return extras.group.keys.map((key) => key.id);
+    } else {
+      return [];
+    }
+  };
```
abuyukyi101198, line 27: This function seems to expect the ids to arrive in the sidePanel context, however, the list item delete button does not use that for ids

abuyukyi101198: Here is what the onDelete does in SSHKeyList:
```ts
onClick={() =>
                navigate({
                  pathname: urls.sshKeys.delete,
                  search: `?ids=${ids.join()}`,
                })
              }
```

That's a weird way of triggering the form, and I guess it should change.

abuyukyi101198 (commented):
I ran into an issue with the delete form rendering, here is what it looks like:
Screenshot 2025-02-27 at 14 11 31

src/app/preferences/views/SSHKeys/DeleteSSHKey/DeleteSSHKey.tsx
ndv99: Thanks for raising this - I agree, it's definitely weird. Every other side panel that I know of in the UI is triggered by setting `sidePanelContent` in context - see `src/app/subnets/views/FormActions/FormActions.tsx` and `src/app/subnets/views/SubnetsList/SubnetsList.tsx` for a good example of this.

IMO we should refactor the whole `preferences` route at some point to behave like this. For now, I'm reverting this form to pull this information from the URL again, and using `renderWithBrowserRouter` to test it for now.

abuyukyi101198 (approved):
LGTM!

PR #5628, opened 2025-02-27
feat(images): Implement design for image table MAASENG-4360

rumanowskak:
## Done
- Added tooltip for delete icon in Images Table
- Aligned header text starting from "Release title" to the text "Ubuntu" in header belowe

## QA steps

- [x] On MAAS-UI go to configuration/images
- [x] Hover over delete icon to see the tooltip
- [x] Look at header if "Release title" is aligned with the "Ubuntu" text

## Fixes

Resolves [MAASENG-4360](https://warthogs.atlassian.net/browse/MAASENG-4360)

## Screenshots

images_table

[MAASENG-4360]: https://warthogs.atlassian.net/browse/MAASENG-4360?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/app/base/components/GenericTable/_index.scss
+++ b/src/app/base/components/GenericTable/_index.scss
@@ -7,15 +7,12 @@
   }
 
   .group-select,
-  .select {
+  .select,
+  .select-alignment {
     width: 2rem;
   }
 
   th.select {
     display: none;
-  }
-
-  th:not(.release):not(.group-select) {
-    padding-left: 2rem;
   }
 }
--- a/src/app/base/components/GenericTable/_index.scss
+++ b/src/app/base/components/GenericTable/_index.scss
@@ -14,4 +14,8 @@
   th.select {
     display: none;
   }
+
+  th:not(.release):not(.group-select) {
+    padding-left: 2rem;
+  }
 }
--- a/src/app/base/components/GenericTable/_index.scss
+++ b/src/app/base/components/GenericTable/_index.scss
@@ -1,10 +1,17 @@
 .p-generic-table {
-  thead, tbody {
+
+  thead,
+  tbody {
     display: table-row-group;
     width: 100%;
   }
 
-  .group-select, .select {
+  .group-select,
+  .select {
     width: 2rem;
+  }
+
+  th.select {
+    display: none;
   }
 }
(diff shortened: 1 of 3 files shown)


abuyukyi101198 (changes requested):
Looks good! However, there is an easy miss in the design, where when aligning the **"Release title"** to the left, the other column headers chould not move, as in all column headers should be left-aligned properly.

Screenshot 2025-02-27 at 17 34 38

Managing to shift the second header as the same amount as the width of the now `display: none` `.select` should align all headers correctly.

rumanowskak:
I tried to just shift the second header "Architecture" but it didn't shift the rest of them, so I did it this way.

src/app/base/components/GenericTable/_index.scss
```diff
@@ -1,10 +1,21 @@
...
     width: 2rem;
   }
+
+  th.select {
+    display: none;
+  }
+
+  th:not(.release):not(.group-select) {
+    padding-left: 2rem;
+  }
```
abuyukyi101198, line 20: Since this is a generic component, we cannot directly refer to image-specific classes like `.release`. It seems that the alignment is not solvable with pure CSS. Maybe we can add an empty `th` element after the first visible header programmatically. I'm not a huge fan of this solution, but CSS just doesn't seem to work.

In `GenericTable.tsx`:
```ts
<DynamicTable className="p-generic-table" variant={variant}>
      <thead>
        {table.getHeaderGroups().map((headerGroup) => (
          <tr key={headerGroup.id}>
            {headerGroup.headers.filter(filterHeaders).map((header, index) => (
              <>
                <TableHeader header={header} key={header.id} />
                {index === 2 ? <th className="select-alignment" /> : null} <-- add this aligner th
              </>
            ))}
          </tr>
        ))}
      </thead>
```

And in `GenricTable/_index.scss`:
```scss
.group-select,
  .select,
  .select-alignment { <-- add select-alignment
    width: 2rem;
  }
```

abuyukyi101198 (changes requested):
The alignment seems to shift in later headers. CSS may not be enough to fix the alignment.

abuyukyi101198 (approved):
LGTM!

PR #5630, opened 2025-02-28
chore(store): Remove zones store MAASENG-4504

abuyukyi101198:
## Done

- Removed `store/zones`
- Replaced store zone types with `apiclient` types
- Removed zone in mock stores
- (drive-by) Replaced `renderWithBrowserRouter` with `renderWithProviders` where drop-in replacement was possible

## Fixes

Resolves: 

[MAASENG-4504](https://warthogs.atlassian.net/browse/MAASENG-4504)

## Notes

`renderWithBrowserRouter` uses cases with `queryData` parameters being passed in, and `store` expected in return are NOT drop-in replacable with `renderWithProviders`.

[MAASENG-4504]: https://warthogs.atlassian.net/browse/MAASENG-4504?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/app/api/query/zones.test.ts
+++ b/src/app/api/query/zones.test.ts
@@ -1,5 +1,3 @@
-import { describe } from "vitest";
-
 import {
   useZones,
   useZoneCount,
@@ -16,7 +14,7 @@ import {
   waitFor,
 } from "@/testing/utils";
 
-setupMockServer(
+const mockServer = setupMockServer(
   zoneResolvers.listZones.handler(),
   zoneResolvers.getZone.handler(),
   zoneResolvers.createZone.handler(),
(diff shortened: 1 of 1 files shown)


src/__snapshots__/root-reducer.test.ts.snap
ndv99, line 1: Just before I do the full review - I actually deleted this snapshot entirely in #5629, and re-wrote the test to account for that, could maybe do that here first if this will land first

src/app/base/components/ZoneSelect/ZoneSelect.test.tsx
```diff
@@ -30,18 +26,10 @@ describe("ZoneSelect", () => {
...
   it("disables select if zones have not loaded", () => {
-    const state = factory.rootState({
-      zone: factory.zoneState({
-        genericActions: factory.zoneGenericActions({ fetch: "idle" }),
-      }),
-    });
-    const store = mockStore(state);
-
-    renderWithMockStore(
+    renderWithProviders(
```
ndv99, line 29: IMO, you should use `mockIsPending()` before rendering the component, as that will force useQuery `isPending` to be true and simulate loading

ndv99 (approved):
LGTM for the most part, one nitpick inline

PR #5683, opened 2025-04-17
feat(base): Add sticky header to GenericTable MAASENG-4718

abuyukyi101198:
## Done

- Made table body scrollable
- Removed `DynamicTable` import, incorporated fully
- Added dynamic height sizing
- Added loading state and skeleton rows
- Refactored code, added accessibility attributes

## QA steps

- [x] Navigate to `images/`
- [x] Verify loading skeleton rows are displayed when polling data
- [x] Verify the grouping and selection checkboxes work as intended
- [x] Navigate to `pools/`
- [x] Verify loading skeleton rows are displayed when querying data
- [x] Verify sorting works as intended 
- [x] Ensure pagination works properly
- [x] Verify the table height is fit into the main section such that the window is not scrollable
- [x] Verify the table body (rows) are scrollable, and the table header and pagination stay fixed during scrolling

## Fixes

Resolves: 

[MAASENG-4718](https://warthogs.atlassian.net/browse/MAASENG-4718)

## Screenshots

Screenshot 2025-04-17 at 14 53 37

## Notes

With this PR, GenericTable can be evaluated for upstreaming to https://github.com/canonical/maas-react-components

[MAASENG-4718]: https://warthogs.atlassian.net/browse/MAASENG-4718?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/app/base/components/GenericTable/GenericTable.tsx
+++ b/src/app/base/components/GenericTable/GenericTable.tsx
@@ -1,12 +1,4 @@
-import {
-  Fragment,
-  useCallback,
-  useEffect,
-  useLayoutEffect,
-  useMemo,
-  useRef,
-  useState,
-} from "react";
+import { Fragment, useEffect, useMemo, useRef, useState } from "react";
 import type { Dispatch, ReactNode, SetStateAction } from "react";
 
 import type {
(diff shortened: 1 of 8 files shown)


src/app/base/components/GenericTable/GenericTable.tsx
```diff
@@ -45,60 +56,176 @@ type GenericTableProps<T extends { id: string | number }> = {
...
+  const [maxHeight, setMaxHeight] = useState("auto");
+
   const [grouping, setGrouping] = useState<GroupingState>(groupBy ?? []);
   const [expanded, setExpanded] = useState<ExpandedState>(true);
   const [sorting, setSorting] = useState<SortingState>(sortBy ?? []);
 
-  if (canSelect) {
-    columns = [
+  // Update table height based on available space
+  useEffect(() => {
```
ndv99, line 132: Since this effect is taking measurements of the DOM and making layout changes based on that, you might want to use `useLayoutEffect` here for the most stable results. Totally optional, since it appears to work fine anyway, just a thought

src/app/base/components/GenericTable/GenericTable.tsx
```diff
@@ -45,60 +56,176 @@ type GenericTableProps<T extends { id: string | number }> = {
...
+ *   }}
+ * />
+ */
 const GenericTable = <T extends { id: string | number }>({
+  className,
   canSelect = false,
-  columns,
-  data,
+  columns: initialColumns,
+  containerRef,
```
ndv99, line 110: Is it definitely necessary to pass a ref through here? The old implementation from DynamicTable didn't need an external ref for the height calculations: https://github.com/canonical/maas-react-components/blob/main/src/lib/components/DynamicTable/DynamicTable.tsx#L116-L126

ndv99 (commented):
This is looking good for the most part - there's a couple of questions inline, and also something I noticed on the images page during QA:

image

There's these little dots in the skeleton rows that only appear on this table, not on Pools

src/app/base/components/GenericTable/GenericTable.tsx
abuyukyi101198: It is an optional prop. The issue with `DynamicTable`, which became apparent in MAAS UI's layout, is that it assumes there are no absolutely-positioned components below the table. The existence of the `StatusBar` component as an `<aside>` footer causes `DynamicTable's` way of sizing to be misled, and results in scrolling of the entire main content.

As it is now, `GenericTable` looks for the `<main>` element, assuming it is its parent, and uses it as a reference for resizing. The `containerRef` prop allows that reference to be swapped with any parent component, allowing for smaller tables to be placed in components other than the main section.

Of course, this resizing is only functional when the `variant` prop is set to `"full-height"`.

abuyukyi101198:
> This is looking good for the most part - there's a couple of questions inline, and also something I noticed on the images page during QA:
> 
> image
> 
> There's these little dots in the skeleton rows that only appear on this table, not on Pools

Yes, I noticed them when the window width is low and the `Placeholder`components get truncated. Since they are `<span>`elements with text inside, those ellipses appear. I couldn't make them disappear entirely through styling, and how we render the loading state is a necessary discussion, since the current `Placeholder` components are difficult to align, and since their animations aren't synced, they look quite crowded. Maybe a spinner instead of placeholder rows could be better.

ndv99 (approved):
LGTM - I think it's a bit of a shame to lose the placeholder/skeleton rows, but ultimately if the ellipses thing can't be fixed in CSS then a loading spinner will definitely suffice.

PR #5696, opened 2025-04-29
fix: sitespeed test

motrilense:
## Done
- Collect and upload MAAS logs on failure

## QA steps

- enable sitespeed in open PR
- verify that the fix works
- disable sitespeed in open PR

## Fixes

Fixes: 

- update the link to the new MAAS database dump (the previous version is incompatible with the current MAAS database schema)
- update the OS used in sitespeed to Ubuntu 24.04

### DIFF
--- a/.github/workflows/sitespeed.yml
+++ b/.github/workflows/sitespeed.yml
@@ -4,10 +4,12 @@ on:
     branches:
       - main
       - "3.*"
+  pull_request:
+    types: [opened, synchronize, reopened]
 jobs:
   run-sitespeed:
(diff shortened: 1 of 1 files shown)


tmerten (approved):
Thanks for adding the failure case 🥳

tmerten:
Although there is still an error creating the issue :(

✖  error     An error occurred while updating the issue. This might be caused by a malformed issue title, or a typo in the labels or assignees. Check .github/ISSUE_TEMPLATE/ci-failure.md!

PR #5735, opened 2025-06-16
fix(users): Fixes for the v3 API users migration

ndv99:
## Done
- Completing the user intro redirects to the machine list instead of doing nothing
- The `IncompleteCard` is no longer shown if the user is an admin
- `me_with_summary` is only called once before the login screen is shown, instead of retrying 3 times

## QA steps

With a freshly initialised MAAS* and a new admin user:

- [ ] Open dev tools
- [ ] Go to MAAS UI
- [ ] Ensure that only one request is made to `me_with_summary`
- [ ] Log in
- [ ] Ensure you are sent to the MAAS intro
- [ ] Complete the MAAS intro
- [ ] Ensure you are sent to the user intro
- [ ] Import SSH keys from any source
- [ ] Click "Finish setup"
- [ ] Ensure you are immediately redirected to the machine list

## Notes

\*If you are unable to spin up a clean MAAS instance, I can provide one as needed.

### DIFF
--- a/src/app/App.tsx
+++ b/src/app/App.tsx
@@ -13,6 +13,7 @@ import { Outlet } from "react-router";
 
 import packageInfo from "../../package.json";
 
+import { getMeWithSummaryQueryKey } from "./apiclient/@tanstack/react-query.gen";
 import NavigationBanner from "./base/components/AppSideNavigation/NavigationBanner";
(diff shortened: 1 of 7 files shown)


abuyukyi101198 (approved):
Looks good

PR #5744, opened 2025-06-23
feat(api): create query and mutation hooks for configurations

ndv99:
## Done
- Created query hooks for:
   - Get configurations (list)
   - Get configuration by ID
- Created mutation hooks for:
   - Set configuration by ID
   - Bulk set configurations
- Created mock resolvers for configurations


abuyukyi101198 (approved):
LGTM!

PR #5751, opened 2025-07-10
fix: remove broken/unsupported nova power driver

r00ta:
## Done

Removing the nova power driver from the frontend. 

On backend side this is coming with https://code.launchpad.net/~r00ta/maas/+git/maas/+merge/488664

## Fixes

Fixes: https://bugs.launchpad.net/maas/+bug/2084719


abuyukyi101198 (approved):
LGTM!

The Cypress test failure is a known CI issue we are investigating, so it's ok if it is failing.

PR #5770, opened 2025-08-25
feat: Generic side panel context MAASENG-5227

abuyukyi101198:
## Done

- Created 'side-panel-context-new.tsx*
- Migrated SSH and SSL key tables
- Added new provider to 'renderWithProviders'
- Removed no longer applicable tests

## QA steps

- [ ] Navigate to SSH keys
- [ ] Open any form
- [ ] Close the form
- [ ] Re-open the form, and switch to SSL keys
- [ ] Verify form is automatically closed
- [ ] Repeat for SSL keys
- [ ] Try auto-closing for old side panel context views e.g. open an SSH keys form, navigate to Pools and vice versa

## Fixes

Resolves:

[MAASENG-5227](https://warthogs.atlassian.net/browse/MAASENG-5227)

[MAASENG-5227]: https://warthogs.atlassian.net/browse/MAASENG-5227?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/app/base/side-panel-context-new.tsx
+++ b/src/app/base/side-panel-context-new.tsx
@@ -17,20 +17,55 @@ interface SidePanelState<TProps = Record<string, unknown>> {
 }
 
 interface SidePanelActions {
-  open: <TProps extends Record<string, unknown> = Record<string, unknown>>(
-    component: ComponentType<TProps>,
-    title: string,
-    props?: TProps,
-    size?: SidePanelSize
-  ) => void;
-  close: () => void;
-  setSize: (size: SidePanelSize) => void;
+  openSidePanel: <
+    TProps extends Record<string, unknown> = Record<string, unknown>,
+  >(params: {
+    component: ComponentType<TProps>;
+    title: string;
+    props?: TProps;
+    size?: SidePanelSize;
+  }) => void;
+  closeSidePanel: () => void;
+  setSidePanelSize: (size: SidePanelSize) => void;
 }
 
 type SidePanelContextValue = SidePanelActions & SidePanelState;
 
 const SidePanelContext = createContext<SidePanelContextValue | null>(null);
 
+/**
+ * Hook for managing side panel state and actions.
+ *
+ * Provides methods to open/close side panels with React components and manages
+ * panel size and visibility state.
+ *
+ * @returns Object containing:
+ *   - `isOpen`: Boolean indicating if the side panel is currently open
+ *   - `title`: Current panel title string
+ *   - `size`: Current panel size ('narrow' | 'regular' | 'wide' | 'large')
+ *   - `component`: Currently rendered component
+ *   - `props`: Props passed to the current component
+ *   - `open(component, title, props?, size?)`: Opens panel with given component
+ *   - `close()`: Closes the panel and resets state
+ *   - `setSize(size)`: Changes the panel width
+ *
+ * @throws Error when used outside SidePanelProvider
+ *
+ * @example
+ * ```tsx
+ * const { open, close, isOpen } = useSidePanel();
+ *
+ * // Open a panel with a form component
+ * const handleEdit = (userId: number) => {
+ *   open(EditUserForm, 'Edit User', { userId }, 'wide');
+ * };
+ *
+ * // Close the panel
+ * const handleCancel = () => {
+ *   close();
+ * };
+ * ```
+ */
 export const useSidePanel = (): SidePanelContextValue => {
   const context = useContext(SidePanelContext);
   if (!context) {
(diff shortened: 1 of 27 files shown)


src/app/base/side-panel-context-new.tsx
```diff
@@ -0,0 +1,99 @@
...
+  ) => void;
+  close: () => void;
+  setSize: (size: SidePanelSize) => void;
+}
+
+type SidePanelContextValue = SidePanelActions & SidePanelState;
+
+const SidePanelContext = createContext<SidePanelContextValue | null>(null);
+
+export const useSidePanel = (): SidePanelContextValue => {
```
ndv99, line 34: If you could add a docstring explaining the returned variables, that would be really awesome :+1:

src/app/base/side-panel-context-new.tsx
```diff
@@ -0,0 +1,99 @@
...
+      title,
+      size,
+      component: component as ComponentType<Record<string, unknown>>,
+      props: props || ({} as Record<string, unknown>),
+    });
+  };
+
+  const close = () => {
+    setState((prev) => ({
+      ...prev,
```
ndv99, line 72: Why do we need to spread the previous state here if we're overriding everything?

src/app/base/side-panel-context-new.tsx
```diff
@@ -0,0 +1,99 @@
...
+  });
+
+  const open = <
+    TProps extends Record<string, unknown> = Record<string, unknown>,
+  >(
+    component: ComponentType<TProps>,
+    title: string,
+    props: TProps = {} as TProps,
+    size: SidePanelSize = "regular"
+  ) => {
```
ndv99, line 60: We should use named parameters if a function has more than 2 parameters

```suggestion
  const open = <
    TProps extends Record<string, unknown> = Record<string, unknown>,
  >({ component, title, props = {} as TProps, size = "regular" }: {
    component: ComponentType<TProps>;
    title: string;
    props: TProps;
    size: SidePanelSize;
  }) => {
```

src/app/base/side-panel-context-new.tsx
```diff
@@ -0,0 +1,99 @@
...
+
+interface SidePanelActions {
+  open: <TProps extends Record<string, unknown> = Record<string, unknown>>(
+    component: ComponentType<TProps>,
+    title: string,
+    props?: TProps,
+    size?: SidePanelSize
+  ) => void;
+  close: () => void;
+  setSize: (size: SidePanelSize) => void;
```
ndv99, line 27: Just my opinion, but I think more descriptive names for returned variables would be nice, e.g. `openSidePanel`, `closeSidePanel`, `setSidePanelSize`. I know it's longer to type but I think it makes downstream code a bit more readable, especially in larger files when it's not obvious that these function calls are referring to the side panel

ndv99 (commented):
This is a cracking start, a few comments/changes/opinions below. But over-all I think this is so much cleaner than the existing implementation of side panel context. I especially like how the component is just passed directly, it's far more flexible and eliminates the need to maintain some massive set of key/value pairs with corresponding union types. Good stuff! :rocket:

src/app/base/side-panel-context-new.tsx
abuyukyi101198: Good catch!

src/app/base/side-panel-context-new.tsx
```diff
@@ -17,20 +17,55 @@ interface SidePanelState<TProps = Record<string, unknown>> {
...
+ *
+ * @returns Object containing:
+ *   - `isOpen`: Boolean indicating if the side panel is currently open
+ *   - `title`: Current panel title string
+ *   - `size`: Current panel size ('narrow' | 'regular' | 'wide' | 'large')
+ *   - `component`: Currently rendered component
+ *   - `props`: Props passed to the current component
+ *   - `open(component, title, props?, size?)`: Opens panel with given component
+ *   - `close()`: Closes the panel and resets state
+ *   - `setSize(size)`: Changes the panel width
```
ndv99, line 50: *Love* this, just make sure the method names match the actual returns

```suggestion
 * @returns Object containing:
 *   - `isOpen`: Boolean indicating if the side panel is currently open
 *   - `title`: Current panel title string
 *   - `size`: Current panel size ('narrow' | 'regular' | 'wide' | 'large')
 *   - `component`: Currently rendered component
 *   - `props`: Props passed to the current component
 *   - `openSidePanel({ component, title, props?, size? })`: Opens panel with given component
 *   - `closeSidePanel()`: Closes the panel and resets state
 *   - `setSidePanelSize(size)`: Changes the panel width
```

PR #5791, opened 2025-09-18
refactor(domains): switch to new side panel context MAASENG-5235

nehjoshi:
## Done

- Refactored all views and components in /domains to use the new side panel context
- Moved the "Add record" button down to the "records" table, instead of it being at the top of the page
- Updated tests to use renderWithProviders() and mockSidePanel(), wherever needed
- Removed unused types and constants
- Updated side-panel-context.ts to no longer import deleted constants

Note: 
- The actions for the resource records table do not currently use the side panel. After creating an inventory of such actions/forms, we can eventually switch those forms to side panel forms.
- Adding to the above point, the "Edit" button for the domain summary has not been changed, since it opens the form in the centre, instead of opening a side panel. 

## QA steps

Go to ```{BASE_URL}/domains```.
- [ ] Verify that clicking "Add domains" opens up the correct side panel
- [ ] Verify that the "Cancel" and "Save domain" buttons close the side panel after a successful action

For any domain, click the arrow in the "Actions" column, and click "Set default".
- [ ] Verify that the "Set default" side panel opens up
- [ ] Verify that the "Cancel" and "Set default" buttons close the side panel after a successful action

Click on any domain to enter its summary page
- [ ] Verify that clicking the "Delete domain" button opens up the correct side panel
- [ ] Verify that the "Cancel" and "Delete domain" buttons close the side panel after a successful action
- [ ] Verify that the "Add record" button is above the resource records table, instead of next to the "Delete domain" button
- [ ] Verify that clicking the "Add record" button opens up the correct side panel
- [ ] Verify that the "Cancel" and "Add record" buttons close the side panel after a successful action

## Fixes

Resolves [MAASENG-5235](https://warthogs.atlassian.net/browse/MAASENG-5235)

## Screenshots

Before:

Screenshot from 2025-09-18 17-44-53

After:

Screenshot from 2025-09-18 17-45-57


abuyukyi101198 (approved):
Looks great! Regarding your notes on the resource records table, the following ticket is for handling those exact issues. Good catch nonetheless. I see you also created a ticket for the table migration. Great work!

[MAASENG-5328](https://warthogs.atlassian.net/browse/MAASENG-5328)

PR #5823, opened 2025-10-30
fix: use underscores in bundle name and .tgz extension

ndv99:
## Done

- changed extension for bundle to `.tgz`
- swapped `-` with `_` in file name


alexsander-souza (approved):
+1

PR #5984, opened 2026-04-16
refactor: implement script incorporated in lint for checking steps duplication MAASENG-6416

rumanowskak:
## Done

- Added scripts/check-cucumber-steps.ts — a TypeScript script that scans all *.steps.ts files under cypress/support/step_definitions/ and reports two categories of issues:
      - Duplicate step patterns — the same Given/When/Then pattern registered more than once
      - Different patterns with identical implementations — step definitions whose callback bodies are character-identical after normalising whitespace and comments, indicating accidental duplication or missed consolidation
- The script handles both arrow function callbacks ((args) => { ... }) and regular function() callbacks (used when steps need this for shared state)
- Exits with code 1 when issues are found (with details printed to stdout), 0 when clean
- Added script to lint command

## Fixes

Resolves [MAASENG-6416](https://warthogs.atlassian.net/browse/MAASENG-6416)

## Notes

[MAASENG-6416]: https://warthogs.atlassian.net/browse/MAASENG-6416?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


scripts/check-cucumber-steps.ts
```diff
@@ -0,0 +1,378 @@
...
+
+  // ── function / async function expression ────────────────────────────────
+  const isAsync = content.startsWith("async", pos);
+  if (isAsync) {
+    pos += 5;
+    while (pos < content.length && /\s/.test(content[pos])) pos++;
+  }
+
+  if (content.startsWith("function", pos)) {
+    pos += 8; // skip "function"
```
ndv99: This is probably pedantic, but I think using the actual string length here is a little more airtight, and makes it clearer what we're actually doing.

```suggestion
    pos += "async".length;
    while (pos < content.length && /\s/.test(content[pos])) pos++;
  }

  if (content.startsWith("function", pos)) {
    pos += "function".length;
```

scripts/check-cucumber-steps.ts
```diff
@@ -0,0 +1,378 @@
...
+  }
+  return results;
+}
+
+// ── Parsing helpers ───────────────────────────────────────────────────────────
+
+/**
+ * Extract the first argument of a step registration call (string or /regex/).
+ * Returns the normalised pattern text and the position after the closing delimiter.
+ */
```
ndv99, line 40: All of these docstrings should have the parameters listed as well/

```suggestion
/**
 * Extract the first argument of a step registration call (string or /regex/).
 * Returns the normalised pattern text and the position after the closing delimiter.
 * 
 * @param content: The string to be parsed
 * @param pos: The starting position in the string
 */
```

scripts/check-cucumber-steps.ts
```diff
@@ -0,0 +1,378 @@
...
+function parsePattern(
+  content: string,
+  pos: number
+): { pattern: string; endPos: number } | null {
+  while (pos < content.length && /\s/.test(content[pos])) pos++;
```
ndv99: You shouldn't be changing the value of function arguments directly. I would suggest calling the parameter `startingPosition`, and then assign a local variable `position` with that as it's initial value. You've done this in the function directly below this, use that for an example

ndv99 (changes requested):
I think the concept here is really solid but the execution needs some work. The check-cucmber-steps file uses a lot of shorthand variable and function names, which can be hard for people to parse and maintain further down the line.

Also, the PR title is misleading as this doesn't introduce any new GitHub actions, it just adds an extra step to the `lint` script.

ndv99 (approved):
This looks much more readable, thanks for the changes 👍 LGTBM

PR #5989, opened 2026-04-21
fix: regenerate api client to remove unknown types

nehjoshi:
## Done

- Regenerated the OpenAPI client with the following command to prevent `unknown` types in `types.gen.ts`
```bash
export NODE_TLS_REJECT_UNAUTHORIZED=0 && openapi-ts && sed -i 's/ | unknown//g' src/app/apiclient/types.gen.ts
```


abuyukyi101198 (approved):
Thanks for the quick fix!

PR #5992, opened 2026-04-21
fix: manually remove unknown types when generating the client

nehjoshi:
## Done

Following the recent Pydantic upgrade, hey-api seems to map fields like the following to `last_name?: string | unknown`  types, leading to several failing tests and linting errors:
```json
 "last_name": {
            "anyOf": [
              {
                "type": "string"
              },
              {
                "type": "null"
              }
            ],
            "title": "Last Name"
          },
```
The obvious fix was to upgrade `hey-api` to a newer version (which I tried), but unfortunately this problem still occurs. While we find a better workaround, it might be safe to manually handle this issue, so we can continue working with the latest API changes. 
- This PR manually handles "null" types coming from the OpenAPI spec so that they are not mapped to `unknown` types


abuyukyi101198 (approved):
Does the job

PR #5994, opened 2026-04-22
fix(settings): change group_id to groupId

nehjoshi:
## Done

- Changed `group_id` to `groupId` to fix the failing tests


abuyukyi101198 (approved):
Missed those, thanks for the fix

PR #6000, opened 2026-04-29
fix: install snapd and core snaps locally before calling setup-maas

nehjoshi:
## Done

- Cypress, accessibility and sitespeed.io tests are currently failing due to missing dependency requirements
- This PR adds in the required snaps before calling `setup-maas`

## QA steps

- [x] Verify that all tests are now working (some cypress tests may fail for `machines_actions.feature` and related, but all others should work; this will be fixed in a following PR)

## Notes
- We may want to revert this change once GitHub has a runner for Resolute Raccoon, as right now we are running Resolute dependencies on Noble runners


ndv99 (approved):
LGTM

PR #6001, opened 2026-04-29
feat: add e2e tests for the groups view MAASENG-6348

nehjoshi:
## Done

- Added new tests and step-definitions for the end-to-end testing of user-groups

## Fixes

Resolves [MAASENG-6348](https://warthogs.atlassian.net/browse/MAASENG-6348)

## Notes

- Can confirm that the tests pass successfully when running them locally. You can try them with:
```sh
yarn cypress run --spec cypress/e2e/with-users/features/settings/users/settingsGroupsList.feature
yarn cypress run --spec cypress/e2e/with-users/features/settings/users/settingsGroupDetails.feature
```

[MAASENG-6348]: https://warthogs.atlassian.net/browse/MAASENG-6348?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


cypress/support/step_definitions/settings/users/settingsGroupDetails.steps.ts
```diff
@@ -0,0 +1,23 @@
...
+  groupName = `cy-group-${nanoid()}`;
+  cy.get("input[name='name']").type(groupName);
+});
+
+Then("the group creation should succeed", () => {
+  cy.findByText("Save group").should("not.exist");
+  cy.findByText(groupName).should("exist");
+});
+
+When("the user clicks on the newly created group link", () => {
```
rumanowskak: we have a common step 

When("the user clicks the {string} link", (link: string) => {
  cy.findByRole("link", { name: link }).click();
});

rumanowskak (changes requested):
All in all very good, just small change requested

cypress/support/step_definitions/settings/users/settingsGroupDetails.steps.ts
nehjoshi: Since this is a newly created entity, we don't know it's exact name, so using the common step fails here. It's similar to this:
```
Scenario: The user can add a tag to the controller
    When the user adds a tag to the controller
    And the user clicks the created tag
    Then the correct tag is displayed in the searchbox
    And the correct number of controllers is displayed
 ```
 We created a custom action for "And the user clicks the created tag"

rumanowskak: maybe then rename it a little more to differentiate those two steps, but your call.

rumanowskak (approved):

PR #6007, opened 2026-05-01
chore: add e2e test for adding DNS records MAASENG-6452

nehjoshi:
## Done

- Added an end-to-end verifying that new DNS records can be added and are correctly hyperlinked

You can run this test locally with:
```sh
yarn cypress run --spec "cypress/e2e/with-users/features/domains/domainsAddRecordAssignsDevice.feature"
```

## Fixes

Resolves [MAASENG-6452](https://warthogs.atlassian.net/browse/MAASENG-6452)

[MAASENG-6452]: https://warthogs.atlassian.net/browse/MAASENG-6452?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

## Notes
- Relies on cypress environment variables to persist data temporarily


nehjoshi:
Failing tests are irrelevant (machine-related e2e tests)

rumanowskak (changes requested):
1. I see you have changed common step when the user clicks the link, but you still have defined two separate steps when the user clicks the record name link and when the user clicks the default domain link. If there are those two separate steps there is no point in changing the common one, also change those separate test names I feature file so that it is different than the common step, because when the user clicks the {string} link is too similar to when the user clicks the record name link  and when the user clicks the default domain link.
2. check if separate step Then("the user should be redirected to the device details page"), couldn't use instead common step Then("the pathname should equal {string}", (expectedPath: string) => {
  cy.location("pathname").should("eq", generateMAASURL(expectedPath));
});

nehjoshi:
Changes made:
1. I modified the common `the pathname should equal ...` action so now it allows dynamic paths. E.g. when you click on a domain `abc123` it'll check that you are redirected  to `/domains/abc123/details`. We'll need this for other views later as well, I think, such as `/groups/group-1/details`.
2. The DNS-specific steps use clearly distinct names that cannot be confused with the common `{string}` patterns

rumanowskak (approved):
LGTM

PR #6011, opened 2026-05-05
chore(machines): refactor MachineDetails routes to centralised routes in router.tsx MAASENG-6079

ndv99:
## Done

- Moved element routes from MachineDetails to object routes in router
- Updated sub-components to pull ID from URL instead of prop
- Removed unnecessary test for checking each route within MachineDetails itself

## QA steps

- [x] Go to a machine's details page
- [x] Ensure it loads on the summary page
- [x] Click all the other tabs and ensure they load correctly
- [x] Poke around for any edge cases

## Fixes

Resolves [MAASENG-6079](https://warthogs.atlassian.net/browse/MAASENG-6079)

[MAASENG-6079]: https://warthogs.atlassian.net/browse/MAASENG-6079?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/app/machines/views/MachineDetails/MachineDetails.tsx
+++ b/src/app/machines/views/MachineDetails/MachineDetails.tsx
@@ -1,21 +1,9 @@
 import { useEffect } from "react";
 
 import { useDispatch } from "react-redux";
-import { Navigate, Route, Routes, useLocation } from "react-router";
+import { Outlet, useLocation } from "react-router";
 
-import MachineConfiguration from "./MachineConfiguration";
 import MachineHeader from "./MachineHeader";
-import MachineInstances from "./MachineInstances";
-import MachineLogs from "./MachineLogs";
-import MachineNetwork from "./MachineNetwork";
-import NetworkNotifications from "./MachineNetwork/NetworkNotifications";
-import MachinePCIDevices from "./MachinePCIDevices";
-import MachineScript from "./MachineScripts";
-import MachineStorage from "./MachineStorage";
-import StorageNotifications from "./MachineStorage/StorageNotifications";
-import MachineSummary from "./MachineSummary";
-import SummaryNotifications from "./MachineSummary/SummaryNotifications";
-import MachineUSBDevices from "./MachineUSBDevices";
 
 import ModelNotFound from "@/app/base/components/ModelNotFound";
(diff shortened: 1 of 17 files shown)


src/app/machines/views/MachineDetails/MachineDetails.tsx
```diff
@@ -53,123 +41,21 @@ const MachineDetails = (): React.ReactElement => {
...
-        modelName="machine"
-      />
+      <>
+        <ModelNotFound
+          id={id}
+          linkURL={urls.machines.index}
+          modelName="machine"
+        />
+        pathname: {pathname}
+        id: {id}
```
nehjoshi: Question: Why do we need to print the `pathname` and `id` here? I don't see this pattern used elsewhere so I'm just wondering

nehjoshi (commented):
A question below ->
Apart from that, looks good!
Also, this is unrelated, but navigating to the details of a non-existent machine seems to infinitely render "Loading..." instead of `<ModelNotFound />`. It's the same on the demo instance (e.g. you can try https://maas-ui-demo.internal:5443/MAAS/r/machine/wrrxhxa/summary where`wrrxhxa` doesn't exist). This probably needs to be fixed separately

ndv99:
> A question below -> Apart from that, looks good! Also, this is unrelated, but navigating to the details of a non-existent machine seems to infinitely render "Loading..." instead of `<ModelNotFound />`. It's the same on the demo instance (e.g. you can try https://maas-ui-demo.internal:5443/MAAS/r/machine/wrrxhxa/summary where`wrrxhxa` doesn't exist). This probably needs to be fixed separately

Good spot - funnily enough I ran into this same issue while updating the unit tests and I thought I'd fixed it. Will take a look!

src/app/machines/views/MachineDetails/MachineDetails.tsx
ndv99: oh... we don't. that was for my own debugging :sweat_smile: thanks again for the good spot!

nehjoshi (approved):
+1. Thanks for addressing the missing machine issue!

PR #6015, opened 2026-05-06
chore: update zones_with_summary to zones:statistics MAASENG-6470

nehjoshi:
## Done

- Updated `/zones_with_summary` to `/zones:statistics`
- Updated tests, resolvers and factories

## QA steps
Note: `maas-ui-demo` still uses `/zones_with_summary`. Use a local MAAS instance with the latest changes to verify.
- [ ] Go to `/MAAS/r/zones` and verify the table is populated as expected

## Fixes

Resolves [MAASENG-6470](https://warthogs.atlassian.net/browse/MAASENG-6470)

[MAASENG-6470]: https://warthogs.atlassian.net/browse/MAASENG-6470?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/app/api/query/zones.ts
+++ b/src/app/api/query/zones.ts
@@ -1,3 +1,4 @@
+import type { UseQueryResult } from "@tanstack/react-query";
 import { useQueryClient, useMutation } from "@tanstack/react-query";
 
 import { useWebsocketAwareQuery } from "@/app/api/query/base";
@@ -15,6 +16,9 @@ import type {
   GetZoneData,
   GetZoneErrors,
   GetZoneResponses,
+  ListZonesData,
+  ListZonesErrors,
+  ListZonesResponses,
   ListZonesWithStatisticsData,
   ListZonesWithStatisticsErrors,
   ListZonesWithStatisticsResponses,
@@ -22,6 +26,7 @@ import type {
   UpdateZoneData,
   UpdateZoneErrors,
   UpdateZoneResponses,
+  ZoneWithStatisticsResponse,
 } from "@/app/apiclient";
 import {
   deleteZone,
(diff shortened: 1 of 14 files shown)


ndv99 (commented):
This doesn't seem quite right - the zones statistics should live in a separate endpoint from the rest of the data, and the zones table should be populated by two hooks - one to get the base data, and another to fetch the statistics data (ideally using the list of IDs from the former to filter the latter). Can you check the back-end implementation for this?

nehjoshi:
Aah yes you're right. I will update this

src/app/api/query/zones.ts
```diff
@@ -15,43 +16,103 @@ import type {
...
+      options,
+      listZonesWithStatistics,
+      listZonesWithStatisticsQueryKey(options)
+    ),
+    enabled,
+  });
 };
 
-export const useZoneCount = (options?: Options<ListZonesWithSummaryData>) => {
+export const useZoneCount = (options?: Options<ListZonesData>) => {
```
nehjoshi, line 109: Since we just want the total, using `ListZones` is better because it has a smaller payload -> faster response

nehjoshi (commented):
Finished separating `useZones` and `useZonesWithStatistics`

src/app/api/query/zones.ts
ndv99: Nice one, that's a good optimisation :+1:

src/app/kvm/views/KVMList/LXDHostsTable/LXDHostsTable.tsx
```diff
@@ -46,7 +46,7 @@ export type LXDKVMHost = {
 
 export const generateSingleHostRows = (
   pods: Pod[],
-  zones: ZoneWithSummaryResponse[] | undefined
+  zones: Pick<ZoneWithStatisticsResponse, "id" | "name">[] | undefined
```
ndv99: `ZoneWithStatisticsResponse` won't have the name, just the additional stats (at least, it shouldn't). You can just use the regular `ZoneResponse` here

src/app/zones/components/DeleteZone/DeleteZone.tsx
```diff
@@ -47,7 +47,7 @@ const DeleteZone: React.FC<DeleteZoneProps> = ({ id }) => {
           onSuccess={() => {
             return queryClient
               .invalidateQueries({
-                queryKey: listZonesWithSummaryQueryKey(),
+                queryKey: listZonesWithStatisticsQueryKey(),
```
ndv99: We should probably invalidate the normal `listZones` query key as well here

src/app/zones/components/ZonesTable/ZonesTable.test.tsx
```diff
@@ -69,9 +70,18 @@ describe("ZonesTable", () => {
...
+          ],
+          total: 1,
+        })
+      );
+
+      mockServer.use(
+        zoneResolvers.listZonesWithStatistics.handler({
+          items: [
+            factory.zoneWithStatistics({
+              name: "default",
```
ndv99: I may have missed something here, but I thought `ZoneWithStatistics` shouldn't have the name?

src/testing/factories/zone.ts
```diff
@@ -1,12 +1,18 @@
...
-  devices_count: random,
-  machines_count: random,
   name: (i: number) => `zone-${i}`,
   id: (i: number) => i,
 });
+
+export const zoneWithStatistics = define<ZoneWithStatisticsResponse>({
+  id: (i: number) => i,
+  name: (i: number) => `zone-${i}`,
+  description: "test description",
```
ndv99: ^ same as above but also description, IIRC this endpoint should only have the additional information and not the base info (except the ID)

ndv99 (commented):
This is looking better but I'm still a bit confused about the response shape here

src/app/kvm/views/KVMList/LXDHostsTable/LXDHostsTable.tsx
nehjoshi: Ah yes, good catch

src/app/zones/components/ZonesTable/ZonesTable.test.tsx
nehjoshi: Oh yeah, I'll remove the name/description fields from the factories, and then I'll open a PR so that the backend doesn't return them either.

nehjoshi:
- Removed `name` and `description` fields from `ZonesResponse` (now they're included only in `ZonesStatistics`)
- Added query invalidation for both `listZones` and `listZonesStatistics` when a zone is deleted

ndv99 (approved):
I think this is good to go now, thanks for addressing all the comments! One nitpick would be that it may be nice to have loading spinners in the stats columns to show a loading state, but tbh it still looks clean even without them, so this is totally optional. Nice work!

PR #6033, opened 2026-05-14
feat: implement tool flagging files that stand out against documentation

rumanowskak:
## Done

- Added scripts/check-standards.ts — a static analysis script that checks source files against the patterns defined in docs/standards/. Detects violations such as data-testid usage, class components, direct apiclient value imports, raw <form> elements, Redux slice/saga usage outside src/app/store/, Playwright in E2E tests, and redux-mock-store usage
- Added yarn check-standards (full codebase scan) and yarn check-standards:changed (changed files only, relative to main) scripts to package.json
- Added --skip=<keyword> flag to filter out violations by doc or message keyword (e.g. --skip=testing to focus on non-testing violations)
- Added --show-legacy flag — violations in known legacy paths (src/app/base/sagas) are hidden by default and only shown when this flag is passed
- Added // check-standards-ignore: <reason> directive — developers can suppress a specific violation by adding this comment on the line above. A reason is required; a bare // check-standards-ignore without a reason is itself flagged as an error
- Output shows a summary table (total count by standards doc) before the per-file details, so key numbers are visible even when the details scroll off the terminal
- Each violation is printed in file:line:col format, making them clickable in VS Code's terminal to jump directly to the offending line
- Added .github/workflows/standards-review.md and compiled .github/workflows/standards-review.lock.yml — a GitHub Copilot agentic workflow with two modes: weekly automated scan of the full src/ and cypress/ directories (Layer 1 static + Layer 2 semantic checks), and on-demand PR-specific check triggered via workflow_dispatch with a PR number
- Updated docs/standards/api-hooks.md to document that import type from @/app/apiclient and query key helpers from @/app/apiclient/@tanstack/react-query.gen are allowed in component files — only other value imports are banned
- Updated tsconfig.json to include scripts/ so TypeScript type-checks the checker script as part of yarn lint
- Added // check-standards-ignore to MachineListPagination.test.tsx where a raw <form> is used intentionally as a test wrapper, not as an application form

## QA steps

- [ ]


ndv99 (approved):
LGTM

PR #6041, opened 2026-05-20
refactor(controllers): ControllerVLANsTable to GenericTable MAASENG-5680

abuyukyi101198:
## Done

- Refactored `ControllerVLANsTable` to use `GenericTable`
- Added fabric grouping
- Fixed tests due to semantic changes

## QA steps

- [x] Open this PR and the demo side-by-side
- [x] Navigate to /controllers
- [x] Go to any controller's details page
- [x] Go to the VLAN tab
- [x] Compare the tables
- [x] Verify the fabric grouping is a sensible addition
- [x] Verify the information and links are correct

## Fixes

Resolves [MAASENG-5680](https://warthogs.atlassian.net/browse/MAASENG-5680)

[MAASENG-5680]: https://warthogs.atlassian.net/browse/MAASENG-5680?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


nehjoshi:
I think there is an issue with sorting the  "VLAN" and the "Subnets" column - clicking on the sort arrow doesn't seem to make a difference. 
Also, I noticed something weird. If you go to this controller (`/controller/y7bdtk/vlans` - `juju-c30a94-prod-maas-machine-lab-13`), the demo instance shows two subnets in two fabrics.
Whereas in this PR, we only see 1 fabric (`fabric-1157`) with both those subnets, and `fabric-1158` isn't visible. Maybe the grouping logic needs to be reviewed?
Also, I think `src/app/controllers/views/ControllerDetails/ControllerVLANs/ControllerVLANsTable/constants.ts` is no longer imported anywhere, so it should be safe to delete

src/app/controllers/views/ControllerDetails/ControllerVLANs/ControllerVLANsTable/useControllerVLANsTableColumns/useControllerVLANsTableColumns.tsx
```diff
@@ -0,0 +1,94 @@
...
+>;
+
+const useControllerVLANsTableColumns = (): ControllerVLANsColumnDef[] => {
+  return useMemo(
+    () => [
+      {
+        id: "fabric",
+        header: "Fabric",
+        accessorKey: "fabric_id",
+        enableSorting: true,
```
nehjoshi: Sorting still doesn't seem to work for this column. I think if we revert it to what it was earlier, that is, we don't allow sorting on any of the columns of this table, that should be fine

nehjoshi (changes requested):
Just one comment below ->
(Also, we can probably nuke `src/app/controllers/views/ControllerDetails/ControllerVLANs/ControllerVLANsTable/constants.ts` as I don't think it's used anymore)

src/app/controllers/views/ControllerDetails/ControllerVLANs/ControllerVLANsTable/useControllerVLANsTableColumns/useControllerVLANsTableColumns.tsx
abuyukyi101198: Oh my bad, the original table had no sorting at all. That's a leftover. I'll remove it.

nehjoshi (approved):
+1 Thanks for making the updates!

PR #6048, opened 2026-05-25
test: Partial Cypress MAASENG-6499

abuyukyi101198:
## Done

- Added src/ and features/ domain change detection for with-users runs
- Defined cron job for full runs
- Separated no-users and with-users into separate jobs
- Added partial runs and "no change" skipping for with-users
- Removed Keycloak setup from no-users
- Renamed dashboard features to networkDiscovery

## Fixes

Resolves [MAASENG-6499](https://warthogs.atlassian.net/browse/MAASENG-6499)

[MAASENG-6499]: https://warthogs.atlassian.net/browse/MAASENG-6499?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


.github/workflows/cypress.yml
```diff
@@ -5,17 +5,207 @@ on:
...
+            settings
+            zones
+          )
+
+          # Collect the app/ directories touched by this PR.
+          # grep || true prevents exit code 1 under pipefail when nothing matches.
+          BASE_SHA="${{ github.event.pull_request.base.sha }}"
+          HEAD_SHA="${{ github.event.pull_request.head.sha }}"
+          CHANGED_APP=$(git diff --name-only "$BASE_SHA" "$HEAD_SHA" | \
+                        (grep '^app/' || true) | \
```
nehjoshi: Shouldn't this be `(grep '^src/app/' || true) | \`?

nehjoshi (commented):
One question ->

.github/workflows/cypress.yml
abuyukyi101198: Good catch, I didn't test this with src changes so it slipped by

abuyukyi101198: Fixed

nehjoshi (approved):
LGTM! If there are still issues with the workflow we can always make changes later. ALso, 19 minutes for cypress tests is already a solid improvement! 🎉

PR #6050, opened 2026-05-25
test: fix on-main runs

abuyukyi101198:
## Done

- Full run on any event other than PR, since diff is only available for PRs


nehjoshi (approved):
+1

PR #6057, opened 2026-05-27
fix: candid/rbac integration for 3.8

alemar99:
Candid/RBAC integration for 3.8 in the UI was broken: logging in resulted in an endless loop.
Furthermore, we cannot rely anymore on the sessionid for the v3 API calls, so we have to use the macaroons (which are set server-side by MAAS, see https://github.com/canonical/maas/pull/279). When this expires, we have to make sure the user is logged out and it is redirected to the login page.

## Done

- The login page uses the `authenticated` redux state
- The external login flow is now only started when `externalAuthURL` exists AND the user is not already authenticated
- Added a dedicated `externalSessionExpired` status action for expired/invalid Candid/RBAC macaroon sessions.
- Added a separate response interceptor `checkExternalSessionExpired`, that logs the user out whenever the macaroon is not valid anymore and the redux state says the user is authenticated

## QA steps

- Setup a MAAS environment with candid/rbac integration enabled (I can provide you a script)
- Go to the login page
- Login through Candid
- Make sure you're redirected and no loop happen

To test the session expired:
- Change the `MACAROON_LIFESPAN` variable in MAAS, set it to 1-2 minutes
- Login through Candid
- Open the dev tools to see the network requests
- Wait until a 401 request to the v3 API is made (This will happen as we constantly fetch the notifications for example)
- See that you are redirected to the login page
- NOTE: in this case you will be automatically re-logged in as the Candid's session is still valid (has a lifespan of 1 day). Just make sure to see that the redirect happens, or
- Once you are logged in, clear the local storage for:
  - CANDID url, usually http://IP:8081
  - identity (Candid Identity)
  - RBAC url, usually http://IP:5000/auth (should not be necessary, but just in case)

## Fixes

Resolves [MAASENG-6532](https://warthogs.atlassian.net/browse/MAASENG-6532)

## Notes

[MAASENG-6532]: https://warthogs.atlassian.net/browse/MAASENG-6532?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/app/api/auth-interceptor.test.ts
+++ b/src/app/api/auth-interceptor.test.ts
@@ -2,10 +2,15 @@ import { describe, it, expect, vi, beforeEach } from "vitest";
 
 import { COOKIE_NAMES } from "../utils/cookies";
 
-import { configureAuthInterceptor } from "./auth-interceptor";
+import {
+  checkExternalSessionExpired,
+  configureAuthInterceptor,
+} from "./auth-interceptor";
 
 import { client } from "@/app/apiclient/client.gen";
+import { statusActions } from "@/app/store/status";
 import { getCookie } from "@/app/utils";
+import { store } from "@/redux-store";
 
 vi.mock("@/app/apiclient/client.gen", () => ({
   client: {
@@ -13,7 +18,17 @@ vi.mock("@/app/apiclient/client.gen", () => ({
       request: {
         use: vi.fn(),
       },
+      response: {
+        use: vi.fn(),
+      },
     },
+  },
+}));
+
+vi.mock("@/redux-store", () => ({
+  store: {
+    dispatch: vi.fn(),
+    getState: vi.fn(),
   },
 }));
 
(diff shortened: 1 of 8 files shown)


nehjoshi (changes requested):
In general, this works as expected, but there are a few tweaks to consider:
1. When external login is enabled and you go to the MAAS login page, you are redirected to the candid login page twice (two new tabs open) instead of just once. I think it's best if we just remove that automatic redirection, and just let the user click the "Go to login page" button, so that the user is in control and the experience is less abrupt. I think this can be fixed by removing the `window.open()` line in `src/bakery.ts`
2. `src/app/login/Login/Login.test.tsx` is missing an import causing the lint issue.

Apart from this, looks good!

nehjoshi (approved):
LGTM, nice one!

PR #6068, opened 2026-06-10
feat: Migrate documentation updating agent to Gemini MAASENG-6553

rumanowskak:
## Done

- migrated documentation updating agent to Gemini

## QA steps

- [ ]

## Fixes

Resolves [MAASENG-6553](https://warthogs.atlassian.net/browse/MAASENG-6553)

## Notes

[MAASENG-6553]: https://warthogs.atlassian.net/browse/MAASENG-6553?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


.github/workflows/docs-maintenance.yml
```diff
@@ -0,0 +1,37 @@
...
+  update-docs:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+        with:
+          fetch-depth: 0
+
+      - uses: actions/setup-node@v4
+        with:
+          node-version: "20"
```
abuyukyi101198: The `setup-node` action was updated recently, and it is better for maintenance to use `.nvmrc` for the Node version:
```suggestion
      - uses: actions/setup-node@v6
        with:
          node-version-file: ".nvmrc"
```

scripts/update-docs.ts
```diff
@@ -0,0 +1,296 @@
+import { execSync } from "child_process";
+import * as fs from "fs";
+import * as path from "path";
+
+const GEMINI_API_KEY = process.env.GEMINI_API_KEY;
+if (!GEMINI_API_KEY) {
+  console.error("❌ GEMINI_API_KEY environment variable is required");
```
abuyukyi101198: A lot of emojis 😬

.github/workflows/docs-maintenance.yml
```diff
@@ -0,0 +1,37 @@
+name: Docs Maintenance
+
+on:
```
abuyukyi101198, line 3: I think the job should also run on `push` to a debug branch like `debug-docs-maintenance-*`.

abuyukyi101198 (changes requested):
Some minor changes inline. Also, are you sure there are enough safeguards for prompt injection? Or is that gonna be a later PR? Because if so, I would recommend merging to a feature branch first.

abuyukyi101198 (approved):
LGTM, but I'm still reluctant on the possibility of prompt injections. We have branch protections, but it still may leave an open door.

I'd ask for additional reviews from the whole team. @ndv99 @nehjoshi

ndv99:
> LGTM, but I'm still reluctant on the possibility of prompt injections. We have branch protections, but it still may leave an open door.
> 
> I'd ask for additional reviews from the whole team.

Is the attack surface here just file content? For branches beginning with the debug string, only Canonical team members have access to create branches directly on the repo, so that should be okay in theory. And otherwise it runs on main, so it pretty much hinges on us being good human code reviewers and catching any attempts at prompt injection in changed files. Which should be fairly easy to spot in theory

scripts/update-docs.ts
```diff
@@ -0,0 +1,334 @@
...
+    return [];
+  }
+}
+
+function mapFilesToDocs(files: string[]): Map<string, Set<string>> {
+  const result = new Map<string, Set<string>>();
+
+  for (const file of files) {
+    for (const mapping of PATH_TO_DOCS) {
+      if (mapping.pattern.test(file)) {
```
nehjoshi: Optional nit, but for better readability, we could return early in the first `if` statement and prevent additional nesting:
```ts
for (const file of files) {
    for (const { pattern, docs } of PATH_TO_DOCS) {
      if (!pattern.test(file)) continue;

      for (const doc of docs) {
                 ...
```

nehjoshi (approved):
LGTM, with one nit below. Also, do we have a team-based API key that the CI will use to talk to Gemini?
As for the concern of prompt injection, I think it should be fine with the current safeguards, since the tool ultimately creates a PR for us to review (provided we are meticulous when reviewing it).

ndv99 (approved):

PR #6069, opened 2026-06-10
fix(kvm): dark mode background for NumaResourcesCard.tsx

abuyukyi101198:
## Done

- Removed the light theme background extension from NUMA summary card and LXD column popovers

## QA steps

- [x] Navigate to LXD list
- [x] Hover over the RAM and CPU Cores cells in both dark and light modes
- [x] Verify the popovers are correctly themed
- [x] Navigate to the details of a KVM with NUMA nodes
- [x] Toggle the "View by NUMA node" switch
- [ ] Verify the NUMA cards are correctly themed across light and dark modes

## Screenshots
Screenshot 2026-06-10 at 16 02 58
Screenshot 2026-06-10 at 16 02 51
Screenshot 2026-06-10 at 15 41 58
Screenshot 2026-06-10 at 15 42 04

### DIFF
--- a/src/app/base/components/DoughnutChart/DoughnutChart.tsx
+++ b/src/app/base/components/DoughnutChart/DoughnutChart.tsx
@@ -4,8 +4,6 @@ import { Tooltip } from "@canonical/react-components";
 import { nanoid } from "@reduxjs/toolkit";
 import classNames from "classnames";
 
-import useDarkMode from "../../hooks/useDarkMode/useDarkMode";
-
 type Segment = {
   /**
    * The colour of the segment.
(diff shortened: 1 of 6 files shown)


nehjoshi:
If you start with dark mode, refresh the page, then change to light mode the text inside the allocation circle seems to disappear:
image

But if you switch back to dark mode, you can see the text (0B) just fine:
image

And if you do it the other way around, so start with light mode, refresh the page, then change to dark mode, you'll get:

image

I guess this is due to the color contrast not being enough, or not updating when it should.

abuyukyi101198:
That update issue is probably caused by that label being an SVG group instead of actual text. It should be fixed now, but not much else to do otherwise.

nehjoshi (approved):
+1 Thanks for the fix!

PR #6077, opened 2026-06-23
refactor: switch to user entitlements for deriving permissions MAASENG-6645

nehjoshi:
## Done
There are 2 major changes this PR introduces:
1. The `UsersTable` in the settings now displays the groups column instead of roles. The `AddUser` and `UpdateUser` forms allow you to add a user to a group when they are created/updated
2. The settings views are now guarded based on the user's entitlements. If the user has `CAN_VIEW_*` entitlements, they can only view, but not edit any of the forms in the settings. The user must have `CAN_EDIT_*` in order to add or edit data.

## How to Review This PR
- Due to the massive diff, I suggest starting with a QA to get an idea of the changes introduced
- You will see that most files have more-or-less the same change, where a new hook has been added, that computes whether the user has the necessary entitlements to perform an operation:
```ts
const canEdit = useHasEntitlements([Entitlement.CAN_EDIT_GLOBAL_ENTITIES]);
```
And `canEdit` is then used to allow a form to be updated, or a button to be enabled. The same applies for `canView`, wherever it is used.
- Also, I suggest starting by reviewing the hook in `src/app/base/hooks/permissions.ts` as it forms the basis for all permission checks
- A significant chunk of the PR is rewriting tests to work with these new changes

## QA steps
Use `maas-ui-demo` as the backend for QAing. There, I have created two users with the usernames `adam` and `bob`. They share the same password `maas`.
- Adam belongs to the `basic` group and has readonly access to all entities
- Bob belongs to no group and has no access to any entities.

### Part 1: QA for authorization
- [x] Clear your cookies or use incognito
- [x] From the login page, sign in as `adam`
- [x] Ensure the intro screen is visible. You may add SSH keys or decide to skip the intro
- [x] After signing in, ensure that you can see the "Settings" tab in the app navigation menu
- [x] Click "Settings"
- [x] Go through each page in settings, and ensure the forms, tables, and buttons are all disabled
- [x] Ensure that you can view the information in each form/table, but cannot perform add/edit/delete operations
- [x] Navigate to other parts of maas. You should be able to view all pages (except some pages that are currently returning 500: Internal Server Error)
- [x] Logout, and sign in as `bob`
- [x] Since bob has no entitlements, ensure the "Settings" tab is not shown in the side navigation
- [x] Ensure navigating to any of the settings pages manually (by going to `/MAAS/r/settings/security/session-timeout` for example) shows an error message
- [x] Ensure the machine page shows "0" machines

### Part 2: QA for the Users Table
- [x] Login as a privilaged user, such as `maas` or your own account
- [x] From settings, go to users
- [x] Ensure the users table has the groups column, and ensure you can visit each group by clicking on them
- [x] Click "Add User"
- [x] Ensure the "Groups" multi-select field is shown
- [x] Click the Edit button for a user from the table
- [x] Ensure the "Groups" field is pre-populated with the groups the user belongs to
- [x] Ensure removing the user from existing group, and then adding them to a new group updates the table accordingly (I suggest doing this on a test user)

## Fixes

Resolves [MAASENG-6645](https://warthogs.atlassian.net/browse/MAASENG-6645)

## Notes

- Note that this PR only covers the "Settings" forms, tables, and buttons. Other forms throughout the application should also be updated for good UX, though it is not critical for 3.8. Previously, if a non-admin user tried to perform a privileged operation, the form submission would still work, but the backend would return 403, which was slightly misleading. The Settings pages, at least, do not have this issue any more.

[MAASENG-6645]: https://warthogs.atlassian.net/browse/MAASENG-6645?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/app/intro/views/Intro.tsx
```diff
@@ -31,9 +36,20 @@ const Intro = (): ReactElement => {
...
+  const isSuperUser = useHasEntitlements([
+    Entitlement.CAN_EDIT_CONFIGURATIONS,
+    Entitlement.CAN_EDIT_GLOBAL_ENTITIES,
+    Entitlement.CAN_EDIT_IDENTITIES,
+    Entitlement.CAN_EDIT_MACHINES,
+    Entitlement.CAN_EDIT_BOOT_ENTITIES,
+    Entitlement.CAN_EDIT_LICENSE_KEYS,
+    Entitlement.CAN_EDIT_CONTROLLERS,
+    Entitlement.CAN_EDIT_NOTIFICATIONS,
+  ]);
```
abuyukyi101198: Seems like an obscure place to maintain these permissions in case a new one is added. Would a separate useIsSuperuser in the same place as useHasEntitlements make sense?

abuyukyi101198 (changes requested):
It looks like LXD, Virsh, Tags, and Networks are still available to no-permission users (bob). Also the forms of Images still seem to be available, although not populated.

src/app/settings/views/UserManagement/views/UsersList/components/AddUser/AddUser.tsx
```diff
@@ -65,41 +84,60 @@ const AddUser = (): ReactElement => {
...
+                throw new FormikFieldChangeError(
+                  "groups",
+                  "setFieldValue",
+                  reason as string
+                );
+              });
+            }}
+            placeholder="Select groups"
+            tags={groupTags}
+          />
```
abuyukyi101198: I think a MultiSelect may be more appropriate than a TagSelector here.

src/app/intro/views/Intro.tsx
nehjoshi: Sure, that's a good idea. Will update

src/app/settings/views/UserManagement/views/UsersList/components/GroupMultiSelect/GroupMultiSelect.tsx
```diff
@@ -0,0 +1,119 @@
...
+          }}
+          placeholder={placeholder}
+          selectedItems={groupItems.filter((item) =>
+            field.value?.includes(item.value)
+          )}
+          variant="condensed"
+        />
+      </div>
+      {help ? (
+        <p className="p-form-help-text" style={{ marginTop: "0.5rem" }}>
```
nehjoshi, line 111: I don't really have a better way to do this. Without it, the help text is overlapping with the input field.

src/app/settings/views/UserManagement/views/UsersList/components/GroupMultiSelect/GroupMultiSelect.tsx
```diff
@@ -0,0 +1,119 @@
...
+  label?: string;
+  name: string;
+  placeholder?: string;
+};
+
+export enum Labels {
+  DefaultLabel = "Groups",
+}
+
+export const GroupMultiSelect = ({
```
nehjoshi, line 22: I created a separate component since this dropdown is used in both AddUser and EditUser, and there was a bunch of repeated logic.  Also, a lot of this code is taken from the dropdown in `SelectUpstreamimagesForm.tsx`.

nehjoshi (commented):
Some explanations below. Also, LXD/Virsh hosts, networks, tags still being rendered is a backend issue that is currently being looked into.

abuyukyi101198 (approved):
Looks great! Thanks for implementing the multi select

nehjoshi:
And thank you for the (massive) review! :)

PR #6080, opened 2026-06-26
fix: add missing condition for enabling redirect_uri form field

nehjoshi:
## Done

- In the single sign on form, the "Redirect URI" form field is always disabled. Instead, it should only be disabled when `disabled={!canEdit}`


ndv99 (approved):
LGTM

PR #6090, opened 2026-07-01
chore(deps): upgrade maas-react-components to 2.4.0

ndv99:
## Done

- upgrade maas-react-components to 2.4.0
- changed queries for "cell" to "gridcell" in tests for generic tables
- changed queries for "grid" to "treegrid" in tests for generic tables

### DIFF
--- a/src/app/base/components/node/networking/NetworkTable/NetworkTable.test.tsx
+++ b/src/app/base/components/node/networking/NetworkTable/NetworkTable.test.tsx
@@ -144,7 +144,7 @@ describe("NetworkTable", () => {
       );
 
       const row = screen.getAllByRole("row")[1];
-      const cells = within(row).getAllByRole("cell");
+      const cells = within(row).getAllByRole("gridcell");
       const subnetCell = cells[6];
       expect(subnetCell).toHaveTextContent("Unconfigured");
       const ipCell = cells[7];
@@ -192,7 +192,7 @@ describe("NetworkTable", () => {
       );
 
       const row = screen.getAllByRole("row")[1];
-      const cells = within(row).getAllByRole("cell");
+      const cells = within(row).getAllByRole("gridcell");
       const subnetCell = cells[6];
       expect(subnetCell).toHaveTextContent("subnet-cidr");
       const ipCell = cells[7];
@@ -247,7 +247,7 @@ describe("NetworkTable", () => {
 
       const rows = screen.getAllByRole("row").slice(1);
       rows.forEach((row, index) => {
-        const cells = within(row).getAllByRole("cell");
+        const cells = within(row).getAllByRole("gridcell");
         const nameCell = cells[1];
         expect(nameCell).toHaveTextContent("alias");
         const subnetCell = cells[6];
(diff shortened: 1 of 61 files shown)


src/app/base/components/node/networking/NetworkTable/NetworkTable.test.tsx
```diff
@@ -296,7 +296,7 @@ describe("NetworkTable", () => {
       const bondRow = rows[1];
       expect(within(bondRow).getByRole("checkbox")).toBeInTheDocument();
       const nicRow = rows[2];
-      expect(within(nicRow).queryByRole("checkbox")).toBeDisabled();
+      expect(within(nicRow).queryByRole("checkbox")).not.toBeInTheDocument();
```
abuyukyi101198, line 299: Is this correct? I am not sure if there is a checkbox visibility change othe than the loading states

src/app/settings/views/UserManagement/views/Groups/components/GroupEntitlementsTable/GroupEntitlementsTable.test.tsx
```diff
@@ -93,7 +93,7 @@ describe("GroupEntitlementsTable", () => {
       );
 
       await waitFor(() => {
-        expect(screen.getAllByRole("checkbox").length).toBeGreaterThan(0);
+        expect(screen.getAllByRole("checkbox").length).toBeGreaterThan(1);
```
abuyukyi101198, line 96: I assume these are previously false tests you noticed?

src/app/settings/views/UserManagement/views/Groups/components/GroupMembersTable/GroupMembersTable.test.tsx
```diff
@@ -86,7 +86,7 @@ describe("GroupMembersTable", () => {
       );
 
       await waitFor(() => {
-        expect(screen.getAllByRole("checkbox").length).toBeGreaterThan(0);
+        expect(screen.getAllByRole("checkbox").length).toBeGreaterThan(1);
```
abuyukyi101198, line 89: Same here.

abuyukyi101198 (approved):
Some questions inline:

abuyukyi101198:
Looks like there are `grid` and `cell` captures in the Cypress tests also.

src/app/base/components/node/networking/NetworkTable/NetworkTable.test.tsx
ndv99: So this is because the checkboxes now have aria-hidden in this state from what I understand  - so they'll still be visible in the DOM but not in the accessibility tree

src/app/settings/views/UserManagement/views/Groups/components/GroupEntitlementsTable/GroupEntitlementsTable.test.tsx
ndv99: Sort-of; the loading state now shows the select-all checkbox, so if we wait for only 1 checkbox to appear then it resolves too early, while waiting for more  than 1 ensures we're seeing the table itself load

ndv99:
Cypress with-users seems to be failing spuriously at the keycloak setup step, but the assertions for cells and grids have been fixed.

PR #6097, opened 2026-07-06
fix: change sitespeed.io db source

nehjoshi:
## Done

- Updated sitespeed.io test db source to a 26.04 sampledata database
- A test run that passes with this change: https://github.com/nehjoshi/maas-ui/actions/runs/28777436016/job/85324641217


ndv99 (approved):
LGTM

PR #6100, opened 2026-07-09
fix(images): ws connection count dependency for boot resource polling

abuyukyi101198:
## Done

- Added WS connection count as a dependency for boot resource polling to prevent refresh failures

## QA steps

- [x] Open the maas-tor3.internal UI
- [x] Navigate to /images
- [x] Verify the table loads
- [x] Refresh
- [x] Verify the table does NOT load
- [x] Connect the local UI to maas-tor3.internal
- [x] Navigate to /images
- [x] Verify the table loads
- [x] Refresh
- [x] Verify the table DOES load


nehjoshi (approved):
+1

PR #6101, opened 2026-07-09
fix: wait for websocket connection before rendering content

nehjoshi:
## Done

- This bug was observed in TOR3, that causes the subnet details page (`/subnet/:id/summary`) to return "Subnet not found" when the page is visited directly via its URL, or by refreshing the page
- This occurs due to the page loading before the websocket connection is established
- The fix is to wait for at least 1 successful websocket connection before rendering the page

## QA steps
Use TOR3 MAAS as the backend (`http://maas-tor3.internal`) 
- [x] Sign in to TOR3 MAAS
- [x] Go to the "networks" page from the side navigation
- [x] Visit the page for any subnet
- [x] Repeatedly refresh the page and ensure you do **not** get a "subnet not found" error
- [x] If you do this on the real, deployed TOR3, you will see that this error frequently occurs when you refresh the page for a subnet

### DIFF
--- a/src/app/App.tsx
+++ b/src/app/App.tsx
@@ -79,6 +79,7 @@ export const App = (): React.ReactElement => {
   const authenticating = useSelector(status.authenticating);
   const connected = useSelector(status.connected);
   const connecting = useSelector(status.connecting);
+  const connectedCount = useSelector(status.connectedCount);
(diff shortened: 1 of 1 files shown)


abuyukyi101198 (approved):
LGTM!

PR #6102, opened 2026-07-10
fix: update app to use /me:get_entitlements endpoint

nehjoshi:
## Done

- Previously, the backend returned the user's entitlements in `/users/me`
- Now, there is a separate endpoint for that: `/users/me:get_entitlements`
- This change makes use of the new endpoint. The majority of the refactor is updating tests to use the new resolver

## QA steps
Use a local MAAS installation (`latest/edge`) as the backend, as `maas-ui-demo` does not have the latest changes yet.
- [x] Sign in to MAAS
- [x] If you are an admin user, ensure the settings tab is visible in side navigation and you can view it
- [x] Ensure the overall behaviour of the app is unchanged
- [x] Sign in as a user that belongs to the "user" group and not in the "administrators" group
- [x] Go to your profile page (`/account/prefs/details`)
- [x] Ensure you can view your profile
- [x] Ensure that making changes and saving does not raise an error


ndv99 (approved):
LGTM! Thanks for this

PR #6110, opened 2026-07-23
chore: MAAS as a snap only - drop install_type from websockets

r00ta:
## Done

- In the backend we removed `install_type`, as every controller is supposed to be running in a snap env in 4.0. So I removed `install_type` and all the logics around that
- updated the image source keyring path, as it'll be always the snap one

## QA steps

- [x] run the UI with the backend running this patch https://github.com/canonical/maas/pull/504
- [x] checked the image source page
- [x] checked the controller page


nehjoshi (changes requested):
Looks good for the most part. There are still some  trailing references to `install_type` to remove:
- `src/app/store/controller/types/enum.ts`: Since we only support snap, we can remove the enum entirely as it leaves only 1 option
- `src/app/store/controller/types/base.ts`: Remove `install_type` from `ControllerVersions`
- `src/app/controllers/views/ControllerList/components/VersionColumn/VersionColumn.tsx`: Conditionally renders controller install type
- `src/app/base/components/node/OverviewCard/ControllerStatusCard/ControllerStatusCard.tsx`: We can display the snap channel directly without checking if `install_type==='deb'`.

- Some references to install_type can be removed from `src/app/base/components/node/OverviewCard/ControllerStatusCard/ControllerStatusCard.test.tsx`

r00ta:
Thanks! I addressed the comments and re-tested

nehjoshi (approved):
LGTM!

PR #6121, opened 2026-08-13
chore: remove vm hosts MAASENG-6761

nehjoshi:
## Done

- Removed KVM hosts from the UI, and all related references and utils following its backend removal
- Updated unit and end-to-end tests

## Purged Content
- `src/app/kvm/**`
- `src/app/store/pod/**`
- `src/app/store/vmcluster/**`
- `src/testing/factories/vmcluster.ts` and pod/vmcluster factories
- All `/kvm` routes removed from `router.tsx`
- All `kvm` URLs removed from `urls.ts`
- The "KVM" nav group is removed from `AppSideNavigation`
- Pod/vmcluster reducers and states have been removed from `root-reducer.ts`

## QA steps
For the most accurate QA, use a `latest/edge` installation of MAAS, ideally with some sample data.
- [x] Sign in to MAAS
- [x] Ensure the side navigation no longer has a "KVM" group
- [x] Ensure that the side navigation no longer has links to "LXD" and "Virsh"
- [x] Ensure going to `MAAS/r/kvm`, `/MAAS/r/kvm/lxd` returns the "Not found" page
- [x] From the machines page, select the "Group By" dropdown
- [x] Ensure "Group by KVM" and "Group by KVM type" are no longer present
- [x] Select a machine and click `Actions > Deploy`
- [x] Ensure the deployment side panel no longer has a "Register as KVM host" radio button
- [ ] (Bonus, if you have a machine with a Virsh/LXD power type) Go to the details page for a LXD/Virsh machine
- [ ] Ensure that the details row (the one with Owner, Zone, Resource Pool etc.) no longer has a "Host" label with a link to the host machine
- [x] Go through the rest of the app by navigating to Images, Devices, Controllers etc.
- [x] Ensure the console does not return any kvm-related errors or warnings, and the overall behaviour of the app is unchanged

## Fixes

Resolves [MAASENG-6761](https://warthogs.atlassian.net/browse/MAASENG-6761)

[MAASENG-6761]: https://warthogs.atlassian.net/browse/MAASENG-6761?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/app/base/components/AppSideNavigation/AppSideNavigation.tsx
```diff
@@ -172,22 +164,8 @@ const AppSideNavigationContainer = (): React.ReactElement => {
...
-      );
-
-      if (virshItemIndex > -1) {
-        navGroups[kvmGroupIndex].navLinks.splice(virshItemIndex, 1);
-      }
-    }
-
     return navGroups;
-  }, [hideVirsh]);
+  }, []);
```
abuyukyi101198: I think you can use `navGroups` directly and remove `filteredGroups`.

src/app/machines/views/MachineDetails/MachineConfiguration/PowerForm/PowerFormFields/PowerFormFields.tsx
```diff
@@ -21,11 +20,10 @@ const PowerFormFields = ({ machine }: Props): React.ReactElement => {
         <PowerTypeFields<PowerFormValues>
           customFieldProps={{
             [PowerTypeNames.LXD]: {
-              canEditCertificate: !isMachineInPod,
+              canEditCertificate: true,
```
abuyukyi101198: This field can be removed.

src/app/store/machine/types/actions.ts
```diff
@@ -285,7 +284,7 @@ type Filters = {
   [FilterGroupKey.Owner]: Machine["owner"];
   [FilterGroupKey.Parent]: Node["system_id"];
   [FilterGroupKey.Pod]: ModelRef["name"];
-  [FilterGroupKey.PodType]: Pod["type"];
+  [FilterGroupKey.PodType]: string;
```
abuyukyi101198: Shouldn't these be removed as well?

src/app/store/machine/utils/common.ts
```diff
@@ -51,10 +51,7 @@ export const getTagCountsForMachines = (machines: Machine[]): TagIdCountMap => {
  * @param machine - The machine to get the applicable field scopes.
  * @returns A list of applicable field scopes.
  */
-export const getMachineFieldScopes = (machine: Machine): PowerFieldScope[] => {
-  if (machine.pod) {
-    return [PowerFieldScope.NODE];
-  }
+export const getMachineFieldScopes = (_machine: Machine): PowerFieldScope[] => {
```
abuyukyi101198: Keeping the signature the same makes sense, but this function is not called too frequently, so you can just remove the `_machine` prop fully.

src/app/store/subnet/types/base.ts
```diff
@@ -44,7 +43,7 @@ export type SubnetBMCNode = {
 
 export type SubnetBMC = Model & {
   nodes: SubnetBMCNode[];
-  power_type: Pod["type"];
+  power_type: string;
```
abuyukyi101198: I think you created a separate constant for this, no?

src/testing/factories/subnet.ts
```diff
@@ -51,7 +50,7 @@ export const subnetBMCNode = define<SubnetBMCNode>({
 
 export const subnetBMC = extend<Model, SubnetBMC>(model, {
   nodes: () => [],
-  power_type: PodType.LXD,
+  power_type: "lxd",
```
abuyukyi101198: Again, doesn't this exist as a const instead of basic string?

abuyukyi101198 (changes requested):
Incredible work! Thanks for tackling this. Some nitpicks inline, mostly to prevent stale code.

src/app/base/components/AppSideNavigation/AppSideNavigation.tsx
ndv99: ^ this, there's nothing else here that we're filtering for

src/app/store/machine/types/actions.ts
ndv99: Yep, every filter group here to do with pods should be removed IMO

ndv99 (commented):
Looks good, provided Ahmet's comments are addressed

ndv99 (approved):
LGTM! Really well done on getting this ready for merge so quickly, will be great to have the UI back

abuyukyi101198 (approved):
LGTM!

PR #6124, opened 2026-08-20
feat(base): Implement upstream layout MAASENG-6783

abuyukyi101198:
## Done

- Updated maas-react-components to 2.8.0
- Added AppLayout local wrapper to upstream Layout
- Incorporate MainContentSection into PageContent, and removed it
- Injected SidePanel into renderWithProviders component render since it is no longer supplied by PageContent
- Introduced Layout.Skeleton as the base loading state when connections are pending, with watch conditions to prevent transient flickers
- (drive-by) CI-flaky tooltip test fix with upstream expectTooltipOnHover

## QA steps

- [x] Navigate to machines
- [x] Refresh page to watch the initial load
- [x] Verify that the page renders full table view skeleton followed by the rendered header and the loading table, and finally the full data
- [x] Repeat the refresh+watch for images
- [x] Verify the same load/hydration flow happens there
- [x] Repeat for other main views
- [x] Repeat for settings/preferences views as well
- [x] Verify the loading state is acceptable, as the secondary nav won't be populated or visible until authentication is loaded, and therefore cannot be shown
- [x] Verify dark mode toggling and defaults work as expected

## Fixes

Resolves [MAASENG-6783](https://warthogs.atlassian.net/browse/MAASENG-6783)

## Notes

Below are some justifications for the design choices:
- Implementing AppLayout within PageContent where the previous layout structure resided would have resulted in impermanent routing state across pages, since navigation is now served by AppLayout as well. Therefore, AppLayout is extracted from PageContent, and PageContent is demoted to a purely header-providing wrapper.
- The extraction of AppLayout as a single layout component shared across the app means that view component tests that used to have access to SidePanel through PageContent no longer do. Therefore, SidePanel is now injected into renderWithProviders' component rendering as a sibling.
- MainContentSection was used only by PageContent and provided no meaningful separation of concerns.
- Layout.Skeleton replaces the <SectionHeader loading/> in App.tsx to comply with the no-layout-spinners convention. This resulted in to skeleton renders; one forced in App.tsx, and one Suspense fallback through upstream Layout. This split was necessary due to the connection dependency of App.tsx rendering, and additional connection transience checks are added to ensure the switch between skeletons is seamless.
- The l-main__content class is removed from PageContent since it is supplied by the upstream Layout, and nesting the same class caused unwanted scrolling in the headers.
- getInitialDarkMode null-guards are tightened because the testing environment would not reliably supply window.watchMedia properties. Should cause no changes in production.

[MAASENG-6783]: https://warthogs.atlassian.net/browse/MAASENG-6783?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


nehjoshi (changes requested):
The loading skeletons look great! However, there's an unusual bug: If you go to any route except `/machines`, and reload the page, the user intro screen flashes briefly before you are redirected to `/machines`. 
Steps to reproduce:
1. Go to `/images`
2. Reload the page
3. You'll see the user intro page flash, and then you're redirected to `/machines`.

abuyukyi101198:
@nehjoshi I am not observing the intro redirect on my end. Can you check the network requests to see if one is failing?

nehjoshi:
@abuyukyi101198 Interesting. If I use `maas-ui-demo` as the backend instance, I don't observe the bug. But I was using a `latest/edge` VM, and that's where I see the problem. I don't think the demo instance is up to date

abuyukyi101198:
@nehjoshi I will investigate, but this looks to be caused by backend changes somehow. This PR should not affect the intro sequence.

nehjoshi:
> @nehjoshi I will investigate, but this looks to be caused by backend changes somehow. This PR should not affect the intro sequence.

Noted, I'll continue reviewing with the demo instance

abuyukyi101198:
The issue was that the useGetCurrentUser hook did not expose the statistics in its pending state, which creates a race condition where, if the statistics query does not resolve before the me query, the hook resolves the pending state without populating statistics, hence the redirect to intro. I implemented the useImages format for staged queries here, so it should be fixed. Please scrutinize @nehjoshi

nehjoshi (approved):
Thanks for the changes! LGTM

PR #6126, opened 2026-08-24
test: header tests to exists instead of assuming one match

abuyukyi101198:
## Done

- Changed header checks to `exists` rather than `contains` due to the addition of a11y header by Layout

## QA

- [ ] Run controllerList, networksSubnets and networksStaticroutes features locally
- [ ] Confirm they pass


ndv99 (approved):
LGTM

PR #6133, opened 2026-08-31
fix(base): re-added the application id

abuyukyi101198:
## Done

- Updated mrc to have Layout expose the application id
- Set the application id to the same const that was removed during refactors


ndv99 (approved):
LGTM

PR #6134, opened 2026-08-31
feat: add read-permissions to all views MAASENG-6854

nehjoshi:
## Done

- Added a `<RequireEntitlements />` parent component that checks if the user has the right entitlements to view a page. If not, it renders an error message. Otherwise, it will render the children.

## Pre-QA steps
I recommend using a `latest/edge` MAAS snap as the backend, as `maas-ui-demo` needs to be updated.

- You should have two authorization groups, "Administrators" and "Users".
- Ensure the "Administrators" group has all entitlements 
- Ensure the "Users" group has no entitlements (i.e. go to the groups page for "Users" and remove all entitlements)
- Create two users. User A should be a fully-privileged user in the "Administrators" Group.
- User B should be in the "Users" group and shall have no permissions.

## QA steps

- [x] Sign in to MAAS as user A
- [x] Navigate to each page one by one, and ensure all the routes render correctly without any permission errors
- [x] Sign out
- [x] Sign in as user B
- [x] Ensure all of the side navigation links, except the profile and "Log out" links, are disabled and grayed out
- [x] Ensure navigating to these links by typing in their URLs manually displays the "You do not have permission to view this page." message
- [x] Ensure the "Controllers" and "Settings" tabs are visible in the side panel, but disabled. 

## Fixes

Resolves [MAASENG-6854](https://warthogs.atlassian.net/browse/MAASENG-6854)

[MAASENG-6854]: https://warthogs.atlassian.net/browse/MAASENG-6854?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


ndv99 (approved):
LGTM

PR #6138, opened 2026-09-04
fix(networks): wait for promises to resolve before validating form values

nehjoshi:
## Done

- The "Configure DHCP" form for a VLAN had a bug causing the "Configure DHCP" button to be disabled despite the form having valid values. This was caused by the validation being run before the form was actually populated.
- This PR waits for the promises (for populating the form) to resolve before triggering validation.

## QA steps

- [x] Sign in to MAAS
- [x] Go to the networks page from the side nav
- [x] Locate a subnet that has "No DHCP", and select the corresponding VLAN
- [x] Select the "Configure DHCP" button
- [x] Check the "MAAS provides DHCP" option, and select any rack controller from the dropdown
- [x] Select a subnet from the "Reserved dynamic range" dropdown
- [x] Ensure the "Configure DHCP" button is now enabled (previously, due to the bug, this would still be disabled until the user focused one of the form fields)
- [x] Ensure clicking the "Configure DHCP" button adds the dynamic range to the list, enables DHCP, and closes the form


ndv99 (approved):
LGTM, thanks for this

PR #6139, opened 2026-09-04
feat: add FIPS and hardening status info to status bar MAASENG-6810

ndv99:
## Done

- Added info for hardening and FIPS to status bar
- Added link to local documentation on hardening
- Used API v3 for fetching version information
- Updated API client

## QA steps

You'll need a MAAS running the fips-3.7 branch.

- [x] Go to MAAS UI
- [x] Ensure the version displays correctly in the status bar
- [x] Either by actually enabling FIPS and hardening, or faking response data, ensure that the correct status information in the footer is displayed for:
- - [x] FIPS only
- - [x] Hardening only
- - [x] Both FIPS and hardening
- [x] Ensure that the link to documentation for hardening works and is appropriate

## Fixes

Resolves [MAASENG-6810](https://warthogs.atlassian.net/browse/MAASENG-6810)

## Screenshots
image
image
image

[MAASENG-6810]: https://warthogs.atlassian.net/browse/MAASENG-6810?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/app/apiclient/client.gen.ts
+++ b/src/app/apiclient/client.gen.ts
@@ -23,5 +23,9 @@ export type CreateClientConfig<T extends DefaultClientOptions = ClientOptions> =
   ) => Config<Required<DefaultClientOptions> & T>;
 
 export const client = createClient(
-  createClientConfig(createConfig<ClientOptions>())
+  createClientConfig(
+    createConfig<ClientOptions>({
+      baseUrl: "http://10.10.0.22:5240",
+    })
+  )
 );
--- a/src/app/apiclient/client.gen.ts
+++ b/src/app/apiclient/client.gen.ts
@@ -25,7 +25,7 @@ export type CreateClientConfig<T extends DefaultClientOptions = ClientOptions> =
 export const client = createClient(
   createClientConfig(
     createConfig<ClientOptions>({
-      baseUrl: "http://10.10.0.22:5240",
+      baseUrl: "http://localhost:8400",
     })
   )
 );
--- a/openapi-ts.config.ts
+++ b/openapi-ts.config.ts
@@ -1,7 +1,7 @@
 import { defineConfig } from "@hey-api/openapi-ts";
 
 export default defineConfig({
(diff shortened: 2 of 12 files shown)


src/app/apiclient/client.gen.ts
```diff
@@ -23,5 +23,9 @@ export type CreateClientConfig<T extends DefaultClientOptions = ClientOptions> =
   ) => Config<Required<DefaultClientOptions> & T>;
 
 export const client = createClient(
-  createClientConfig(createConfig<ClientOptions>())
+  createClientConfig(
+    createConfig<ClientOptions>({
+      baseUrl: "http://10.10.0.22:5240",
```
nehjoshi: This should probably be reverted to what it was before

nehjoshi (changes requested):
One comment below, apart from that LGTM.
Also, the failing `VirshDetails.test.tsx` test seems to be a fluke, I ran it locally and it passed.

src/app/apiclient/client.gen.ts
ndv99: Oh yeah good catch. I had a thought on this actually - should we set this baseUrl to localhost:8400? Since the API docs will also be proxied by a locally running UI, we can always point it to whatever instance we need through .env.local and get the API spec we need, whether that's 3.7, this branch, latest or whatever

nehjoshi: I think that would work, yeah! Might just wanna try out the docs and generate the API spec to confirm they still work as expected

ndv99: Docs seem to work fine, will test an API spec generation and lyk

ndv99: It works!

nehjoshi (approved):
LGTM, thanks for the url fix!

PR #6144, opened 2026-09-07
fix(settings): remove switch provisioning feature flag and switch scripts

nehjoshi:
## Done

- Removed the `VITE_APP_SWITCH_PROVISIONING` feature flag, as we have a "Enable experimental switch provisioning" config option
- Removed the "Switch scripts" settings view, as we do not currently have separate scripts for switches
- Updates types, selectors and types

## QA steps

- [x] Sign in to MAAS
- [x] Go to settings > general
- [x] Check switch commissioning and click "Save"
- [x] Ensure the side navigation includes the "Switches" tab under "Machines"
- [x] Ensure in the Settings secondary side navigation, the "Switch Scripts" tab under "Testing Scripts" is no longer present

## Notes
- To be backported to `3.8`


rumanowskak (approved):
LGTM

PR #6152, opened 2026-09-15
fix(settings): update change password form

nehjoshi:
## Done

- The `PUT /users/me` endpoint accepts new fields, `old_password` and `new_password`. 
- The endpoint checks if the `old_password` is correct, thus, the UI no longer needs to do `authenticate.mutate(oldPassword)` in the change-password form as it is handled by the backend

## QA steps
Note: `maas-ui-demo` can be used as the backend
- [x] Sign in to MAAS
- [x] Go to the user's details page (`/account/prefs/details`)
- [x] Expand "Change password"
- [x] Enter your old password, and a new password
- [x] Click "Save"
- [x] Sign out, and sign back in with the new password, and ensure it works without any errors


abuyukyi101198 (approved):
LGTM

